PR #215, opened 2023-03-30
fix(locales): fix french

kapik:


riderx (approved):
I feel bad aha, I'm french and the first translation improvement made is french hahaha, but thank you so much for this!

PR #267, opened 2023-08-14
fix: Find a better way to block google play test device 

### LINKED ISSUE #229: Find a better way to block google play test device
For a period i did set a ip ban for google IP sadly it was also blocking some user with vpn.
i need to find a better option

rishi-raj-jain:
fix #229
/claim #229


riderx:
This is not google bot.
What google does is when user submit to google they new app update google start 7 to 14 android device to start the app in it.
That what we want to block

rishi-raj-jain:
@riderx 

Can you help me understand that what's the frequency of this is like? Is this causing a large addition to user billing and if rate limiting is a viable solution?

riderx:
For users who have CI/CD setup it represents a lot of they bill.
The problem is the app is the same the final user will use, we cannot make it different.
Something has to be done with IP like I was doing but more smart

rishi-raj-jain:
How about IP based rate limiting?

riderx:
Each device get its own IP and if the user send 30 update in one hour users should receive it as well.
Here is the current solution I was doing 
https://github.com/Cap-go/capgo/blob/07f3d52e4f916cc50ea748635592c8886e1797a0/supabase/functions/updates/index.ts#L267C5-L267C5

rishi-raj-jain:
// TODO: find better solution to check if device is from apple or google, currently not qworking in netlify-egde
    
    We can do that right now: 
    
    ```
     // Extract the User-Agent header from the request
  const userAgent = headers['user-agent'] || '';

  // Check if the User-Agent indicates an iOS device
  const isiOS = userAgent.includes('iPhone') || userAgent.includes('iPad') || userAgent.includes('iPod');

  // Check if the User-Agent indicates an Android device
  const isAndroid = userAgent.includes('Android');
  ```

rishi-raj-jain:
So we can do is maintain an attribute to the database that contains the app updates? So say someone made an update, we detect:

- if the last update time is lesser than the current time
- the current version isn't equal to the last version (recorded at the time of the last update)

fi these are true, then we'd proceed. I think we can do this with redis like database where we can say IP as the key to a hlist, the version as the key and the time updated as the value of it.

something like

await redis.hset(IP, version, timeUpdated)

does that make sense?

rishi-raj-jain:
Okay so now we can when it enters the if case of android/ios and is yet to receive an update, we can associate an IP with the last update version. Does that make sense?

rishi-raj-jain:
@riderx how does the logic look now? Let me know if anything doesn't make sense there.

rishi-raj-jain:
@riderx 

Rather than looking for IP or anything, just see if an update is available. Shifted the logic to a better.

rishi-raj-jain:
If yes, it has to anyways update, if not it just returns. Hope this works?

rishi-raj-jain:
Updated the logic even better to not even call the update even if the conditions don't match.

rishi-raj-jain:
If you want, what we could now do is before the deviceupdates, we can add rate limiting. How it'd work is as follows:

- by default it's set to forever blocked
- as a version updates, the rate limit is set to 1, once done, it gets back to forever blocked (by deleting key in redis)

riderx:
> // Your comment // TODO: find better solution to check if device is from apple or google
> 
> We can do that right now:
> 
> ```
> // Extract the User-Agent header from the request
> const userAgent = headers['user-agent'] || '';
> 
> // Check if the User-Agent indicates an iOS device
> const isiOS = userAgent.includes('iPhone') || userAgent.includes('iPad') || userAgent.includes('iPod');
> 
> // Check if the User-Agent indicates an Android device
> const isAndroid = userAgent.includes('Android');
> ```

What i mean was by they device farm, not generally

riderx:
Currently we don't have redis in the stack, so it's not an option for now.
Only Supabase will be use to store data.

Capgo has currently 1M device in DB so storing they IP will cost a lot, for something happening only for few clients.

Can you add a schema to your idea solution to help me understand it better?

rishi-raj-jain:
@riderx 

Sure, I'll be on this in abour 6 hours!

rishi-raj-jain:
@riderx as per https://github.com/Cap-go/capgo/issues/229#issuecomment-1705741679, I'll get onto this.

rishi-raj-jain:
client-ip is always included in netlify event function: https://github.com/netlify/cli/issues/412

supabase/functions/updates/index.ts
```diff
@@ -32,6 +59,25 @@ async function main(url: URL, headers: BaseHeaders, method: string, body: AppInf
...
 
+  const ipAddress = [headers['x-real-ip'], headers['x-forwarded-for'], headers['client-ip']].find(i => i !== undefined && i !== null)
+  if (ipAddress) {
+    const getCurrentCount = await getDeviceRequestCount(ipAddress)
+    if (getCurrentCount) {
+      if (getCurrentCount === -1) {
+        return sendRes({ error: `Blacklist` }, 403)
+      }
+      // TODO: Update limit to blacklist flag
+      else if (getCurrentCount > 10 ** 5) {
```
rishi-raj-jain, line 72: what's a fair limit to block? @riderx & @WcaleNieWolny

rishi-raj-jain:
I can also add a term limit, that's like X updates within Y before the IP gets blocked.

WcaleNieWolny:
Also why is this in redis cache? Redis cache does not add to usage as far as i know

WcaleNieWolny:
We just return a cached response and do not do anything as far as I know. You do not pay for cached request, and IF the device is not cached then the old logic is executed. I would place this limit there

riderx:
@rishi-raj-jain this goal here is to have zero device from Google data center to download update, otherwise it's counted in the stats of the user.
So the aim is to find the right solution to identify them, with the headers they send

rishi-raj-jain:
Hey @riderx,

It indeed will help us establish what we aim at finding.

Using this method, we can automatically block those IPs that come in with large amounts of requests. As you told in a comment, that this is very particularly occuring for a few customers, I believe having this check in place is just gonna help us achieve the intended.

Thoughts? @WcaleNieWolny @riderx

rishi-raj-jain:
do you have any apps published where we can test this flow?
as in a test app where we can manually update and see what all request come in for that particular app? 
that can be as simple as just us storing all the requests for a specific app_id and inspecting the logs.
we can save all the headers in redis for the timebeing (for that app) and clear the records later 
soo wait I think the easiest way to test is:

you have an app that you can release an update to
set the update URL as a URL that I'll give to you
leave the rest to me 😎
basically, we change the capgo update endpoint for that app to /update_fake and keep logging whatever comes to it, and inspect.
we can also do the same on production right now by setting up a middleware that'll make an upstream call to an endpoint with all the data we need to store before /update is processed
I can do that in my Upstash redis so that you don't bear any cost of it
but yeah the safest way seems to be store some data and try to see if some X requests can be grouped together by some logic

supabase/functions/_utils/redis.ts
```diff
@@ -39,6 +43,14 @@ class RedisRedisPipeline implements RedisPipelineInterface {
   async hset(key: string, field: string, value: RedisValue): Promise<void> {
     await this.pipeline.hset(key, field, value)
   }
+
+  async hincrby(key: string, field: string, increment: number): Promise<number> {
+    return await this.pipeline.hincrby(key, field, increment)
```
wesleymatosdev, line 48:
```suggestion
    return this.pipeline.hincrby(key, field, increment)
```

supabase/functions/_utils/redis.ts
```diff
@@ -39,6 +43,14 @@ class RedisRedisPipeline implements RedisPipelineInterface {
...
+  }
+
+  async hget(key: string, field: string): Promise<string | null> {
+    return await this.pipeline.hget(key, field);
```
wesleymatosdev, line 52:
```suggestion
    return this.pipeline.hget(key, field);
```

supabase/functions/_utils/redis.ts
```diff
@@ -68,6 +80,15 @@ class RedisUpstashPipeline implements RedisPipelineInterface {
...
     this.size++
   }
+
+  async hincrby(key: string, field: string, increment: number): Promise<number> {
+    await this.pipeline.hincrby(key, field, increment);
+    return 0
+  }
+
+  async hget(key: string, field: string): Promise<string | null> {
+    return await this.pipeline.hget(key, field);
```
wesleymatosdev, line 90:
```suggestion
    return this.pipeline.hget(key, field);
```

supabase/functions/_utils/redis.ts
```diff
@@ -100,6 +121,14 @@ export class RedisRedis implements RedisInterface {
   async hmget(key: string, ...fields: string[]): Promise<(string | null | undefined)[]> {
     return await this.redis.hmget(key, ...fields)
   }
+
+  async hincrby(key: string, field: string, increment: number): Promise<number> {
+    return await this.redis.hincrby(key, field, increment);
```
wesleymatosdev, line 126:
```suggestion
    return this.redis.hincrby(key, field, increment);
```

supabase/functions/_utils/redis.ts
```diff
@@ -100,6 +121,14 @@ export class RedisRedis implements RedisInterface {
...
+  }
+
+  async hget(key: string, field: string): Promise<string | null> {
+    return await this.redis.hget(key, field);
```
wesleymatosdev, line 130:
```suggestion
    return this.redis.hget(key, field);
```

supabase/functions/_utils/redis.ts
```diff
@@ -143,6 +172,15 @@ export class RedisUpstashImpl implements RedisInterface {
     })
   }
 
+  async hincrby(key: string, field: string, increment: number): Promise<number> {
+    const result = await this.redis.hincrby(key, field, increment);
+    return result;
+  }
+
+  async hget(key: string, field: string): Promise<string | null> {
+    return await this.redis.hget(key, field);
```
wesleymatosdev, line 181:
```suggestion
    return this.redis.hget(key, field);
```

supabase/functions/updates/index.ts
```diff
@@ -25,6 +25,33 @@ const headersSchema = z.object({
...
+  if (redis) {
+    const deviceIpHashKey = `device_requests:${createHash("md5").update(ipAddress).toString()}`;
+    const count = await redis.hget(deviceIpHashKey, 'count');
+    if (count !== null) {
+      return Number(count);
+    }
+  }
+}
+
+async function blacklistDeviceRequest(ipAddress: string): Promise<void> {
```
wesleymatosdev, line 47: Can we use another term than blacklist? This one has an implicit racial connotation, you can read more about it [here](https://www.splunk.com/en_us/blog/learn/blacklist-whitelist-inclusivity.html#:~:text=The%20term%20%E2%80%9Cblacklist%E2%80%9D%20contains%20the,and%20contribute%20to%20systematic%20racism.).

We could use block list instead.

And change the name to something like 
```suggestion
async function addDeviceToBlockList(ipAddress: string): Promise<void> {
```

supabase/functions/updates/index.ts
```diff
@@ -34,6 +61,25 @@ async function main(url: URL, headers: BaseHeaders, method: string, body: AppInf
...
+      // TODO: Update limit to blacklist flag
+      else if (getCurrentCount > 10 ** 5) {
+        blacklistDeviceRequest(ipAddress)
+        return sendRes({ error: `Blacklist` }, 403)
+      }
+      await incrementDeviceRequestCount(ipAddress);
+    }
+    else {
+      await incrementDeviceRequestCount(ipAddress);
+    }
```
wesleymatosdev, line 80: If on the else we'll run the `incrementDeviceRequestCount` anyways, we could do it outside the condition block instead

```suggestion
    }
    await incrementDeviceRequestCount(ipAddress);
```

ologbonowiwi (changes requested):
Left some comments. Most regarding unnecessary awaits on returns.

When you `return promise`, the receptor it'll already need to use `await`, so the `return await` expression gives exactly the same result as `return`

supabase/functions/updates/index.ts
```diff
@@ -25,6 +25,33 @@ const headersSchema = z.object({
...
+
+async function getDeviceRequestCount(ipAddress: string): Promise<number | undefined> {
+  const redis = await getRedis();
+  if (redis) {
+    const deviceIpHashKey = `device_requests:${createHash("md5").update(ipAddress).toString()}`;
+    const count = await redis.hget(deviceIpHashKey, 'count');
+    if (count !== null) {
+      return Number(count);
+    }
+  }
```
wesleymatosdev, line 44: Do you mind using early return? It makes the code more readable 
```suggestion
  if (!redis) return;

  const deviceIpHashKey = `device_requests:${createHash("md5").update(ipAddress).toString()}`;
  const count = await redis.hget(deviceIpHashKey, 'count');
  if (count !== null) {
    return Number(count);
  }
```

supabase/functions/updates/index.ts
```diff
@@ -25,6 +25,33 @@ const headersSchema = z.object({
...
+}
+
+async function blacklistDeviceRequest(ipAddress: string): Promise<void> {
+  const redis = await getRedis();
+  if (redis) {
+    const deviceIpHashKey = `device_requests:${createHash("md5").update(ipAddress).toString()}`;
+    await redis.hset(deviceIpHashKey, 'count', -1);
+  }
```
wesleymatosdev, line 52:
```suggestion
  if (!redis) return;

  const deviceIpHashKey = `device_requests:${createHash("md5").update(ipAddress).toString()}`;
  await redis.hset(deviceIpHashKey, 'count', -1);
```

riderx:
yes @rishi-raj-jain we can do it with Capgo app, who is auto send to google play internal tester who trigger the google thing

riderx:
please look the comment of @ologbonowiwi

ologbonowiwi (commented):
I made these comments while ago but forgot to submit the review. Submitting it now 😄

rishi-raj-jain:
@riderx 

> yes @rishi-raj-jain we can do it with Capgo app, who is auto send to google play internal tester who trigger the google thing

The approach we talked about in Discord, correct?

riderx:
yes the idea is to have a custom endpoint who store all headers received, and use this endpoint in Capgo app to let the CI do the job and google run the app

ologbonowiwi:
Hey @rishi-raj-jain!

Don't forget to re-open this

ologbonowiwi:
hey @rishi-raj-jain, will you re-open this, or can I take it up from here?

Issue #269, opened 2023-08-16
Allow change email


riderx:
Make feature to change email.

It should update auth user account, and table users column email in public schema.
The email should be checked if exist with our custom Postgres method `exist_user`

riderx:
to update the table users, you need to manage to allow it with RLS without breaking security ( like anyone can update users then)

riderx:
Agree the security allow it now, I would like to improve that, I can do a separate task.
![CleanShot 2023-08-16 at 15 59 10@2x](https://github.com/Cap-go/capgo/assets/4084527/00fc5b82-3f77-4002-a3a4-972b2dc05f8c)
Right now there are one other problems, in production a confirmation email is sent and when clicked the email is updated, so it's at this moment the update should happen for the public user.
It's also necessary to check if the confirmation work for email change

riderx:
Hey @neo773 thanks for the update !
I'm not sure to understand how @WcaleNieWolny was involved.
The bounty was self close so i couldn't do the split, but i can tips him

riderx:
Oh okay thanks a lot <3
/tip 30 @WcaleNieWolny

riderx:
Thanks for the screenshot that help me to understand how shitty the dev env is ^^

riderx:
with the last commit i made yesterday it should be easier to work on bounty, i will create a discord channel to help each others

PR #272, opened 2023-08-16
feat: Allow users to change email

neo773:
This PR allows users to change their email

https://github.com/Cap-go/capgo/assets/62795688/bc3b7834-9454-42fd-b528-a1249db705dd

/claim #269


neo773:
> Almost good, you need to update the table users as well, and manage to allow it with RLS without breaking security

The user table is still updated and `supa.auth.updateUser` is needed to update the user's auth email which is separate from users table.

image

riderx (approved):

Issue #270, opened 2023-08-16
Removing bundle does not remove data from S3


WcaleNieWolny:
Currently removing the bundle removes the bundle ONLY from supabase and keeps it in S3. I assume that this is not the intented behavior and this could be causing unnecessary S3 costs. Adding this functionality is critical for solving   https://github.com/Cap-go/capgo/issues/257

WcaleNieWolny:
Note: This is a very complex issue that goes beyond the documented self hosted setup. I am aware that self hosting S3 is not official supported, but if this does not work on self hosted then neither does it work in production. This issue generates no errors as far as I can see, and it involves ALL of the web hooks (automatic triggers) as all of them are misconfigured

https://github.com/Cap-go/capgo/assets/50914789/c5a89090-61aa-4d44-973d-59b10d61dc99

WcaleNieWolny:
I believe the solution to this is as follows:

A) Fix all automatic triggers so they accualy work
B) Document/add functionality to CLI to make `upload` work on self hosted setup so the community can work on S3 related issue
C) Document how to get S3 running in a sensible way, This might be difficult as edge functions are running in a seperated docker container that has no access to localhost, for now I have just exposed my local S3 server by a SSH tunnel but this is a very temporarily institution

WcaleNieWolny:
Unless I am missing something or the webhooks are written this way on purpose (to make it harder to self host or to prevent errors from happening on the self hosted cloud functions) then this is really not great

WcaleNieWolny:
Is it okay if I claim this in this with https://github.com/Cap-go/capgo/pull/273 ? There is still no docs for S3, but self hosting S3 requires changes in the CLI and it might break the official S3 server (see below)

![image](https://github.com/Cap-go/capgo/assets/50914789/3a6e9442-1330-432f-9a94-79b9f93fa2e1)

As far as I know these headers broke `minio` which is the most popular (as far as I know) self hosted S3 server

riderx:
The point is to claim it with it yes :)
I think for self-hosted we will keep the supabase solution, s3 is here for cost reason for the cloud mainly

WcaleNieWolny:
I do not think I can claim this (Unless algora is down)
![image](https://github.com/Cap-go/capgo/assets/50914789/828fe419-cf26-4519-8a6a-5c4c40bf5f63)

riderx:
They maybe didn't expect 2 PR for 2 claim, I will do it for you

[...]

Issue #257, opened 2023-07-27
Allow to set in your account bundle retention


riderx:
I can set after X days my old bundle will be removed.
Create_at has to be before than now - days.
Bundle meta updated_at has to be before than now - days.
Bundle has to be not in use in any channel

akhill10:
Hello @riderx , I would like to take this up.

If I understand it correctly, we need to perform the following tasks:

1. Introduce a new column `retention` to the `apps` table which stores the day's value. And an API to upsert the retention value for the apps
2. A cron job that runs every day, which is responsible for checking the `app_versions` table and set the bundles `deleted` to true for those whose created_at has to be before than now - days,   updated_at has to be before than now - days and bundle has to be not in use in any channel. And also responsible for deleting the corresponding buckets.


/attempt #257

riderx:
Hello thanks for the questions:

- yes for the new column, but no need for API since we use Supabase it will just be done with Supabase SDK what is necessary is allowing setting it from the CLI.
With a new command `npx @capgo/cli@latest app set [appId] --retention`
if they put 0 then it's infinite.
- yes, but the cron has to look at created_at and last update of app_versions_meta. We don't want to delete version who have been in use in retention period. 
app_versions_meta is update each time user download the bundle.
You have also to check if app_versions is not in use in any channel, if yes they are ignored.

Then when you delete is same as current workflow you delete the storage but not the row in the db for stats purposes

akhill10:
@riderx Thanks for more information.

But I am facing a hard time setting up the project in local. `supabase start` throwing an error. Basically it's not initialising the schema from `schema.sql`.

```
Seeding data supabase/seed.sql...
ERROR: relation "public.plans" does not exist (SQLSTATE 42P01)
Try rerunning the command with --debug to troubleshoot the error.
```
Can you please help me out ?

riderx:
I just updated the repo with new schema yesterday. Maybe that could help you to pull ?
Then try see if your cli supabase is the latest

akhill10:
@riderx Yeah, I am on the latest commit and also latest supabase cli `1.83.7`

riderx:
I will check Monday

WcaleNieWolny:
@akhill10 Can I attempt this? I already have supabase set up and it works on my machine

akhill10:
[...]

PR #273, opened 2023-08-17
Implement retention

WcaleNieWolny:
This PR partially fixes https://github.com/Cap-go/capgo/issues/270 and it implements https://github.com/Cap-go/capgo/issues/257. I have added a `retention` column in the `apps` table that defaults to `2592000`. 

I also added a cron job to remove any version that is older then retention as mentioned in the issue.
I also fixed the webhooks from not firing. The wierd IP was found in some supabase issue, but I no longer have the link to that.
I tested all of my changes. Here is a video of these changes in action:

https://github.com/Cap-go/capgo/assets/50914789/f2933805-bf30-4925-a8ed-2fdb4ca6f70f

There is a seperate PR [here](https://github.com/Cap-go/CLI/pull/120) in the CLI to set retention from the CLI

/claim #257

/claim #270


supabase/seed.sql
```diff
@@ -111,4 +111,15 @@ SELECT http_set_curlopt(''CURLOPT_TIMEOUT_MS'', ''15000'');
            ''application/json'',
            ''{}''
         )::http_request)
-    ', 'localhost', 5432, 'postgres', 'supabase_admin', 't', 'cron_everyday_web_stats');
+    ', 'localhost', 5432, 'postgres', 'supabase_admin', 't', 'cron_everyday_web_stats'),
+(15, '0 0 * * *', '
+    UPDATE app_versions
+    SET deleted = true
+    FROM apps
+    WHERE app_versions.storage_provider=''r2''
```
riderx: any storage is ok, I don't think we need to select only r2

supabase/seed.sql
```diff
@@ -111,4 +111,15 @@ SELECT http_set_curlopt(''CURLOPT_TIMEOUT_MS'', ''15000'');
...
+    AND app_versions.id not in (select app_versions.id from app_versions join channels on app_versions.id = channels.version)
+    AND app_versions.deleted = false
+    AND apps.retention > 0
+    AND extract(epoch from now()) - extract(epoch from app_versions.created_at) > apps.retention
+    AND extract(epoch from now()) - extract(epoch from app_versions.updated_at) > apps.retention
```
riderx: I think the updated_at check should be on app_version_meta, that where the stats are stored, and stats are produce on active bundles

riderx:
Thanks a lot, that a very good solution you made !
I like the cron job only in SQL.
I think that the best solution, after talking about my comments

supabase/seed.sql
WcaleNieWolny: Fixed in latest commit

supabase/seed.sql
WcaleNieWolny: Would you prefer
```
WHERE app_versions_meta.id = app_versions.id
``` 
or
```
WHERE app_versions_meta.app_id = app_versions.app_id
```
(both work)

WcaleNieWolny: I prefer the second option as it is less likely to break

riderx: agree

WcaleNieWolny:
@riderx Resolved all of the issues mentioned by you

riderx:
I think we have tiny issue the CLI Pr say the retention is set in days (what is logical for me)
but the cron task look for seconds right ?

WcaleNieWolny:
That is true, timestamp diff is in seconds. I did not checked my PR with not my CLI changes

riderx:
I just modified the CLI code to received days and transform it in sec

PR #266, opened 2023-08-14
feat: implement regenerating API key

WcaleNieWolny:
This PR implements the ability to regenerate API keys. I have included a demo of that below

https://github.com/Cap-go/capgo/assets/50914789/4a7d38ce-fbbc-46b3-bc53-fbbfbdc73dd0

Please note that that I18n translations would be nice, also if someone has any idea on how to improve `computedValue` please let me know

/claim #265

PR #274, opened 2023-08-17
Fix regenerate api keys components

WcaleNieWolny:
This PR fixes [this](https://github.com/Cap-go/capgo/pull/266#issuecomment-1680852310)

Before:
image

After this change:

image
image


riderx:
Thanks!

PR #275, opened 2023-08-17
feat: bundle retention cron job

### LINKED ISSUE #257: Allow to set in your account bundle retention

akhill10:
- [x] Added a new column `retention` which defaults to `null`

Cron job that does the following: 
- [x] Check if created_at has to be before than now - days, updated_at has to be before than now - days 
- [x] Check if app_versions is not in use in any channel, if yes they are ignored.
- [x] delete bucket: set app version deleted to true and bucket_id to null , set size to 0 and remove the bucket from the storage 

Fixes #257 
/claim #257


riderx:
close in favor of the other PR

Issue #278, opened 2023-08-19
Fix delete account


riderx:
Make delete account really delete account, the app and all the data the user had.
Modify the deleted_account table to save a hash of the email instead of the email itself for GDPR compliance.

riderx:
2 postgres function should be created:
add_deleted_user who encrypt the email and add it to deleted_account
is_deleted_user who take email in param and encrypt it to see if it match any row in deleted_account

riderx:
example for add_deleted_user
```
create
or replace function add_deleted_user () returns void as $$
BEGIN
  INSERT INTO deleted_account (email) VALUES (crypt(auth.email(), gen_salt('bf')));
END;
$$ language plpgsql;
```

riderx:
Hey @pbkompasz thanks for this PR, here some comment:
- you didn’t took in account some table have delete on cascade so you didn’t needed to delete yourself the data.
- The function to check if account is delete is not created and not used in the register form 
The function your created don’t use auth so anyone can add email on this table and prevent everyone to login

PR #279, opened 2023-08-19
Fix: delete account

neo773:
This PR makes the delete user button delete all associated user data
hashed email in `deleted_user`

image

/claim #278


riderx:
Hey @neo773 thanks for this, the db has already on delete cascade for this, so I don't believe a function is needed.
The deleted_account should be done by the database itself, not in the code.
https://supabase.com/docs/guides/database/column-encryption

Szczurowsky (changes requested):
This PR does not check when registering an account if the email used in said registration has been used before

PR #281, opened 2023-08-20
feat: add_deleted_user, is_deleted_user

pbkompasz:
/claim #278


WcaleNieWolny:
@pbkompasz Where are the frontend changes for this? What is the point of these changes if you do not use them?

Szczurowsky (changes requested):
There's no front-end utility therefore what's the point of these changes

PR #290, opened 2023-08-21
Fix deleting account

WcaleNieWolny:
Fixes https://github.com/Cap-go/capgo/issues/278

Changes:
 
 * It generates hash client side
 * Adds a rule allowing for inserting hashed email
 * Fixes cascade for notifications
 
 Feel free to reject if you do not like these changes or think that I should not get this bounty as I am the 3rd person to claim it

/claim #278

### DIFF
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -317,6 +317,7 @@ usage: Usage
 usage-success: Your plan has been setup this page will be update, in few minutes
 usage-title: Total
 use-the-command: 'Use the command:'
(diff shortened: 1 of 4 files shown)


Szczurowsky (approved):
LGTM!

riderx:
This will be close in favor of https://github.com/Cap-go/capgo/pull/279

riderx:
The check login part will be used in your PR so for this i tip you

PR #291, opened 2023-08-21
fix: handle delete account

akhill10:
- created a trigger upon user delete.
- The above trigger is responsible for creating a hash of email and storing into deleted_account table
- updated account.vue to actually delete the user

/claim #278


Szczurowsky (changes requested):
By encrypting the key instead of hashing we still have access to users email. By publicly exposing the key in the schema this counter measure is basically useless. If someone gains access to supabase he will be easily able to decrypt emails and thus the app is not GDPR complaint

riderx:
Close in favor of https://github.com/Cap-go/capgo/pull/279

PR #296, opened 2023-08-22
Improve language selector

neo773:
This PR improves the language selector by fixing overflow, scrolling and adding actual language names

https://github.com/Cap-go/capgo/assets/62795688/b2ce0d52-4f2b-46d0-b3de-2effaecd5f0c


riderx:
Thanks !

Issue #297, opened 2023-08-22
Missing API keys on the Dashboard of a New User


ayewo:
There’s a JS bug that pops up in the browser console in prod that prevents the API key from being shown for a new sign up.

![image](https://github.com/Cap-go/capgo/assets/20957603/b4ef2c98-8c3a-4ccf-a67f-93065df2aa51)

It seems no API keys were created for the user account during sign up since the error is coming from this line: 
https://github.com/Cap-go/capgo/blob/d970fa9b10fd8f1b11904b8b5c3ef31fcf93628b/src/components/dashboard/Steps.vue#L106 
where the expression `data[0].key` was evaluated as undefined.

You can check this account `capgoupdater1@ayewo.com` to investigate the issue further.

PR #309, opened 2023-08-25
fix api key not working

neo773:
This PR fixes API key not being displayed to new sign ups

https://github.com/Cap-go/capgo/assets/62795688/3ab7912d-e65c-4a7f-a59c-7c1daf8ca4e5

/claim #297


WcaleNieWolny:
Can you use the existing code to generate API keys? I wrote some functions to do that some time ago

neo773:
> Can you use the existing code to generate API keys? I wrote some functions to do that some time ago

I tried that but Vue doesn't support exporting functions inside Component code

riderx (approved):

Issue #258, opened 2023-07-27
Organization system


riderx:
Instead of link app to user, we want to create org for user and link app to org.
That will allow us to give org right to more than one user.

riderx:
Current status:
- org model have been added in the DB
- org user role have been added in the DB
- default org are created for every new user who register
- one org by user has been created for previous users
- admin user's rights have been granted to all org owners
- channel users have been copied as org right
- Figma for frontend has been drafted: https://www.figma.com/proto/18gJfn8NiwEMxSosZzX2G7/Capgo-Organisation?page-id=0%3A1&type=design&node-id=38-81&viewport=4233%2C229%2C0.51&t=Uz41DeV2hkH37JFj-1&scaling=min-zoom&starting-point-node-id=38%3A81&mode=design

riderx:
To do: 
- Finish Figma design
- Implement Figma in frontend
- display org
- allow creating delete or edit your org in front
- display org users
- allow adding delete org users in front
- allow update org users right in front
- add feature to add user to org in front
- add security rule in new tables, org, and org rights
- create a Postgres function to check user rights for org
- create a Postgres function to check user rights for app
- create a Postgres function to check user rights for channel
- transfer apps to org
- update security rules to use new org rights
- transfer billing info from user to org
- display org billing
- ...

YacineMessaadi:
- [ X ]  Finish Figma Design
=> Desktop : [click here](https://www.figma.com/proto/18gJfn8NiwEMxSosZzX2G7/Capgo-Organisation?page-id=0%3A1&type=design&node-id=38-81&viewport=1274%2C415%2C0.31&t=3GKnOtAxieWGO77b-1&scaling=min-zoom&starting-point-node-id=38%3A81&show-proto-sidebar=1&mode=design)
=> Mobile : [click here](https://www.figma.com/proto/18gJfn8NiwEMxSosZzX2G7/Capgo-Organisation?page-id=106%3A35&type=design&node-id=106-269&viewport=6230%2C195%2C0.68&t=zoqdwuws9D2FzwZR-1&scaling=min-zoom&starting-point-node-id=106%3A269&mode=design)

riderx:
@YacineMessaadi thanks for the work.
Some remarks though :
- the current style of tabs don’t match the one in figma, i would prefer we follow the same as we have already.
- Org will have logo as well to edit and should have the same design as current setting screen
- I don’t see the edit user
- Apikey need to have action: delete recreate, added 
[...]

PR #315, opened 2023-08-28
Organization system

pbkompasz:
Part of #258


WcaleNieWolny:
What is your vision for this members panel? I have a couple of questions.

A) What is `member.position` `member.role` These values are not present in the current db scheme and I am not sure what they suppose  to be.
B) What is `members.value = organizationStore.getMembers(currentOrganization.id)` `getMembers` does not exist
C) How does the member tab should look like? I managed to get the members pannel to the following state, however I really do not like how it looks

image

WcaleNieWolny:
Also please explain what on earth is the `get_orgs` function. I thought that the assumption was that if you are owner of an organization you are not in members of said organization

WcaleNieWolny:
It it pointless to have the owner in members. He cannot remove himself from the guild nor can he change his own role

PR #322, opened 2023-09-03
fix: weird scrollbar in the UI

kremedev:
fix: #320 
/claim #320

PR #323, opened 2023-09-03
fix: different behaviour between browsers

### LINKED ISSUE #320: Difftent behaviour between browsers
In brave (chromium based) I found a weird scroll bar in the UI
image

In firefox this scrollbar does not exist
image

I tested this in prod, here is the setup I used

image
image

chirag-jn:
/claim #320 
fixes #320

### DIFF
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -84,7 +84,7 @@ const tabs = ref<Tab[]>([
     <div
       id="sidebar"
       ref="sidebar"
-      class="left-0-safe top-0-safe no-scrollbar lg:sidebar-expanded:!w-64 absolute z-40 h-full min-h-screen w-64 flex shrink-0 flex-col overflow-y-scroll bg-slate-800 transition-all duration-200 ease-in-out lg:static lg:left-auto lg:top-auto lg:w-14 lg:translate-x-0 lg:overflow-y-auto 2xl:!w-64"
+      class="left-0-safe top-0-safe no-scrollbar lg:sidebar-expanded:!w-64 absolute z-40 h-full min-h-screen w-64 flex shrink-0 flex-col overflow-y-auto bg-slate-800 transition-all duration-200 ease-in-out lg:static lg:left-auto lg:top-auto lg:w-14 lg:translate-x-0 lg:overflow-y-auto 2xl:!w-64"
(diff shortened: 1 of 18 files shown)


WcaleNieWolny:
I will take a look at this in about 30 minutes

chirag-jn:
Related information: https://www.w3schools.com/css/css_overflow.asp

WcaleNieWolny:
It makes things better, and I would even say that it is good enough. However there still is some differences between chrome and firefox

https://github.com/Cap-go/capgo/assets/50914789/3abfdecf-8db5-458f-87c1-73cdadd3f1d1

Firefox has 2 scrollbars where brave (chromium) only has 1

chirag-jn:
@WcaleNieWolny can you give it another try?

WcaleNieWolny:
On it!

WcaleNieWolny:
It does make the behavior the save in both browsers, so it does resolve the issue but I think that if we are not going to fix this double scroll bar issue then having it only in firefox is better then in both browsers

https://github.com/Cap-go/capgo/assets/50914789/982b47d1-f2fe-446a-b207-2d77bf462294

chirag-jn:
As an alternative, we can either set `height: 100%` or set class as `overflow-y-hidden` on the parent div. What say?

WcaleNieWolny:
> set class as `overflow-y-hidden`

Replace your `overflow-y-auto` with `overflow-y-hidden` and it does work in both browsers.


https://github.com/Cap-go/capgo/assets/50914789/82bfb5fd-5fc0-400a-8c4e-48a620cc9b8a

chirag-jn:
Yep! That's what I am talking about. Made a patch for the same.

rishi-raj-jain:
Way to go, @chirag-jn!

WcaleNieWolny (approved):

riderx:
Hey @chirag-jn thanks for the work, the other PR was ready to merge when this one was having conflict so i used the other one.
I will still tip you for trying

chirag-jn:
Probably the conflict happened because you merged PR #322 before this one.

kremedev:
> Probably the conflict happened because you merged PR #322 before this one.

At least thank him for the tip.

riderx:
You are right,
I went fast, I'm sorry I didn't compare them properly.
Next time I will be more careful

riderx:
Thanks @kremedev all together we will improve

PR #334, opened 2023-09-06
fix: stats landing 

rishi-raj-jain:
fix https://github.com/Cap-go/website/issues/62
/claim https://github.com/Cap-go/website/issues/62

### DIFF
--- a/src/components/dashboard/Usage.vue
+++ b/src/components/dashboard/Usage.vue
@@ -54,22 +54,22 @@ async function getAppStats() {
   const date_id = new Date().toISOString().slice(0, 7)
   if (!main.user)
     return { data: [], error: 'missing user' }
-
+  console.log(props)
   if (props.appId) {
     // console.log('appID', props.appId)
     return supabase
-      .from('app_stats')
+      .from('app_usage')
       .select()
-      .eq('user_id', main.user?.id)
+      // .eq('user_id', main.user?.id)
       .eq('app_id', props.appId)
       .like('date_id', `${date_id}%`)
   }
   else {
     return supabase
-      .from('app_stats')
+      .from('app_usage')
       .select()
-      .eq('user_id', main.user?.id)
-      .like('date_id', `${date_id}%`)
+    // .eq('user_id', main.user?.id)
+    // .like('date_id', `${date_id}%`)
   }
 }
 
@@ -84,36 +84,41 @@ async function getUsages() {
   const { data, error } = await getAppStats()
   if (data && !error) {
     datas.value.mau = Array.from({ length: getDaysInCurrentMonth() }).fill(undefined) as number[]
-    // refactor this line without new Array
-    datas.value.mau
-    = datas.value.storage = Array.from({ length: getDaysInCurrentMonth() }).fill(undefined) as number[]
+    datas.value.storage = Array.from({ length: getDaysInCurrentMonth() }).fill(undefined) as number[]
     datas.value.bandwidth = Array.from({ length: getDaysInCurrentMonth() }).fill(undefined) as number[]
     let currentStorage = 0
-    data.forEach((item: Database['public']['Tables']['app_stats']['Row']) => {
-      if (item.date_id.length > 7) {
-        const dayNumber = Number(item.date_id.slice(8)) - 1
-        if (datas.value.mau[dayNumber])
-          datas.value.mau[dayNumber] += item.devices || 0
-        else
-          datas.value.mau[dayNumber] = item.devices || 0
-        if (datas.value.storage[dayNumber])
-          datas.value.storage[dayNumber] += item.version_size ? bytesToGb(item.version_size) : 0
-        else
-          datas.value.storage[dayNumber] = item.version_size ? bytesToGb(item.version_size) : 0
-        if (datas.value.bandwidth[dayNumber])
+    data.forEach((item: Database['public']['Tables']['app_usage']['Row']) => {
+      if (item.created_at) {
+        const dayNumber = new Date(item.created_at).getDate()
+        if (datas.value.mau[dayNumber]) {
+          datas.value.mau[dayNumber] += item.mau
+        }
+        else {
+          datas.value.mau[dayNumber] = item.mau
+        }
+        if (datas.value.storage[dayNumber]) {
+          datas.value.storage[dayNumber] += item.storage ? bytesToGb(item.storage) : 0
+        }
+        else {
+          datas.value.storage[dayNumber] = item.storage ? bytesToGb(item.storage) : 0
+        }
+        if (datas.value.bandwidth[dayNumber]) {
           datas.value.bandwidth[dayNumber] += item.bandwidth ? bytesToGb(item.bandwidth) : 0
-        else
+        }
+        else {
           datas.value.bandwidth[dayNumber] = item.bandwidth ? bytesToGb(item.bandwidth) : 0
+        }
       }
-      else if (item.date_id.length === 7) {
-        currentStorage += item.version_size ? bytesToGb(item.version_size) : 0
-      }
+      // else if (item.date_id.length === 7) {
+      //   currentStorage += item.version_size ? bytesToGb(item.version_size) : 0
+      // }
     })
     const storageVariance = datas.value.storage.reduce((p, c) => (p + (c || 0)), 0)
     // console.log('storageVariance', storageVariance, currentStorage)
     datas.value.storage[0] = currentStorage - storageVariance
-    if (datas.value.storage[0] < 0)
+    if (datas.value.storage[0] < 0) {
       datas.value.storage[0] = 0
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


WcaleNieWolny (changes requested):
Please lint the code with `pnpm lint --fix`, also how have you tested your changes? I do not see any default data i n `app_usage` and when I ran 

```sql
SELECT update_app_usage(5)
SELECT calculate_daily_app_usage()
SELECT calculate_cycle_usage()
```

I got
```
column "mau" of relation "app_usage" does not exist
--
column "mau" does not exist
--
column app_usage.mau does not exist
```

rishi-raj-jain:
done with the lint

WcaleNieWolny:
I still cannot get this to work in the local env.
image

I restarted my local db with`supabase db reset`

rishi-raj-jain:
try to run the prod one, but I don't see why are you testing that?

rishi-raj-jain:
Log into @riderx's account and then you can see all the data coming in.

WcaleNieWolny:
First of all, how do you expect me to log into his account? Second of all, I tested this function as it was one of the only fns that inserted into app_usage. Perhaps it works in production because ridex changed something but I cannot get this to work in local env. There exists a cron job that relies on this function and right now I do not thing it runs properly.

rishi-raj-jain:
@riderx can you review

@WcaleNieWolny to run the account, just follow the steps I've done and log into @riderx's account using the credentials

WcaleNieWolny:
This is an open source project, so if it does not work in local then neither does it work for people who want to self host capgo. I do not think it is a good idea to just test this in production

rishi-raj-jain:
Replied in Discord

riderx:
Hey @WcaleNieWolny I understand your concern and your reply, just understand my priority right now with this ask was to fix current paying users problem.
So that why I shared my account to @rishi-raj-jain to let him try more easily.
I will fix the issue with local env :)

riderx:
@rishi-raj-jain can merge main in your branch, it should fix local test.
I found a small issue here:
CleanShot 2023-09-07 at 00 44 17@2x
On the arrow, you can see the "predicted data" is not looking like a prediction, it should continue the line of the actual data.
so it seems the prediction is correct, just the scale look not aligned

riderx:
also the first point of the prediction should be the last point of the actual data, not earlier

src/pages/app/package/[package].vue
```diff
@@ -30,19 +35,70 @@ async function loadAppInfo() {
...
-
-    const date_id = new Date().toISOString().slice(0, 7)
     const { data } = await supabase
-      .from('app_stats')
+      .from('stats')
       .select()
       .eq('app_id', id.value)
-      .eq('date_id', date_id)
-      .single()
+      .eq('action', 'set')
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: you need to limit this query with the cycle

src/pages/app/package/[package].vue
```diff
@@ -30,19 +35,70 @@ async function loadAppInfo() {
...
+      .from('channels')
+      .select()
+      .eq('app_id', id.value)
+    if (channelsData)
+      channelsNb.value = channelsData.length
+
+    const { data: devicesData } = await supabase
+      .from('devices')
+      .select()
+      .eq('app_id', id.value)
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: same here put the limit in the query

src/components/dashboard/Usage.vue
```diff
@@ -51,70 +51,89 @@ const allLimits = computed(() => {
...
+
   const date_id = new Date().toISOString().slice(0, 7)
   stats.value = await getTotalStats(main.user?.id, date_id)
 }
+
 async function getUsages() {
+  const currentStorage = 0
   const { data, error } = await getAppStats()
   if (data && !error) {
     datas.value.mau = Array.from({ length: getDaysInCurrentMonth() }).fill(undefined) as number[]
```
riderx: getDaysInCurrentMonth should be modified to get the number of day between the start and end of cycle.
Chart should now display day of cycle not of the month

src/components/dashboard/Usage.vue
```diff
@@ -51,70 +51,89 @@ const allLimits = computed(() => {
...
 async function getAppStats() {
-  const date_id = new Date().toISOString().slice(0, 7)
   if (!main.user)
     return { data: [], error: 'missing user' }
 
   if (props.appId) {
-    // console.log('appID', props.appId)
     return supabase
-      .from('app_stats')
+      .from('app_usage')
```
riderx, line 61: get the app_usage with mode = day, and created_at between the cycle limit.

src/components/dashboard/AppCard.vue
```diff
@@ -93,10 +96,18 @@ async function deleteApp(app: Database['public']['Tables']['apps']['Row']) {
...
+  }
+}
+
 async function loadData() {
   if (!props.channel) {
     devicesNb.value = 0
-
     try {
       const date_id = new Date().toISOString().slice(0, 7)
       const { data } = await supabase
```
riderx: this should be remove app_stats shouldn't be used anymore

src/pages/app/package/[package].vue
```diff
@@ -30,19 +35,70 @@ async function loadAppInfo() {
...
+          }
+          // if not anything of the above, it is false and proceed
+          if (!notContinue)
+            updatesNb.value = updatesNb.value + 1
+        }
+      })
+    }
+    const { data: bundlesData } = await supabase
+      .from('app_versions')
+      .select()
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: limit to deleted = false

rishi-raj-jain:
Screenshot 2023-09-07 at 7 29 16 AM

fixed some of the checks by handling the null case as well.

src/components/dashboard/Usage.vue
rishi-raj-jain: I've added the mode=day filter, but can't get between the cycle limit as right now we need to increment whatever's there in created_at to see some data coming in.

src/components/dashboard/Usage.vue
rishi-raj-jain: resolved

rishi-raj-jain:
@riderx everything including predictions is now fixed

src/pages/app/package/[package].vue
```diff
@@ -31,18 +36,59 @@ async function loadAppInfo() {
...
+    if (cycleStart && cycleEnd) {
+      const { data } = await supabase
+        .from('stats')
+        .select()
+        .eq('app_id', id.value)
+        .eq('action', 'set')
+        .gte('created_at', getConvertedDate2(cycleStart))
+        .lte('created_at', getConvertedDate2(cycleEnd))
+      if (data)
+        updatesNb.value = data.length
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: This don't work above 1k, supabase limit the output, you need to use count:
CleanShot 2023-09-07 at 15 03 28@2x

src/pages/app/package/[package].vue
```diff
@@ -31,18 +36,59 @@ async function loadAppInfo() {
...
+    }
+    else {
+      const { data } = await supabase
+        .from('stats')
+        .select()
+        .eq('app_id', id.value)
+        .eq('action', 'set')
+      if (data)
+        updatesNb.value = data.length
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: same here otherwise some user get wrong numbers:
CleanShot 2023-09-07 at 15 04 06@2x

riderx:
I fixed yesterday storage calculation in backend:
CleanShot 2023-09-07 at 15 08 05@2x
And it seems again the conversion is not good

rishi-raj-jain:
@riderx 

The conversion to storage still followed bytes to Gb while the conversion for bandwidth uses octet to gb as storage wasn't showing anything earlier.

It shall be fixed now.

rishi-raj-jain:
updated the code to get the days if the there's an info in the main store

rishi-raj-jain:
@riderx ready for your review.

rishi-raj-jain:
@riderx ready for your review.

rishi-raj-jain:
@riderx ready for your review.

rishi-raj-jain:
@riderx

This is ready for your review.

riderx:
I think we have issue in shared feature:
CleanShot 2023-09-11 at 14 22 30@2x

src/components/dashboard/AppCard.vue
```diff
@@ -168,11 +196,8 @@ watchEffect(async () => {
       </div>
     </td>
     <td class="w-1/4 p-2">
-      <div v-if="!isLoading && !props.channel" class="text-center">
-        {{ devicesNb }}
-      </div>
-      <div v-else class="text-center">
```
riderx, line 174: here is the code who should stay

rishi-raj-jain: For the shared thing as you shared above?

riderx: yes

src/pages/app/package/[package].vue
```diff
@@ -31,18 +36,59 @@ async function loadAppInfo() {
...
     app.value = dataApp || app.value
 
-    const date_id = new Date().toISOString().slice(0, 7)
-    const { data } = await supabase
-      .from('app_stats')
-      .select()
+    if (cycleStart && cycleEnd) {
+      const { count: statsCount } = await supabase
+        .from('stats')
+        .select('*', { count: 'exact' })
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: You need to use `head: true` to make the request lighter and get only the count not the data

src/pages/app/package/[package].vue
```diff
@@ -31,18 +36,59 @@ async function loadAppInfo() {
...
+        .eq('app_id', id.value)
+        .eq('action', 'set')
+        .gte('created_at', getConvertedDate2(cycleStart))
```
Comment (riderx) [inline src/pages/app/package/[package].vue]: can you factorize this code, by not await directly.
then you can add the `gte` with and if and then await all request at once, it will make the page load faster

pnpm-lock.yaml
```diff
@@ -9851,7 +9851,7 @@ packages:
   /isomorphic-fetch@3.0.0:
     resolution: {integrity: sha512-qvUtwJ3j6qwsF3jLxkZ72qCgjMysPzDfeV240JHiGZsANBYd+EEuu35v7dfrJ9Up0Ak07D7GGSkGhCHTqg/5wA==}
     dependencies:
-      node-fetch: 2.6.11
+      node-fetch: 2.7.0
       whatwg-fetch: 3.6.2
     transitivePeerDependencies:
       - encoding
```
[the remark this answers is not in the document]
riderx: @rishi-raj-jain you get reviewed by your own code ahah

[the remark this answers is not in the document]
rishi-raj-jain: @riderx Hahah yeah! That's the world of code we live in :) This change looks good on node-fetch though.

src/components/LineChartStats.vue
```diff
@@ -7,14 +7,14 @@ import {
...
 import { computed, ref } from 'vue'
 import { Line } from 'vue-chartjs'
 import annotationPlugin from 'chartjs-plugin-annotation'
 import { useI18n } from 'vue-i18n'
 import { isDark } from '~/composables'
 import { getCurrentDayMonth, getDaysInCurrentMonth } from '~/services/date'
+import { useMainStore } from '~/stores/main'
 
 const props = defineProps({
   title: { type: String, default: '' },
```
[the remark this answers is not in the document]
rishi-raj-jain: This is LGTM so good.

PR #356, opened 2023-09-13
Lint backend + CI/CD

WcaleNieWolny:


neo773:
@rishi-raj-jain Can we configure the GPT Bot Action to not say anything if the risk is none will clear a lot of noise.

PR #357, opened 2023-09-14
Fix Pipeline is empty exception

WcaleNieWolny:
Sometimes when device was overwritten redis caching would fail with the following exception
```
[Error] Error: Pipeline is empty
    at Pipeline.exec (https://deno.land/x/upstash_redis@v1.22.0/pkg/pipeline.ts:67:19)
    at RedisUpstashPipeline.flush (file:///home/deno/functions/_utils/redis.ts:29:29)
    at main (file:///home/deno/functions/updates/index.ts:93:14)
    at async Server.#respond (https://deno.land/std@0.200.0/http/server.ts:222:24)
```

This PR makes sure that if  pipeline is empty the `pipeline#exec` function is not called.
I tested my changes, and they resolve this exception

### DIFF
--- a/supabase/functions/_utils/redis.ts
+++ b/supabase/functions/_utils/redis.ts
@@ -1,4 +1,4 @@
-import type { Redis, RedisPipeline } from 'https://deno.land/x/redis@v0.24.0/mod.ts'
+import type { Integer, Redis, RedisPipeline } from 'https://deno.land/x/redis@v0.24.0/mod.ts'
 import { Redis as RedisUpstash } from 'https://deno.land/x/upstash_redis/mod.ts'
 import { connect, parseURL } from 'https://deno.land/x/redis@v0.24.0/mod.ts'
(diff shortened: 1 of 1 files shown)


supabase/functions/_utils/redis.ts
```diff
@@ -42,25 +42,31 @@ class RedisRedisPipeline implements RedisPipelineInterface {
 }
 
 class RedisUpstashPipeline implements RedisPipelineInterface {
+  size: Integer
```
thoro23: I dont think introducing a new size variable and handling it manually is the right way to go here. It is risk-prone because it can be forgotten to be updated when introducing new features. In this case here it also should be reset to 0 in the flush method else the error would still occur when reusing the RedisUpstashPipeline.
If the main goal is to ignore the "Pipeline is empty" error, the exec function should be wrapped in a try/catch and the specific error should be ignored there.
Another approach would be to use the length of the command array on the pipeline instance (see line 231 in https://github.com/upstash/upstash-redis/blob/main/pkg/pipeline.ts).

Issue #277, opened 2023-08-19
Weekly email stats


riderx:
**Describe the bug/issue**
The idea is to do a cron job, who send to active account an email every week to give them stats about they account.
I will give the email content

riderx:
Email content:
```md
```

riderx:
Email content:
```md
Hello Capgo User,

We hope you're having a fantastic week! Here's your weekly dose of Capgo stats, served hot and fresh, just the way you like it!

Updates Dispatched: This week, we sent out a whopping {{ weekly_updates }} updates! That's like delivering [fun comparison]!

Updates Installed: Out of these, {{ weekly_install }} found their new homes in user devices. That's a {{ weekly_install_success }}% success rate! High five!

Oops Moments: Unfortunately, {{ weekly_fail }} updates couldn't make it and tripped on the way. But don't worry, we're already working on their band-aids!

App Openings: Your app was the star of the show, being opened {{ weekly_open }} times this week! That's more popular than [fun comparison]!

Time Spent: Users spent a total of {{ weekly_open_time }} hours on your app. That's equivalent to watching [fun comparison]!

Remember, every update is a step forward, and we're here to make sure your steps are as smooth as possible. Keep rocking the app world, and we'll be back next week with more fun facts!

Stay Updated,
Your Capgo Team
```

Here is the variable content I want to add:
```md
Updates Dispatched:

Little: "That's like delivering a cupcake to every student in a small school!"

Medium: "That's like delivering a pizza to every resident of a small town!"

Big: "That's like delivering a burger to everyone in a big city!"

Updates Installed:

Little: "That's a 90% success rate! Even cats don't land on their feet that often!"

Medium: "That's a success rate higher than the average pass rate of a tough university exam!"

Big: "That's a success rate that even the best basketball players would envy!"

App Openings:

Little: "Your app was opened more times than a popular local bakery's door!"

Medium: "Your app was more popular than the latest episode of a hit TV show!"

Big: "Your app was opened more times than a blockbuster movie on its opening weekend!"

Time Spent:

Little: "That's equivalent to watching the entire 'Finding Nemo' movie!"

Medium: "That's equivalent to watching the entire 'Lord of the Rings' trilogy!"

Big: "That's equivalent to watching the entire 'Game of Thrones' series!"
```

The email will be sent by plunk, the template is already created there, only the sentence variation will be done in code and the stats.

I can find a way to export data to have them in the seed if you need more.

riderx:
[...]

PR #359, opened 2023-09-15
Cron email stats

WcaleNieWolny:
This PR adds #277 

This email is sent every 12:00 on Saturday

Here is the setup I used in plunk:
Template:
```
Hello Capgo User,

We hope you're having a fantastic week! Here's your weekly dose of Capgo stats for the {{app_id}} app, served hot and fresh, just the way you like it!

Updates Dispatched: This week, we sent out a whopping {{weekly_updates}} updates! That's like delivering {{fun_comparison}}!

Updates Installed: Out of these, {{weekly_install}} found their new homes in user devices. That's a {{weekly_install_success}}% success rate! {{fun_comparison_2}}

Oops Moments: Unfortunately, {{weekly_fail}} updates couldn't make it and tripped on the way. But don't worry, we're already working on their band-aids!

App Openings: Your app was the star of the show, being opened {{weekly_open}} times this week! {{fun_comparison_3}}

Remember, every update is a step forward, and we're here to make sure your steps are as smooth as possible. Keep rocking the app world, and we'll be back next week with more fun facts!

Stay Updated,
Your Capgo Team
```

image
image

Here is how this email looks in gmail:
image

/claim #277

### DIFF
--- a/supabase/functions/cron_email/index.ts
+++ b/supabase/functions/cron_email/index.ts
@@ -109,6 +109,9 @@ async function main(_url: URL, headers: BaseHeaders, _method: string, _body: any
       continue
     }
 
+    if (weeklyStats.all_updates === 0)
+      continue
+
     const sucessUpdates = weeklyStats.all_updates - weeklyStats.failed_updates
     if (sucessUpdates < 0) {
       console.error('Cannot send email for app, sucessUpdates < 0', weeklyStats, mapApp)
--- a/supabase/functions/cron_email/index.ts
+++ b/supabase/functions/cron_email/index.ts
@@ -1,5 +1,4 @@
 import { serve } from 'https://deno.land/std@0.199.0/http/server.ts'
-import { map } from 'https://deno.land/x/fonction@v1.6.2/mod.ts'
 import { getEnv, methodJson, sendRes } from '../_utils/utils.ts'
 import type { BaseHeaders } from '../_utils/types.ts'
 import { trackEvent } from '../_utils/plunk.ts'
@@ -8,6 +7,7 @@ import { supabaseAdmin } from '../_utils/supabase.ts'
 // This is required propoably due to this https://github.com/supabase/postgrest-js/issues/408
 interface AppWithUser {
   app_id: string
+  name: string | null
   user_id: {
     email: string
   }
@@ -19,12 +19,53 @@ const thresholds = {
     1000,
     10000,
   ],
-  fail_rate: [
+  failRate: [
     0.80,
     0.90,
     0.95,
   ],
+  appOpen: [
+    500,
+    1500,
+    5000,
+  ],
+}
 
+const funComparisons = {
+  updates: [
+    'That\'s like delivering a cupcake to every student in a small school!',
+    'That\'s like delivering a pizza to every resident of a small town!',
+    'That\'s like delivering a burger to everyone in a big city!',
+  ],
+  failRate: [
+    'That\'s above 80% success rate! Even cats don\'t land on their feet that often!',
+    'That\'s a success rate higher than the average pass rate of a tough university exam!',
+    'That\'s a success rate that even the best basketball players would envy!',
+  ],
+  appOpen: [
+    'Your app was opened more times than a popular local bakery\'s door!',
+    'Your app was more popular than the latest episode of a hit TV show!',
+    'Your app was opened more times than a blockbuster movie on its opening weekend!',
+  ],
+}
+
+function getFunComparison(comparison: 'updates' | 'failRate' | 'appOpen', stat: number): string {
+  console.log('stat', stat)
+  const thresholdsForComparisons = thresholds[comparison]
+  const index = thresholdsForComparisons.map((threshold, index) => {
+    if (threshold >= stat)
+      return index
+    else if ((index === 2 && stat >= threshold))
+      return 2 // Last index
+
+    return undefined
+  }).find(i => i !== undefined)
+
+  if (index === undefined || index >= 3)
+    throw new Error(`Cannot find index for fun comparison, ${index}`)
+
+  console.log('final i', index)
+  return funComparisons[comparison][index]
 }
 
 async function main(url: URL, headers: BaseHeaders, method: string, body: any) {
@@ -39,12 +80,15 @@ async function main(url: URL, headers: BaseHeaders, method: string, body: any) {
     return sendRes({ message: 'Fail Authorization', authorizationSecret }, 400)
   }
 
+  console.log('body', ((1 / 8) * 100).toFixed(2))
+
   const supabase = await supabaseAdmin()
 
   const { data: apps, error: appsError } = await supabase
     .from('apps')
     .select(`
       app_id,
+      name,
       user_id ( id )
     `)
 
@@ -70,6 +114,27 @@ async function main(url: URL, headers: BaseHeaders, method: string, body: any) {
 
     console.log('weeklyStats', weeklyStats)
 
+    if (!weeklyStats || generateStatsError) {
+      console.error('Cannot send email for app', mapApp.app_id, generateStatsError, mapApp.user_id.email)
+      continue
+    }
+
+    const sucessUpdates = weeklyStats.all_updates - weeklyStats.failed_updates
+    if (sucessUpdates < 0) {
+      console.error('Cannot send email for app, sucessUpdates < 0', weeklyStats, mapApp)
+      continue
+    }
+
+    const successPercantage = (sucessUpdates / weeklyStats.all_updates)
+
+    const metadata = {
+      app_id: mapApp.name ?? mapApp.app_id,
+      weekly_updates: weeklyStats.all_updates,
+      fun_comparison: getFunComparison('updates', weeklyStats.all_updates),
+      weekly_install: sucessUpdates,
+      // weekly_install_success: successPercantage.t
+    }
+
     if (generateStatsError) {
       console.error('error', generateStatsError)
       return sendRes()
@@ -85,8 +150,6 @@ async function main(url: URL, headers: BaseHeaders, method: string, body: any) {
     weekly_fail: '12',
     weekly_open: '100',
     fun_comparison_2: 'idk',
-    weekly_open_time: '10min',
-    fun_comparison_3: '??',
   }
 
   // await trackEvent('isupermichael007@gmail.com', data, 'cron-stats')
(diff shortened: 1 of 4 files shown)


supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  failRate: [
+    0.80,
+    0.90,
+    0.95,
+  ],
+  appOpen: [
+    500,
+    1500,
+    5000,
+  ],
```
wesleymatosdev: There's any variable name that we could use to know what these magic numbers are? I mean, is not clear if it's ms, seconds or any other thing
```suggestion
  updates: [
    100,
    1000,
    10000,
  ],
  failRate: [
    0.80,
    0.90,
    0.95,
  ],
  appOpen: [
    500,
    1500,
    5000,
  ],
```

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,152 @@
...
+    'That\'s a success rate that even the best basketball players would envy!',
+  ],
+  appOpen: [
+    'Your app was opened more times than a popular local bakery\'s door!',
+    'Your app was more popular than the latest episode of a hit TV show!',
+    'Your app was opened more times than a blockbuster movie on its opening weekend!',
+  ],
+}
+
+function getFunComparison(comparison: 'updates' | 'failRate' | 'appOpen', stat: number): string {
```
wesleymatosdev: you could do something like 
```suggestion
function getFunComparison(comparison: ComparisonEnum, stat: number): string {
```

or even 
```suggestion
function getFunComparison(comparison: keyof funComparisons, stat: number): string {
```

This would likely improve readability + avoid code repetition between key names

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,152 @@
...
+  const thresholdsForComparisons = thresholds[comparison]
+  const index = thresholdsForComparisons.map((threshold, index) => {
+    if (threshold >= stat)
+      return index
+    else if ((index === 2 && stat >= threshold))
+      return 2 // Last index
```
ologbonowiwi, line 58:
```suggestion
      return thresholdsForComparisons.length - 1
```

supabase/functions/cron_email/index.ts
WcaleNieWolny: I will add comments

supabase/functions/cron_email/index.ts
WcaleNieWolny: Keyof will not work
image

As for enum, I do not think we really need it, this `funComparisons` are only used once (in this `getFunComparison` fn)

supabase/functions/cron_email/index.ts
WcaleNieWolny: Comments added [here](https://github.com/Cap-go/capgo/pull/359/commits/f3d138508065d6846e2790301eb663e1ec8ff16d)

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    return undefined
+  }).find(i => i !== undefined)
```
wesleymatosdev: Instead of map + find, we could use [`findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) here

supabase/functions/cron_email/index.ts
wesleymatosdev: I messed up here, the right would be `keyof typeof funComparisons`; can you check if it works?

WcaleNieWolny: Yes it does, thanks for this suggestion

supabase/functions/cron_email/index.ts
WcaleNieWolny: Changed [here](https://github.com/Cap-go/capgo/pull/359/commits/6092f16622d27cf99f4ef74fee974bbd583e5c94)

supabase/functions/cron_email/index.ts
WcaleNieWolny: Changed [here](https://github.com/Cap-go/capgo/pull/359/commits/6092f16622d27cf99f4ef74fee974bbd583e5c94)

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,156 @@
...
+function getFunComparison(comparison: keyof typeof funComparisons, stat: number): string {
+  const thresholdsForComparisons = thresholds[comparison]
+  const index = thresholdsForComparisons.findIndex((threshold, index) => {
+    if (threshold >= stat)
+      return true
+    else if ((index === 2 && stat >= threshold))
+      return true // Last index
+
+    return false
+  })
```
wesleymatosdev: I know that I already gave a suggestion here, but I guess this would give the same result

```suggestion
  const index = thresholdsForComparisons.findIndex((threshold) => threshold >= stat)
```

WcaleNieWolny: No it will not, for example if you had 10 000 app openings this week then the last threshold would we `5000` and `10 000` is bigger then `5000` thus index would not be found

wesleymatosdev: oh, I didn't see the second one was inverted. sorry

we could simplify it and clarify what these conditions mean (I'm pretty new to the project, and the conditions don't tell me what this code does.)

what if we can do something like
```suggestion
  const index = thresholdsForComparisons.findIndex((threshold, index) => {
    const firstCondition = threshold >= stat;
    const secondCondition = index === 2 && stat >= threshold;

    return firstCondition || secondCondition;
  })
```

and change the variables name explicitly to something like `thresholdIsSomething`, `thresholdMatchXCondition`, etc.; what do you think?

WcaleNieWolny: Rename `firstCondition` to `thresholdGreaterThenStat` and to `secondCondition` to `statGreaterOrEqualThresholdAndLastIndex`, suggest changes and I will commit this

wesleymatosdev:
```suggestion
  const index = thresholdsForComparisons.findIndex((threshold, index) => {
    const thresholdGreaterThenStat = threshold >= stat;
    const statGreaterOrEqualThresholdAndLastIndex = index === 2 && stat >= threshold;

    return thresholdGreaterThenStat || statGreaterOrEqualThresholdAndLastIndex;
  })
```

thanks for accepting my suggestion 😄

WcaleNieWolny:
@ologbonowiwi If you do not have more suggestions could you please approve these changes?

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Check what treshold does the stat qualify for and return the fun comparison
+function getFunComparison(comparison: keyof typeof funComparisons, stat: number): string {
+  const thresholdsForComparisons = thresholds[comparison]
+  const index = thresholdsForComparisons.findIndex((threshold, index) => {
+    const thresholdGreaterThenStat = threshold >= stat;
+    const statGreaterOrEqualThresholdAndLastIndex = index === 2 && stat >= threshold;
+
+    return thresholdGreaterThenStat || statGreaterOrEqualThresholdAndLastIndex;
```
wesleymatosdev:
```suggestion
    const lastIndexAndStatGreaterOrEqualThreshold = index === 2 && stat >= threshold;

    return thresholdGreaterThenStat || lastIndexAndStatGreaterOrEqualThreshold;
```

Seeing now, makes more sense inverted

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  return funComparisons[comparison][index]
+}
+
+async function main(_url: URL, headers: BaseHeaders, _method: string, _body: any) {
+  const API_SECRET = getEnv('API_SECRET')
+  const authorizationSecret = headers.apisecret
+  if (!authorizationSecret)
+    return sendRes({ status: 'Cannot find authorization secret' }, 400)
+
+  if (!authorizationSecret || !API_SECRET || authorizationSecret !== API_SECRET) {
```
wesleymatosdev:
```suggestion
  if (!API_SECRET || authorizationSecret !== API_SECRET) {
```

we don't need to check `authorizationSecret` because it was already checked on 74

supabase/functions/cron_email/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    console.error('Fail Authorization', { authorizationSecret, API_SECRET })
```
wesleymatosdev: Can we log `API_SECRET` like this? Does this have no change of leaking?

ologbonowiwi (approved):

ologbonowiwi:
I left more comments, but overall, LGTM

supabase/functions/cron_email/index.ts
WcaleNieWolny: I have no idea. I just copied the template

WcaleNieWolny: I am not touching the security of this, if there is a bug it there then it is already in production

riderx: @ologbonowiwi 😖, you are right it was a local test who ended up in prod

WcaleNieWolny: Should I remove the entire `console.error`?

wesleymatosdev: @riderx I removed the log from `supabase/functions/cron_good_plan/index.ts` on #368. I'm not sure if we should also remove the `authorizationSecret` please let me know if so

supabase/functions/cron_email/index.ts
WcaleNieWolny: Fixed in latest commit

supabase/functions/cron_email/index.ts
WcaleNieWolny: Fixed this in `cron_email`

riderx:
@WcaleNieWolny thanks for the work, why the no deploy file is present in the folder of the cron function ?
Do this function. will execute fast enough at scale ?
otherwise, we can make it as background function in Netlify

WcaleNieWolny:
When I started working on this these "background" functions were not a thing. This function should be quite fast and it scales in O(N) where N is the number of apps. If we assume that 1 app takes 100ms this fn will likely time out as the timeout in Postgress as the timeout is 15 seconds. These background functions were added after I resolved a merge conflict and I am happy to implement this as a background fn

riderx:
@WcaleNieWolny thanks, background function are in use since a year and half, the way to call them has been simplified a week ago.
I can do the background things it's not the hard part, I believe the issue is more about the timeout you mention.
Can you switch to a query who run by user, and we loop on user list, that way we should never reach timeout right ?

WcaleNieWolny:
The timeout for netlify background is 15 mins. I do not think we will reach it

Issue #367, opened 2023-09-17
Make a script to backup logs (stats table) and delete from db logs older than X days


riderx:
The goal is to make the DB stop to be bloated.
Currently, the Device ID list and logs are making the Db slow the idea is to make a script who table a table in parameter and can delete entry older than x days.

The date of creation and updated_at should be checked to be bigger than limit
This would be run every day to keep the DB cost low.

This backup has to be store in R2.

This script has to run in GitHub Action

riderx:
i made it myself

riderx:
/tip 20 @ayusham001 thanks for trying, i made shell script who conect direct to pg behind supabase

PR #369, opened 2023-09-17
script to backup logs

ayusham001:
/claim #367

### DIFF
--- a/.github/workflows/backup.yml
+++ b/.github/workflows/backup.yml
@@ -1,0 +1,40 @@
+name: Backup and Delete Old Entries
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  backup_and_delete:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Set up Deno
+      uses: denolib/setup-deno@v2
+      with:
+        deno-version: '1.16.2' 
+        
+    - name: Install dependencies
+      run: deno cache --unstable --lock=import_map.json --import-map=import_map.json script.ts
+
+    - name: Execute backup script
+      run: deno run --unstable -A script.ts
+      env:
+        SUPABASE_URL: ${{ secrets.SUPABASE_URL }}
+        SUPABASE_SERVICE_ROLE_KEY: ${{ secrets.SUPABASE_SERVICE_ROLE_KEY }}
+        R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
+        R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+        R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+        S3_ENDPOINT: ${{ secrets.S3_ENDPOINT }}
+        S3_REGION: ${{ secrets.S3_REGION }}
+        S3_PORT: ${{ secrets.S3_PORT }}
+        S3_SSL: ${{ secrets.S3_SSL }}
+
+    - name: Upload backup to GitHub Artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: backup
+        path: backup
--- a/supabase/functions/_script/backup_and_cleanup.ts
+++ b/supabase/functions/_script/backup_and_cleanup.ts
@@ -28,7 +28,6 @@ function useSupabase() {
 async function backupAndDeleteOldEntries() {
   const currentDate = new Date();
   const backupFilenames: string[] = [];
-  const recordsToDelete: Database['public']['Tables']['app_versions']['Row'][] = [];
 
   try {
     if (!await fs.exists(backupFolder)) {
@@ -43,12 +42,15 @@ async function backupAndDeleteOldEntries() {
 
     while (hasMoreData) {
       const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000);
+      const query = `
+        SELECT * FROM app_versions
+        WHERE updated_at < '${cutoffDate.toISOString()}'
+        OFFSET ${offset}
+        LIMIT ${pageSize}
+      `;
 
       const { data: oldEntries, error: queryError } = await supabase
-        .from<Database['public']['Tables']['app_versions']['Row']>('app_versions')
-        .select()
-        .lt('updated_at', cutoffDate.toISOString())
-        .range(offset, offset + pageSize - 1);
+        .rpc('exec', { sql: query });
 
       if (queryError) {
         console.error('Error querying the Supabase table:', queryError.message);
@@ -68,8 +70,6 @@ async function backupAndDeleteOldEntries() {
         await r2.upload(backupPath, new TextEncoder().encode(backupData));
         console.log(`Backup saved to R2: ${backupPath}`);
         backupFilenames.push(backupFilename);
-
-        recordsToDelete.push(...oldEntries);
       } catch (backupError) {
         console.error('Error saving backup to R2:', backupError);
       }
@@ -77,17 +77,16 @@ async function backupAndDeleteOldEntries() {
       offset += pageSize;
     }
 
-    if (recordsToDelete.length > 0) {
-      const { error: deleteError } = await supabase
-        .from('app_versions')
-        .delete()
-        .in('id', recordsToDelete.map(record => record.id));
+    const deleteQuery = `
+      DELETE FROM app_versions
+      WHERE updated_at < '${cutoffDate.toISOString()}'
+    `;
+    const { error: deleteError } = await supabase.rpc('exec', { sql: deleteQuery });
 
-      if (deleteError) {
-        console.error('Error deleting entries from the Supabase table:', deleteError.message);
-      } else {
-        console.log('Deletion completed successfully.');
-      }
+    if (deleteError) {
+      console.error('Error deleting entries from the Supabase table:', deleteError.message);
+    } else {
+      console.log('Deletion completed successfully.');
     }
   } catch (e) {
     console.error('An error occurred:', e.message);
--- a/supabase/functions/_script/backup_and_cleanup.ts
+++ b/supabase/functions/_script/backup_and_cleanup.ts
@@ -1,0 +1,64 @@
+import { createClient } from 'https://esm.sh/@supabase/supabase-js@^2.2.3';
+import 'https://deno.land/x/dotenv/load.ts';
+import { r2 } from '../_utils/r2.js';
+import type { Database } from '../_utils/supabase.types.ts';
+
+const supabaseUrl = Deno.env.get('SUPABASE_URL') || '***';
+const supabaseServiceRoleKey = Deno.env.get('SUPABASE_SERVICE_ROLE_KEY') || '***';
+const backupFolder = 'backup'; 
+
+function useSupabase() {
+  const options = {
+    auth: {
+      autoRefreshToken: true,
+      persistSession: true,
+      detectSessionInUrl: false,
+    },
+  };
+  return createClient<Database>(supabaseUrl, supabaseServiceRoleKey, options);
+}
+
+async function backupAndDeleteOldEntries() {
+  const supabase = useSupabase();
+  const daysToKeep = 30; 
+
+  const currentDate = new Date();
+  const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000);
+
+  const { data: oldEntries, error: queryError } = await supabase
+    .from<Database['public']['Tables']['app_versions']['Row']>('app_versions')
+    .select()
+    .lt('updated_at', cutoffDate.toISOString());
+
+  if (queryError) {
+    console.error('Error querying the Supabase table:', queryError.message);
+    return;
+  }
+
+  const backupData = JSON.stringify(oldEntries);
+  const backupFilename = `backup-${currentDate.toISOString()}.json`;
+  const backupPath = `${backupFolder}/${backupFilename}`;
+
+  try {
+    await r2.upload(backupPath, new TextEncoder().encode(backupData));
+    console.log(`Backup saved to R2: ${backupPath}`);
+  } catch (backupError) {
+    console.error('Error saving backup to R2:', backupError);
+  }
+
+  const { error: deleteError } = await supabase
+    .from('app_versions')
+    .delete()
+    .lt('updated_at', cutoffDate.toISOString());
+
+  if (deleteError) {
+    console.error('Error deleting entries from the Supabase table:', deleteError.message);
+  } else {
+    console.log('Deletion completed successfully.');
+  }
+}
+
+backupAndDeleteOldEntries().catch((error) => {
+  console.error('An error occurred:', error.message);
+  Deno.exit(1);
+});
--- a/supabase/functions/_script/backup_and_cleanup.ts
+++ b/supabase/functions/_script/backup_and_cleanup.ts
@@ -26,6 +26,10 @@ function useSupabase() {
 }
 
 async function backupAndDeleteOldEntries() {
+  const currentDate = new Date();
+  const backupFilenames: string[] = [];
+  const recordsToDelete: Database['public']['Tables']['app_versions']['Row'][] = [];
+
   try {
     if (!await fs.exists(backupFolder)) {
       await fs.ensureDir(backupFolder);
@@ -34,16 +38,12 @@ async function backupAndDeleteOldEntries() {
     const supabase = useSupabase();
     const daysToKeep = 30;
     const pageSize = 1000;
-
-    const currentDate = new Date();
-    const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000);
-
     let offset = 0;
     let hasMoreData = true;
-    let fileNumber = 0;
-    let recordsToDelete: Database['public']['Tables']['app_versions']['Row'][] = []; 
 
     while (hasMoreData) {
+      const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000);
+
       const { data: oldEntries, error: queryError } = await supabase
         .from<Database['public']['Tables']['app_versions']['Row']>('app_versions')
         .select()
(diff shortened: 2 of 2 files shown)


rishi-raj-jain (changes requested):
@ayusham001 Can you add a GitHub Action Workflow as well (as per https://github.com/Cap-go/capgo/issues/367#issue-1899787045)?

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,64 @@
...
+    .select()
+    .lt('updated_at', cutoffDate.toISOString());
+
+  if (queryError) {
+    console.error('Error querying the Supabase table:', queryError.message);
+    return;
+  }
+
+  const backupData = JSON.stringify(oldEntries);
+  const backupFilename = `backup-${currentDate.toISOString()}.json`;
```
rishi-raj-jain: can update this to have the pagination numbers: say first 1000 records will be backed up as:

`backup-date-0-1000.json`

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,64 @@
...
+  const backupPath = `${backupFolder}/${backupFilename}`;
```
rishi-raj-jain: can you use path.join instead?

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,64 @@
...
+    await r2.upload(backupPath, new TextEncoder().encode(backupData));
+    console.log(`Backup saved to R2: ${backupPath}`);
+  } catch (backupError) {
+    console.error('Error saving backup to R2:', backupError);
+  }
+
+  const { error: deleteError } = await supabase
+    .from('app_versions')
+    .delete()
+    .lt('updated_at', cutoffDate.toISOString());
```
rishi-raj-jain: this should occur only in the case the r2 upload completes fine, so you might want to put this after the code in try that says

```
console.log(`Backup saved to R2: ${backupPath}`);
```

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (deleteError) {
+    console.error('Error deleting entries from the Supabase table:', deleteError.message);
+  } else {
+    console.log('Deletion completed successfully.');
+  }
+}
```
rishi-raj-jain: again this shall be in the scope of the try block after the r2.upload completes fine

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,64 @@
...
+
+backupAndDeleteOldEntries().catch((error) => {
+  console.error('An error occurred:', error.message);
+  Deno.exit(1);
+});
```
rishi-raj-jain: in which event do you see the catch occuring, given that all the operations that could through are already scoped in try/catch

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,78 @@
...
+      await r2.upload(backupPath, new TextEncoder().encode(backupData));
+      console.log(`Backup saved to R2: ${backupPath}`);
+    } catch (backupError) {
+      console.error('Error saving backup to R2:', backupError);
+    }
+
+    const { error: deleteError } = await supabase
+      .from('app_versions')
+      .delete()
+      .lt('updated_at', cutoffDate.toISOString());
```
akhill10: If there is an error in uploading the backup data to R2. The delete query still executes, if it happens then data is lost forever.

It should not delete the data from the database. Instead, it should throw an error and the job should stop executing and retry if necessary.

akhill10 (changes requested):

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: yeah, I've commented the same in https://github.com/Cap-go/capgo/pull/369#:~:text=occur%20only%20in-,the,-case%20the%20r2

ayusham001:
Hi @rishi-raj-jain and @akhill10, I have made the necessary changes. Please take a look at it.

rishi-raj-jain:
@ayusham001 can you resolve the comments from above you think are done? I still see some of them are not resolved after your latest commit.

ayusham001:
> @ayusham001 can you resolve the comments from above you think are done? I still see some of them are not resolved after your latest commit.
Which one are you talking about? I made some changes in the recent one.

ayusham001:
> @ayusham001 can you resolve the comments from above you think are done? I still see some of them are not resolved after your latest commit.


Which one are you talking about? I made some changes in the recent one

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,89 @@
...
+        console.log(`Backup saved to R2: ${backupPath}`);
+      } catch (backupError) {
+        console.error('Error saving backup to R2:', backupError);
+        throw new Error('Backup to R2 failed. Stopping execution to prevent data loss.');
+      }
+
+      const { error: deleteError } = await supabase
+        .from('app_versions')
+        .delete()
+        .lt('updated_at', cutoffDate.toISOString());
```
rishi-raj-jain: this in a while loop, so this operation is occuring multiple times while the result is same. Might want to take this out from the loop and run only if the whole loop works as expected / track number of records to delete.

rishi-raj-jain:
Here are the summary of changes that I think are still pending:

1.
Screenshot 2023-09-18 at 1 31 50 PM

2.
Screenshot 2023-09-18 at 1 33 56 PM

3.
Screenshot 2023-09-18 at 1 34 12 PM

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  } catch (e) {
+    console.error('An error occurred:', e.message);
+  }
+}
+
+backupAndDeleteOldEntries().catch((error) => {
+  console.error('An error occurred:', error.message);
+  Deno.exit(1);
+});
```
akhill10: This catch never gets caught as you're already using the catching in the `backupAndDeleteOldEntries` function. Remove the try catch from  `backupAndDeleteOldEntries`
```suggestion
}

backupAndDeleteOldEntries().then(() => 'Backup logs executed successfully').catch((error) => {
  console.error('An error occurred:', error.message);
  Deno.exit(1);
});
```

akhill10 (changes requested):

akhill10:
@rishi-raj-jain It looks like you haven't submitted your review as it still shows `pending` in you're screenshots.

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: Even this won't execute as everything in the function itself is wrapped under try & catch.

Can just replace with backupAndDeleteOldEntries().

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,86 @@
+import { createClient } from 'https://esm.sh/@supabase/supabase-js@^2.2.3';
+import 'https://deno.land/x/dotenv/load.ts';
+import { r2 } from '../_utils/r2';
+import type { Database } from '../_utils/supabase.types.ts';
+import * as fs from 'https://deno.land/std/fs/mod.ts';
+
+const supabaseUrl = Deno.env.get('SUPABASE_URL') || '***';
+const supabaseServiceRoleKey = Deno.env.get('SUPABASE_SERVICE_ROLE_KEY') || '***';
```
rishi-raj-jain: if these are not present, the script shall exit and not be marked as "***", right?

rishi-raj-jain:
@ayusham001 I've submitted my review now. Thanks for informing!

ayusham001:
@rishi-raj-jain Please have a look at it.

ayusham001:
Hi @riderx, I have made the changes that have been suggested
please have a look at it.

rishi-raj-jain:
@ayusham001 taking a look right now

riderx:
@ayusham001 did you evaluate to do it in SQL directly ? I'm afraid to bloat the db with the loop query that way

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,98 @@
...
+      const backupPath = path.join(backupFolder, backupFilename);
+
+      try {
+        await r2.upload(backupPath, new TextEncoder().encode(backupData));
+        console.log(`Backup saved to R2: ${backupPath}`);
+        recordsToDelete = recordsToDelete.concat(oldEntries);
+      } catch (backupError) {
+        console.error('Error saving backup to R2:', backupError);
+        throw new Error('Backup to R2 failed. Stopping execution to prevent data loss.');
+      }
```
rishi-raj-jain: I think the last thing to optimize is that the case:

- The loop ran fine for the first iteration, now recordsToDelete has some X records to be deleted.
- The loop errored for the second iteration, which now throws a new Error in catch and terminates the whole script.
- The recordsToDelete that's outside the loop set to be deleted won't fire, and while the original X items are backed up, aren't deleted from the supabase.
- When the script runs again, they are backed up again leading to duplicate calls.

rishi-raj-jain:
I've left a comment to be addressed ^^

ayusham001:
@rishi-raj-jain please have a look at it:)

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,97 @@
...
+      persistSession: true,
+      detectSessionInUrl: false,
+    },
+  };
+  return createClient<Database>(supabaseUrl, supabaseServiceRoleKey, options);
+}
+
+async function backupAndDeleteOldEntries() {
+  const currentDate = new Date();
+  const backupFilenames: string[] = [];
```
rishi-raj-jain, line 30: What's this used for?

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,97 @@
...
+
+      const { data: oldEntries, error: queryError } = await supabase
+        .from<Database['public']['Tables']['app_versions']['Row']>('app_versions')
+        .select()
+        .lt('updated_at', cutoffDate.toISOString())
+        .range(offset, offset + pageSize - 1);
+
+      if (queryError) {
+        console.error('Error querying the Supabase table:', queryError.message);
+        break;
```
rishi-raj-jain, line 59: Do we want to just skip this iteration? and not stop the processing for all the entries?

rishi-raj-jain (changes requested):

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: I mean `backupFilenames`

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: So in here if one of the processing fails the loop doesn't proceed forward, correct?

ayusham001: Do we need to skip the current iteration with an error but continue processing the remaining records?

rishi-raj-jain: I think @riderx is the one to make that call here, I'd say that you can still process the rest but yeah not sure if even those would be succesful.

supabase/functions/_script/backup_and_cleanup.ts
ayusham001: it is used to keep track of the filenames of the successfully backed-up data.

rishi-raj-jain: but is it logged or shown somewhere? I just see an array object being pushed items to..

rishi-raj-jain: What you could do is something like at the end when everything is done:

console.table({ name: backupFilenames, done: ['X', 'tick', ...]  }) based on each backupFileName

but this is totally optional and you can remove it as well if not showing up anywhere

ayusham001: It accumulates successful backup filenames for potential future use, but it's not currently logged or displayed.

supabase/functions/_script/backup_and_cleanup.ts
riderx: As said earlier, I'm not sure the Supabase SDK is the right approach for this. Pure SQL could be better.
For the answer if anything fail the script should stop

rishi-raj-jain: Awesome, @ayusham001 you can resolve this but shall edit to use SQL query instead.

ayusham001:
@riderx I have used SQL query instead of Supabase SDK.
Please review it:)

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: Awesome, feel free to remove.

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      offset += pageSize;
+    }
+
+    const deleteQuery = `
+      DELETE FROM app_versions
+      WHERE updated_at < '${cutoffDate.toISOString()}'
+    `;
+    const { error: deleteError } = await supabase.rpc('exec', { sql: deleteQuery });
```
rishi-raj-jain: if some instance of the script backing up failed, this'd still be triggered as the entire operation is not cancelled. Also, the LIMIT and offset are not used to delete the records (like it was for dates in before)

rishi-raj-jain: this'd end up deleting the data from database even if not backed up.

ayusham001: I have updated it in my recent commit.

rishi-raj-jain:
the lint test still fails as it's not checking out the ref from the latest commit

rishi-raj-jain:
added ref commit, this shall fix i

supabase/functions/_script/backup_and_cleanup.ts
ayusham001: Can I mark it as resolved?

rishi-raj-jain: feel free to!

.github/workflows/backup.yml
```diff
@@ -0,0 +1,40 @@
+name: Backup and Delete Old Entries
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
rishi-raj-jain, line 5: @riderx do you want this as a cron job?

rishi-raj-jain: or something that you can click on and trigger?

supabase/functions/_script/backup_and_cleanup.ts
ayusham001: Thank you!

supabase/functions/_script/backup_and_cleanup.ts
rishi-raj-jain: @ayusham001 can you close this conversation as well?

.github/workflows/backup.yml
rishi-raj-jain: Pinged Martin on Discord.

akhill10:
> @ayusham001 did you evaluate to do it in SQL directly ? I'm afraid to bloat the db with the loop query that way

I think @riderx was pointing out that, instead of using supabase SDK. Create a SQL function in postgres db that executes this logic in directly maybe with the help of storage extensions.

.github/workflows/backup.yml
riderx: as cron job is good

rishi-raj-jain: @ayusham001 feel free to close this review then as well. great work!

rishi-raj-jain:
> I think @riderx was pointing out that, instead of using supabase SDK. Create a SQL function in postgres db that executes this logic directly maybe with the help of storage extensions.

How is this different from using Supabase SQL rpc?

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,103 @@
...
+    const supabase = useSupabase();
+    const daysToKeep = 30;
+    let offset = 0;
+    let hasMoreData = true;
+    
+    const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000); // Declare cutoffDate outside the loop
+
+    while (hasMoreData) {
+      const query = `
+        SELECT * FROM app_versions
```
riderx: i think there issue here.
The table we talk about is stats and devices not app_versions

ayusham001: @riderx  I have updated the code Please have a look.

rishi-raj-jain:
LGTM

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,111 @@
...
+    const daysToKeep = 30;
+    let offset = 0;
+    let hasMoreData = true;
+
+    const cutoffDate = new Date(currentDate.getTime() - daysToKeep * 24 * 60 * 60 * 1000); // Declare cutoffDate outside the loop
+
+    while (hasMoreData) {
+      const query = `
+        SELECT * FROM ${tableName}
+        WHERE updated_at < '${cutoffDate.toISOString()}'
```
riderx, line 49: the only way this to not fail would be to set the timeout for the session, but not sure we can do it

ayusham001: IG setting a session timeout is not feasible in this script. Proper error handling and retry mechanisms should suffice.

riderx: @ayusham001 i understand your point but you have to understand the current table:
CleanShot 2023-09-27 at 16 25 29@2x
This request in loop will take too much CPU for too long that will break prod

ayusham001: Thank you for information.
It will indeed put a lot of overhead, so should i put the timeout function right away or not?

ayusham001: @rishi-raj-jain your suggestions on this?

rishi-raj-jain: @ayusham001 why is adding a session timeout not possible in this file?

rishi-raj-jain: You can just make something like:

when it's in while (hasMoreData) just add

await new Promise(r => setTimeout(r, 3000))

as the next line to process, that shall do it, correct?

ayusham001: Yes, just asked for a suggestion
M on it.

rishi-raj-jain: > Yes, just asked for a suggestion

Oh yeah that's fine, I just meant to ask simply no other intentions. Pardon if the framing seemed not apt 🙏🏻

rishi-raj-jain: Also, appreciate the patience you have man, you've been editing this mutliple times now.

ayusham001: No I totally get the concern.
Also, I have pushed the changes

ayusham001: Thank you @rishi-raj-jain Grateful for the compliment, it's just part of the work.

rishi-raj-jain:
LGTM

supabase/functions/_script/backup_and_cleanup.ts
```diff
@@ -0,0 +1,112 @@
...
+      await new Promise(r => setTimeout(r, 3000))
+      const query = `
+        SELECT * FROM ${tableName}
+        WHERE updated_at < '${cutoffDate.toISOString()}'
+        OFFSET ${offset}
+        LIMIT ${pageSize}
+      `;
+
+      const { data: oldEntries, error: queryError } = await supabase
+        .rpc('exec', { sql: query });
```
riderx, line 55: By the way you again didn't tried your code, this doesn't exist in supabase

Issue #384, opened 2023-09-19
Protect Update and stats endpoint from bad actors


riderx:
![CleanShot 2023-09-19 at 19 53 37@2x](https://github.com/Cap-go/capgo/assets/4084527/2ede4391-904d-4d71-be2f-7756f98206f7)
Our redis got bloated of fake date from update enpoind.
A good thing could be to update zod check in update enpoint to be less permissive and force correct type.
And make stats endpoint to use the same logic

rishi-raj-jain:
Might also want to see the key and evaluate if it doesn't contain any $ or {}

riderx:
my ask is to use zod package to validate the fields, and no only the appId but all fields

ayewo:
Ok but Zod is already being used to parse fields in the /updates endpoint but not in the /stats endpoint.

ayewo:
Wait, the screenshot above is for the Redis keys created by the /updates endpoint, right?

Even with some Zod parsing in place, the fake data still got through … unless regex is used to check the content as I have done.

WcaleNieWolny:
> Even with some Zod parsing in place

The parsing is not sufficient, zod just check if app is is a string

riderx:
yes i believe it's required to add regex in zod as well, my point is to have only one way to check to make it all with zod

ayewo:
> yes i believe it's required to add regex in zod as well, my point is to have only one way to check to make it all with zod

I've made validation more strict using Zod.

Please see commit [`6402f3c`](https://github.com/Cap-go/capgo/pull/385/commits/6402f3c111e8593a9faa07a5a89a5ae2382d5cb4).

ayewo:
@riderx bump.

PR #385, opened 2023-09-19
fix: ignore invalid app_ids in /update and /stats requests

ayewo:
Fix to /claim https://github.com/Cap-go/capgo/issues/384


riderx:
Hey @ayewo thanks a lot for the fix, my ask was to use zod package to validate the fields, and no only the appId but all fields

WcaleNieWolny:
I do not think you can validate 
all fields. I do not see a way to implement that. Validating only app is is okay as far as I am concerned.

The only things are missing are tests for this. @ayewo Can you please add some tests for this? (See the latest merged PR with backend tests)

This test should test the following edge cases:
 - Empty app name
 - weird app name (like the one Martin shared)
 - a very long app id (that is valid but has 30 dots for example)

supabase/functions/stats/index.ts
```diff
@@ -13,9 +14,54 @@ const failActions = [
...
+const jsonRequestSchema = z.object({
+  app_id: z.string({
+    required_error: "App ID is required",
+    invalid_type_error: "App ID name must be a string",
+  }),
+  device_id: z.string({
+    required_error: "Device ID is required",
+    invalid_type_error: "Device ID must be a string",
+  }).max(36),
+  platform: z.string({
```
WcaleNieWolny: A platform could a union perhaps?

supabase/functions/stats/index.ts
```diff
@@ -13,9 +14,54 @@ const failActions = [
...
+    invalid_type_error: "Version OS must be a string",
+  }),
+  version_code: z.optional(z.string()),
+  version_build: z.optional(z.string()),
+  action: z.optional(z.string()),
+  custom_id: z.optional(z.string()),
+  channel: z.optional(z.string()),
+  plugin_version: z.optional(z.string()),
+  is_emulator: z.optional(z.boolean()),
+  is_prod: z.optional(z.boolean()),
```
WcaleNieWolny: Can we use the default values? Like is_emulator the default would be false and is prod would be by default true

supabase/functions/updates/index.ts
```diff
@@ -3,20 +3,34 @@ import { serve } from 'https://deno.land/std@0.200.0/http/server.ts'
...
+    required_error: "Device ID is required",
+    invalid_type_error: "Device ID must be a string",
+  }).max(36),
+  version_name: z.string({
+    required_error: "Version name is required",
+    invalid_type_error: "Version name must be a string",
+  }),
   is_emulator: z.optional(z.boolean()),
-  is_prod: z.optional(z.boolean()),
+  is_prod: z.optional(z.boolean())
```
WcaleNieWolny: I made a mistake while working on this zod schema, if you want you can port the fixed version from [here](https://github.com/Cap-go/capgo/pull/387) or wait until that PR is merged and resolve merge conflicts

supabase/functions/updates/index.ts
```diff
@@ -26,6 +40,13 @@ const headersSchema = z.object({
 const bypassRedis = false
 
 async function main(url: URL, headers: BaseHeaders, method: string, body: AppInfos) {
+  let parseResult = null
+  try {
+      parseResult = jsonRequestSchema.parse(body)
```
WcaleNieWolny: There is something called a “safeParse” you can use that instead of a try catch block

supabase/functions/stats/index.ts
```diff
@@ -13,9 +14,54 @@ const failActions = [
...
+}).refine((data) => deviceIdRegex.test(data.device_id), {
+  message: "Device ID must be a valid UUID string"
+});
+
 async function main(url: URL, headers: BaseHeaders, method: string, body: AppStats) {
   try {
     console.log('body', body)
+    let parseResult = null
+    try {
+        parseResult = jsonRequestSchema.parse(body)
```
WcaleNieWolny: Perhaps use zod safe parse, there should exist a method that does not throw but instead returns the error

WcaleNieWolny (changes requested):
Please resolve the comments I left above, but over all the changes are a big improvement over what we had

supabase/functions/stats/index.ts
ayewo: When I checked now, nope since it's used as a container for mixing types (e.g string and number).

ayewo:
@WcaleNieWolny thanks for your comments.

I've pulled in the latest from `main` by discarding my commits, but that ended up closing this PR so I opened a fresh PR in #399.

supabase/functions/stats/index.ts
ayewo: In my testing, `safeParse` is lenient. It will let incorrectly formatted fields through.

it doesn't flag invalid JSON fields unless you wrap `parse` inside a `try / catch`.

supabase/functions/updates/index.ts
ayewo: See my explanation regarding `safeParse` below.

Issue #391, opened 2023-09-21
Allow to have 1 default channel by platform


riderx:
Currently, you can only have one default channel, but over and over users have asked to have one default by platform (Android and IOS)

So it could be nice to change the rules and allow 2 default channels if each of them is bound to one platform.

This should be doable from CLI or Web app and protect the user to do 2 default channels if any of them has the same platform enabled.

It's also important to check if that doesn't break redis cache implementation

PR #394, opened 2023-09-21
Allow 2 default channels

kremedev:
/claim #391

https://github.com/Cap-go/capgo/assets/134702704/d6441cf0-30a2-4df0-bf0b-5b2b4b3f33a2

### Summary by CodeRabbit

- New Feature: Introduced the ability to set default channels for Android and iOS. Users can now select a channel as the default for each platform.
- Improvement: Enhanced the user interface with new translations for setting default channels, providing a more intuitive experience for users across different languages.
- Test: Updated test cases to reflect changes in channel update behavior.
- Chore: Added seed data for testing the functionality of handling multiple default channels.


kremedev:
https://github.com/Cap-go/capgo/assets/134702704/d2b4b0e8-1ba6-49cc-b93f-56fd6243a8f8

kremedev:
https://github.com/Cap-go/capgo/assets/134702704/f0559fcc-713d-43ef-bf1c-823830f70289

WcaleNieWolny (changes requested):
There are fillies missing from this PR. The changes showed in the video is not implemented in this PR

supabase/functions/on_channel_update/index.ts
```diff
@@ -27,7 +27,31 @@ serve(async (event: Request) => {
...
+    }
+
+    if (record.public && record.android) {
+      // find all other channels with same app_i with android true and update them to false
+      const { error } = await supabaseAdmin()
+        .from('channels')
+        .update({ public: false })
+        .eq('app_id', record.app_id)
+        .eq('android', true)
+        .neq('id', record.id)
```
wesleymatosdev: since this query is used three times in this code, you could pre-create it, save the query in a variable, and then resolve it later

something like:
```
      const updateChannels = supabaseAdmin()
        .from('channels')
        .update({ public: false })
        .eq('app_id', record.app_id)
        .neq('id', record.id)
```

and then based on the conditional add one more filter 
```
      const { error } = await updateChannels
        .eq('android', true)
```

ologbonowiwi (changes requested):
left one comment regarding code duplicity on the db queries

supabase/functions/on_channel_update/index.ts
```diff
@@ -27,8 +27,41 @@ serve(async (event: Request) => {
...
+        .from('channels')
+        .update({ public: false })
+        .eq('app_id', record.app_id)
+        .eq('android', false)
+        .eq('ios', false)
+      if (androidError || hiddenError)
+        console.log('error', androidError || hiddenError)
+    }
+
+    if (record.public && (record.ios === record.android)) {
```
wesleymatosdev, line 64: what `record.ios === record.android` means?

this can lead to maintainability and readability issues in the future; it needs to be clarified if both should be false or true (and we should handle both cases equally?)

supabase/functions/on_channel_update/index.ts
```diff
@@ -27,8 +27,41 @@ serve(async (event: Request) => {
...
+        .update({ public: false })
+        .eq('app_id', record.app_id)
+        .eq('android', true)
+        .neq('id', record.id)
+      const { error: hiddenError } = await supabaseAdmin()
+        .from('channels')
+        .update({ public: false })
+        .eq('app_id', record.app_id)
+        .eq('android', false)
+        .eq('ios', false)
```
WcaleNieWolny, line 59: What is
```js
        .eq('android', false)
        .eq('ios', false)
 ```
responsible for?

Comment (kremedev) [reply supabase/functions/on_channel_update/index.ts:59]: It's for the edge case of having default channel with ios and adroid false and making new ios/android channel default.

**Review by WcaleNieWolny — COMMENTED** [review]
Let's talk frontend:
image
This does not say anything about what channel this will be the default for. Will this be the default for android? IOS? Both? This message should say something about the platform

Now for the backend:
Adding the second channels is not as easy as allowing it in the `on_channel_update`. You need to change the `updates` endpoint (and if you do change updates endpoint I cannot not approve that change without writing proper tests) because right now the updates endpoint fails with your change. [This line](https://github.com/kremedev/capgo/blob/08a883967a49731fbbc34bb347e8d34c344dc29f/supabase/functions/_utils/update.ts#L141) forces only 1 default channel to exist.

Here is how I tested your changes:
image
image

```bash
curl -X "POST" "http://localhost:54321/functions/v1/updates" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "app_id": "com.demo.app",
  "version_build": "1.1.0",
  "version_code": "10203999",
  "device_id": "BDACE2AB-53F9-411F-AF7A-C22D104DA632",
  "platform": "ios",
  "version": "1.1.0",
  "version_os": "16.0.2",
  "plugin_version": "4.3.4",
  "version_name": "builtin"
}'

and here is the output I got
```

kremedev:
Frontend with translated platform messages

https://github.com/Cap-go/capgo/assets/134702704/30ef2e4b-0d16-4b75-aa73-fe81f1f35701

WcaleNieWolny:
Now the most important thing would be the backend + tests for it
Also perhaps a warning? For example if you have only the default channel for 1 platform you could display a warning here
image

Let me know what do you think 😄

kremedev:
> The changes look good to me, but this should just have been 1 commit
> 
> image

I was using a public computer. I'm sorry.



> Now the most important thing would be the backend + tests for it Also perhaps a warning? For example if you have only the default channel for 1 platform you could display a warning here image
> 
> Let me know what do you think 😄

I talked with Martin about moving the default setting to channels page but we are using paginated data there and we need all channel data to develop default channel logic and the warning message you want.

WcaleNieWolny:
I have wrote some tests for this PR + general fixes (Like made prod channel the android channel and the new `two_default` channel the IOS channel)

[Here is the patch](https://gist.github.com/WcaleNieWolny/940aeb60469fc3ef7ca47302e6fb28c6)

riderx:
Thanks a lot for the work ! Can you look on the conflict to fix them ?

kremedev:
Is everything okay @WcaleNieWolny ?

WcaleNieWolny:
The one thing I would recommend is renaming variables in the new tests or adding comments to that. But overall I think this looks good :rocket:

WcaleNieWolny (approved):

supabase/functions/on_channel_update/index.ts
wesleymatosdev: you'll address this one? I still think it can lead to issues

supabase/functions/on_channel_update/index.ts
wesleymatosdev, line 59: We could re-use a base query rather than repeat it multiple times.

It'd be awesome do something as in the example below, to avoid repeat the query so many times

```suggestion
    const publicChannelsQuery = supabaseAdmin()
        .from('channels')
        .update({ public: false })
        .eq('app_id', record.app_id)
    if (record.public && record.ios) {
      const { error: iosError } = await publicChannelsQuery
        .eq('ios', true)
        .neq('id', record.id)
      const { error: hiddenError } = await publicChannelsQuery
        .eq('android', false)
        .eq('ios', false)
      if (iosError || hiddenError)
        console.log('error', iosError || hiddenError)
    }

    if (record.public && record.android) {
      const { error: androidError } = await publicChannelsQuery
        .eq('android', true)
        .neq('id', record.id)
      const { error: hiddenError } = await publicChannelsQuery
        .eq('android', false)
        .eq('ios', false)
```

supabase/functions/on_channel_update/index.ts
kremedev: It ensures there is only one default channel if both platforms are true or false.

supabase/functions/on_channel_update/index.ts
WcaleNieWolny: This will only work if you clone the query, you merged 2 queries into 1 and I think this is going to break the logic (?)

tests_backend/tests/updates_test.ts
```diff
@@ -190,6 +190,104 @@ export async function testUpdateEndpoint(backendBaseUrl: URL, supabase: Supabase
...
+    assert(getSecondChannelError === null, `Supabase get second channel error ${JSON.stringify(getSecondChannelError)} is not null`)
+    assert(secondChannel!.public === false, 'Second channel is public')
+  }
+  finally {
+    // Reset to default values
+    const { error: changeProductiuonChannelError2 } = await supabase.from('channels').update({ ios: false }).eq('id', 22)
+    assert(changeProductiuonChannelError2 === null, `Supabase change production channel (ios) error ${JSON.stringify(changeProductiuonChannelError2)} is not null`)
+
+    const { error: changeSecondChannelError } = await supabase.from('channels').update({ public: true }).eq('id', 24)
+    assert(changeSecondChannelError === null, `Supabase change production channel (ios) error ${JSON.stringify(changeSecondChannelError)} is not null`)
```
wesleymatosdev: this could be a helper

```suggestion
    await resetChannels()
```

ologbonowiwi (changes requested):
I left some comments, most of them regarding code duplicity

supabase/functions/on_channel_update/index.ts
wesleymatosdev: Yeah, I'm not sure about the snippet but it'd be nice to have less duplicity on this code

tests_backend/tests/updates_test.ts
kremedev: @WcaleNieWolny thoughts?

WcaleNieWolny: This could be a helper, feel free to implement this @kremedev

ologbonowiwi (approved):
LGTM.

I have some thoughts regarding possible refactors but I can raise a PR after it to get it merged.

Thanks for the contribution!

WcaleNieWolny:
@riderx Can you take a look at this?

Also:
/test backend

kremedev:
@WcaleNieWolny can you try again please?

kremedev:
pass

WcaleNieWolny:
Keep in mind that CI/CD is way slower then your local machine. Can you please set the timeout back to 7 secs?

WcaleNieWolny:
I fixed base.sql and that got merged

/test backend

ologbonowiwi:
@kremedev @WcaleNieWolny @riderx is there anything missing for this PR to be merged?

riderx:
What is missing is my time to test it.
I'm still under water with my log/billing issue and cannot add anything more on my plate.
I'm working 18h a day since days and start to be very drowned by it, very sorry about the wait guys

Issue #380, opened 2023-09-19
Add E2E for upload from CLI


riderx:
CLI Upload should be test on every PR to be sure a PR doesn't break the upload command

WcaleNieWolny:
/attempt #380

This is the same as https://github.com/Cap-go/CLI/issues/156. My solution will fix both of these issue. A detailed idea will be there

PR #397, opened 2023-09-24
CLI tests

WcaleNieWolny:
This PR fixes https://github.com/Cap-go/capgo/issues/380 and https://github.com/Cap-go/CLI/issues/156

This PR is only part of the changes that are required to resolve those issues. I have created a github bot (that is still not 100% complete but I am working on improving it) [in this new repo](https://github.com/Cap-go/github-bot) to manage PRs and run tests for those PR. Here is the idea:

 * I or Martin type `/test cli/backend/all` in a PR and the bot starts a new CI/CD workflow in `Cap-go/github-bot` repo (right now it starts a new workflow in my private repo but the workflow file is the same. (This is done for security reasons so underused code is never run on the CI/CD)
 * This bot posts a comment with the CI/CD workflow link. Here is how such message might look:
 
run
start
fail
image

 * This bot has the ability to link PRs. If someone submits 2 PRs (one to `capgo` and the other one to `CLI`) I or martin can run `/linkpr CLI_OR_CAPGO_PR_LINK` and after that any further `/test` will use both PRs

Here is what I changed:

 * Deleted CI/CD runs for tests (both backend + frontend). This workflow will no longer be used due to this bot
 * Deleted the login E2E test as it was not really used, and I think it was only a test if playwright works
 * Created a `bundle.spec.ts` that has the code from the `login` test + checks if the bundle upload works
 * Added the `testPlaywright` function to the test framework to run playwright tests directly from the test framework
 * Added tests for CLI upload
 * Added `CONTRIBUTING.md` explaining how testing works in capgo
 * Made the `run_backend_tests.ts` way more modular by dynamically loading tests instead of hard importing them
 * Added a sort of steps system to the testing framework so that 1 step can occur after another making the test lest prone to weird bugs
 
A review to both this PR and the new [github-bot](https://github.com/Cap-go/github-bot) repo would be nice ;-)

/claim #380

### Summary by CodeRabbit

---
- New Feature: Added a new test case for bundle functionality in `bundle.spec.ts`.
- Refactor: Split the `test:backend` script into `test:backend`, `test:cli`, and `test:all` scripts in `package.json`.
- Test: Introduced changes to improve backend testing codebase, including modifications to `updates_test.ts`, `test_cli.ts`, and `utils.ts`.
- Documentation: Added a `CONTRIBUTING.md` file with instructions for running tests locally and information about the GitHub Capgo bot.
- Chore: Removed an E2E login test from `login.spec.ts` and replaced it with a more comprehensive test in `bundle.spec.ts`.
---


WcaleNieWolny:
@neo773 @ologbonowiwi A review for this PR would be nice 😄 

Also @neo773 I would love to know you opinion on integrating playwright into this testing framework I wrote

CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
...
+
+All types of contributions are encouraged and valued. See the [Table of Contents](#table-of-contents) for different ways to help and details about how this project handles them. Please make sure to read the relevant section before making your contribution. It will make it a lot easier for us maintainers and smooth out the experience for all involved. The community looks forward to your contributions. 🎉
+
+### Running tests localy
+
+This project uses a custom test runner located in [tests_backend](https://github.com/Cap-go/capgo/tree/main/tests_backend).
+There exists some requirements to run the tests:
+ * Have minio running (see `tests_backend/gh_actions` for referance) For testing the CLI this is not enough, you have to set up the `MINIO_URL` variable. In linux you can use 
+ ```sh
+ export MINIO_URL=$(docker inspect minio1 | grep 'Gateway' | head -n 1 | sed -e 's/            "Gateway": "//g' | sed -e 's/",//g')
```
wesleymatosdev, line 14: can we put part of this command as an npm script? the ideal would be `pnpm test:setup` and would export already, but I know that sometimes this can lead to conflicts, so we can have the `docker inspect minio1 | grep 'Gateway' | head -n 1 | sed -e 's/            "Gateway": "//g' | sed -e 's/",//g'` at least; so this line it would turn into something like 

```suggestion
 export MINIO_URL=$(pnpm get-minio-url)
```
or something like that

Today, that's one of the only things new contributors need to do to set up the project, but in the future, it can have more stuff, so this kind of stuff will be helpful in the long run.

package.json
```diff
@@ -23,7 +23,9 @@
     "reset": "supabase db reset",
     "dev": "npx netlify dev --port 8881",
     "test:e2e": "cypress open",
-    "test:backend": "deno run --allow-all tests_backend/run_backend_tests.ts",
+    "test:backend": "deno run --allow-all tests_backend/run_backend_tests.ts backend",
+    "test:cli": "deno run --allow-all tests_backend/run_backend_tests.ts cli",
+    "test:all": "deno run --allow-all tests_backend/run_backend_tests.ts all",
```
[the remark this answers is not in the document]
ayewo: @CodeRabbit use of `--allow-all` in this context should be a non-issue since it was added for running tests unhindered.

tests/bundle.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  await page.getByRole('button', { name: 'Log in' }).click()
+
+  // Expect the URL to change to the logged in dashboard
+  await expect(page).toHaveURL('http://localhost:5173/app/home')
+
+  // Go to the production channel
+  page.goto('http://localhost:5173/app/p/com--demo--app/channel/22')
+  // eslint-disable-next-line n/prefer-global/process
+  await expect(page.locator('.cursor-pointer > div:nth-child(1) > span:nth-child(1)')).toHaveText(process.env.BUNDLE!)
+})
```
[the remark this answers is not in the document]
ayewo: I agree with the suggestions offered here. It would be nice to pull the credentials and `BASE_URL` into their own `const`.

tests_backend/run_backend_tests.ts
```diff
@@ -72,13 +59,35 @@ async function genTempUpstashEnvFile(upstashToken: string, upstashUrl: string):
...
+
+  const minioUrl = Deno.env.get('MINIO_URL')
+  if (minioUrl) {
+    p.log.info('Minio URL is not null, creating a new env file...')
+    const tempEnvFile = await Deno.makeTempFile()
+    const readEnvFile = await Deno.readTextFile(envFilePath)
+    const finalFile = readEnvFile.replace('host.docker.internal', minioUrl)
+    await Deno.writeTextFile(tempEnvFile, finalFile)
+    return tempEnvFile
+  }
```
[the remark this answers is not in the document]
ayewo: Please also consider defining the string `supabase/.env.exemple` used on lines 60 and 72 as a `const` so that it is easy to keep track of. 

Though it is a valid spelling in French, I think it has been suggested before to fix the tiny typo in the file name by changing ".env.ex**e**mple" to ".env.ex**a**mple".

CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
...
+ * Having `node`, `npx`, `pnpm` installed (Only for CLI tests)
+ * Having the [supabase cli](https://supabase.com/docs/guides/cli) installed
+ * Having a running supabase (`supabase start`)
+
+The tests can be run with the following commands:
+ * `pnpm test:backend` (backend only)
+ * `pnpm test:cli` (cli only)
+ * `pnpm test:backend`
+
+**Running tests localy WILL make changes supabase**
```
ayewo: Minor typo:
> Running tests localy WILL make changes [to] Supabase

CONTRIBUTING.md
ayewo, line 14: You might also want to demonstrate use of the `CLI_PATH` env var in an example.

tests_backend/tests/cli/test_cli.ts
```diff
@@ -0,0 +1,203 @@
...
+        localSupa: `http://${baseUrl.host}`,
+        localSupaAnon: getSupabaseSecret(),
+        statsUrl: new URL('stats', baseUrl).toString(),
+        channelUrl: new URL('channel_self', baseUrl).toString(),
+        updateUrl: new URL('updates', baseUrl).toString(),
+      },
+    },
+  }
+}
+
```
ayewo, line 52: Help me understand your design here.

Why not re-use code from the [demo-app](https://github.com/Cap-go/demo-app) by creating a separate branch that contains the code for what you are trying to do here? 

Since you are already spawning subprocesses, I'm guessing it should be straight forward to `git clone` your branch of the repo then invoke the demo-app from your tests, right?

Of course, you don't have to redo your code but I just want to better understand why you took this approach.

ayewo (commented):
Nice work 👍! I've added a few review comments.

tests_backend/tests/cli/test_cli.ts
WcaleNieWolny: > Why not re-use code from the [demo-app](https://github.com/Cap-go/demo-app)

I originally wanted to clone the `demo-app`, even asked martin about it but then I thought that that it would be wasteful as cloning it and then running `pnpm install` would install a lot of dependencies we do not need. Also this bundle is never used in any device so building this demo app would add extra unnecessary cost (in developers time and in the CI/CD execution time)

I have wrote code that generates a minimal project that can be uploaded without requiring any unnecessary  dependencies. This `generateDefaultJsonCliConfig` has to be here to not hard code any values and although I could have just copied `index.js` and `package.json` I thought it was easier to generate it in code instead of copying a file.

If there ever is a test that will start android emulator with the newly uploaded bundle then I think we should clone `demo-app` but right now it is not needed

tests_backend/utils.ts
```diff
@@ -17,4 +41,76 @@
...
+    finalString += string
+  }
+
+  const stausCode = await subprocess.status
+  if (stausCode.code !== 0) {
+    p.log.error('Playwright output:')
+    console.log(finalString)
+    throw new Error('Playwright test failed')
+  }
+}
```
WcaleNieWolny: I do not like this error, it removes the merged stream and only prints out stderr or stdout

tests_backend/run_backend_tests.ts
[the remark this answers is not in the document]
WcaleNieWolny: If I understand correctly you want to make the `supabase/.env.exemple` read occur only once?

tests/bundle.spec.ts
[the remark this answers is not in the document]
WcaleNieWolny: Can we hard code those values in this `const`? Right now we do not need to dynamically pass it

CONTRIBUTING.md
WcaleNieWolny: Added in latest commit

CONTRIBUTING.md
WcaleNieWolny: Added in latest commit

CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
...
+ export MINIO_URL=$(docker inspect minio1 | grep 'Gateway' | head -n 1 | sed -e 's/            "Gateway": "//g' | sed -e 's/",//g')
+ ```
+ * Have redis running (see `tests_backend/gh_actions` for referance, only if `USE_LOCAL_REDIS` env variable is set to true, tests run fine without it)
+ * Having the `UPSTASH_TOKEN` and `UPSTASH_URL` env variables set (Only if you want to test upstash, tests run fine without it)
+ * Having `node`, `npx`, `pnpm` installed (Only for CLI tests)
+ * Having the [supabase cli](https://supabase.com/docs/guides/cli) installed
+ * Having a running supabase (`supabase start`)
+
+The tests can be run with the following commands:
+ * `pnpm test:backend` (backend only)
```
wesleymatosdev: we could add a `pnpm test:setup` to do stuff like start supabase/up redis (if we need to up it)

tests_backend/run_backend_tests.ts
```diff
@@ -70,15 +57,37 @@ async function getAdminSupabaseTokens(): Promise<{ url: string; serviceKey: stri
...
+async function getEnvFile(redis: 'none' | 'local') {
+  const envFilePath = `supabase/.env.example${redis === 'local' ? '.redis' : ''}`
+
+  const minioUrl = Deno.env.get('MINIO_URL')
+  if (minioUrl) {
+    p.log.info('Minio URL is not null, creating a new env file...')
+    const tempEnvFile = await Deno.makeTempFile()
+    const readEnvFile = await Deno.readTextFile(envFilePath)
+    const finalFile = readEnvFile.replace('host.docker.internal', minioUrl)
+    await Deno.writeTextFile(tempEnvFile, finalFile)
```
wesleymatosdev:
```suggestion
    const fileContent = readEnvFile.replace('host.docker.internal', minioUrl)
    await Deno.writeTextFile(tempEnvFile, fileContent)
```

tests_backend/run_backend_tests.ts
```diff
@@ -99,12 +108,19 @@ async function startSupabaseBackend(
...
+    Deno.exit(1)
+  }
+
   const command = new Deno.Command('supabase', {
     args: [
       'functions',
       'serve',
       '--env-file',
-      (redis !== 'upstash' || !envFile) ? `supabase/.env.exemple${redis === 'local' ? '.redis' : ''}` : envFile,
+      (redis !== 'upstash' || !envFile) ? (redis === 'none' ? noRedisEnvFilePath : localRedisEnvFilePath) : envFile,
```
wesleymatosdev, line 123: the parentheses that you put in there already made pretty clear what this data is, but instead of two ternary expressions we could have the first one already solved

```suggestion
      (redis !== 'upstash' || !envFile) ? envPathCredtedPreviously : envFile,
```

tests_backend/run_backend_tests.ts
```diff
@@ -189,69 +205,108 @@ async function testLoop(functionsUrl: URL, supabase: SupabaseType): Promise<bool
...
-    await startSupabaseBackend('upstash', tempUpstashEnvFilePath)
-    ok = await testLoop(functionsUrl, supabase)
+  }
+
+  for await (const dirEntry of Deno.readDir(folder)) {
+    const firstArg = Deno.args[0]
+
+    // this last conditions checks if we are supposed to run all tests or just the specific one
+    // Like /test cli will run only tests from the cli folder
+    if (dirEntry.isDirectory || !dirEntry.name.endsWith('.ts') || (firstArg !== 'all' ? shortName !== firstArg : false))
```
wesleymatosdev: could we change it to 
```suggestion
    if (dirEntry.isDirectory || !dirEntry.name.endsWith('.ts') || (firstArg !== 'all' && shortName !== firstArg))
```

the result is the same but it seems more readable

tests_backend/run_backend_tests.ts
```diff
@@ -189,69 +205,108 @@ async function testLoop(functionsUrl: URL, supabase: SupabaseType): Promise<bool
...
+      continue
+
+    // Make sure the first time we allways have "normal" backend
```
wesleymatosdev:
```suggestion
    // Make sure the first time we always have "normal" backend
```

tests_backend/run_backend_tests.ts
```diff
@@ -189,69 +205,108 @@ async function testLoop(functionsUrl: URL, supabase: SupabaseType): Promise<bool
...
+    if (backendType !== 'none') {
+      // Only if backend !== null
+      if (backendType) {
+        p.log.info('Killing supabase...')
+        killSupabase()
+      }
+
+      p.log.info('Starting normal backend...')
+      // None means no redis
+      await startSupabaseBackend('none')
```
wesleymatosdev: why not use it explicitly? like 
```suggestion
      await startSupabaseBackend('redis')
```

if `redis` is the actual instead of the "fake" one, it could be `test_redis`, `local_redis` or something like that

tests_backend/tests/backend/updates_test.ts
```diff
@@ -29,17 +34,18 @@ const baseDataIos = {
...
-}
-
 function getBaseDataIos(): typeof baseData {
   return structuredClone(baseDataIos)
 }
 
 const noNew = { message: 'No new version available' }
 
-export async function testUpdateEndpoint(backendBaseUrl: URL, supabase: SupabaseType) {
+async function perapreUpdateTest(_backendBaseUrl: URL, supabase: SupabaseType) {
```
wesleymatosdev:
```suggestion
async function prepapreUpdateTest(_backendBaseUrl: URL, supabase: SupabaseType) {
```

tests_backend/tests/cli/test_cli.ts
```diff
@@ -0,0 +1,203 @@
...
+export function getTest(): RunnableTest {
+  return {
+    fullName: 'Test cli',
+    testWithRedis: false,
+    tests: [
+      {
+        name: 'Prepare cli',
+        test: prepareCli,
+        execute: 1,
```
wesleymatosdev: maybe we could use `skip: false` instead, and the default would be false; what do you think?

set skip to true it would be more clear since execute a test is usually the standard behavior

ologbonowiwi (changes requested):
I left some comments, but it is a great job, though

CONTRIBUTING.md
WcaleNieWolny: perhaps, however please be aware that this is a linux only example. I have no idea how to run this in powershell and making this into a script could perhaps make this command compatible with both linux and windows. But I am worried will happen if this script fails

tests_backend/run_backend_tests.ts
WcaleNieWolny: I am not 100% sure you understood this change. We test the following scenarios:

 * Redis not configured
 * Redis configured (hosted localy)
 * Redis configured (hosted on upstash)
 
You just replaced the code that starts supabase without redis with code that starts supabase with redis local

tests_backend/tests/cli/test_cli.ts
WcaleNieWolny: This is more about how many times should the test execute (like the big update endpoint test [here](https://github.com/WcaleNieWolny/capgo/blob/cc80c5866154380533b90395fe547fa96f951041/tests_backend/tests/backend/updates_test.ts#L17))

CONTRIBUTING.md
WcaleNieWolny: I do not think this is a great idea, we are already doing this from the [CI/CD](https://github.com/Cap-go/github-bot/blob/8d796aa830627b7d70e169b02de71f1bda86aec4/.github/workflows/test_cli.yml#L109) and this script in `gh_actions` sets minio in a very temporary way (Making the data folder in the `/tmp` making the entire data disappear on a system reset).

Also again: This script in `gh_actions` is only for linux as a reference and it does not work on windows.

Also by default only minio and supabase is required and the tests will run fine without redis
image

tests_backend/tests/backend/updates_test.ts
```diff
@@ -1,18 +1,23 @@
...
+        test: perapreUpdateTest,
+        execute: 1,
+      },
+      {
+        name: 'Test updates endpoint (big)',
+        test: testUpdateEndpoint,
+        execute: 3,
+      },
+    ],
+  }
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in latest commit

tests_backend/tests/backend/updates_test.ts
WcaleNieWolny: I will not commit this thru github (as you have not changed this in the test declaration) but I fixed this in the latest commit

tests_backend/run_backend_tests.ts
WcaleNieWolny: I do not understand what are out asking for to be honest

WcaleNieWolny: I do not see how this `envPathCredtedPreviously` could improve the codebase, could you please explain it?

neo773:
Removing the playwright action workflows is a bad idea. It lacks integration and misses out on logs, binary cache, and artifact reporting. 
Also, duplicating code in `run_backend_test.ts ` is unnecessary when `playwright.config.ts` already handles starting backend and frontend. Could be much more cleaner with playwright api testing https://playwright.dev/docs/api-testing

WcaleNieWolny:
I do want to integrate playwright into the E2E tests. Also I do have to start the backend in this E2E framework. How do you see integrating playwright into this E2E testing system?

WcaleNieWolny:
You always test a specific part (like bundle upload from the cli) and then you can run a playwright test to make sure everything worked

neo773:
So here’s how we could tackle this
- Extracting the frontend & backend runner logic to a separate file
- Run it before running before running the actuals tests
- This will unify the frontend & backend E2E testing with Playwright
- CLI test will probably have to stay as it is 

This will ensure no code duplication while retaining the existing advantages of running tests in workflow actions

CONTRIBUTING.md
wesleymatosdev: but do we need to take care of windows also?

all of the OSS I've contributed only works on Unix systems; this is pretty standard, and the provided code already works only on Linux. make it a script that'd make it easy to run (instead of copying/pasting, it'd be just run the script).

although I think this would make easier to execute these stuff to newcomers, it's optional, so feel free to solve this issue as you want

CONTRIBUTING.md
wesleymatosdev: IDK if we currently support Windows as a development OS, but I don't think it's required to support Windows to have stuff on node scripts if it's not something that we'll usually run locally to test against new changes; feel free to solve this issue as it is

tests_backend/run_backend_tests.ts
wesleymatosdev: yeah sure

the thing is that having two ternary expressions can lead to future issues in maintainability and readability

it would make it more accessible to solve the expression `redis === 'none' ? noRedisEnvFilePath : localRedisEnvFilePath` first in a variable and then use its value, rather than `expression1 ? expression2 ? value : otherValue : falseValue`

it's easy to read now cause it's fresh, but it definitely can bring some confusion to less experienced developers

tests_backend/run_backend_tests.ts
wesleymatosdev: it is confusing that we set `none`, but `none` means the local redis that the app starts. I got the idea but the code needs to explain itself.

tests_backend/tests/cli/test_cli.ts
wesleymatosdev: ohhhh, I thought it was a boolean value; in this case, we could use `executions` maybe? `executionsNumber`? IDK, but something that explains that this is the number of executions, `execute` seems like a true/false flag :eyes:

WcaleNieWolny:
So... this entire testing framework was a waste of time?

WcaleNieWolny:
> Extracting the frontend & backend runner logic to a separate file

Not possible, this test (CLI test), uses both the backend (to upload the bundle) and only then it spawn playwright to make sure the frontend updated. Let me know how do you see spiting this logic

> This will unify the frontend & backend E2E testing with Playwright

We COULD just remove the backend spawninbg from playwright, if we only ever spawn playwright thru this custom E2E framework then we can be 100% that the backend is up when running playwright

As for CI/CD i am not 100% you have looked at the whole picture of this. As I said there exists a WIP [github bot](https://github.com/Cap-go/github-bot) to start the tests in CI/CD

The advantages of starting tests from this bots are:
 * No code duplication (Workflow file does not have to be changed in both `capgo` repo and in `CLI` repo
 * Hardened security (No insecure or potentially malicious is ever started in the CI/CD) This is achieved as only users with merge permissions would be able to start CI/CD
 * Less CI/CD spam (Do we REALLY need an E2E test for a typo fix?)
 * Linking CLI and capgo PR when testing this
   

It does have some drawback:
 * Making the entire CI/CD workflow harder to follow
 * Requiring approval for starting CI/CD
 
 [Here](https://github.com/Cap-go/github-bot/blob/main/.github/workflows/test_cli.yml) is the new workflow responsible for tests and although as of writing this it does not have playwright integrated, it could reuse the caching playwright bins from your workflow that I have removed in this PR.
 
 The main reason for removing your workflow and my E2E test workflow is security. Right now if a bad actor submits a malicious PR with the whole intention of stealing secrets (like the supabase production secret) his code will run on the CI/CD. I am not 100% it is possible to escalate this to stealing this secrets but running untrusted code even with the slightest possibility that this attack is possible just sound like a terrible idea

neo773:
> Not possible, this test (CLI test), uses both the backend (to upload the bundle) and only then it spawn playwright to make sure the frontend updated. Let me know how do you see spiting this logic

Can we not do something like this? Don't see why it would not work.

`deno run start.ts`
`start test:cli`
`start test:e2e`

> We COULD just remove the backend spawninbg from playwright, if we only ever spawn playwright thru this custom E2E framework then we can be 100% that the backend is up when running playwright

Yeah, that's what I meant, this would be taken care of by `deno run start.ts`

As for the security, It shouldn't be a concern since we're not using any supabase production secrets for any test workflows (common practice is to have staging secrets if it did involve using it)




> It does have some drawback:
> 
> * Making the entire CI/CD workflow harder to follow
> * Requiring approval for starting CI/CD

Exactly, it's hard enough as it is to keep track of the changes, this would only fragment it further.
I think moving `CLI` repo into `capgo` repo would make it easier, since we already use pnpm making it a monorepo would be as simple as creating a `pnpm-workspace.yml` file.

cc @riderx

tests_backend/tests/cli/test_cli.ts
riderx: @WcaleNieWolny why you are not using capgo itself as test upload ?
Capgo is a mobile app as well look the file capacitor.config.ts, it uses Capgo cloud to self update.

WcaleNieWolny:
@riderx this is ready to be merged

WcaleNieWolny:
Again testing cli. This is due to the latest commits. Sorry for the spam

/test cli

riderx:
Thanks all for the work guys !

PR #335, opened 2023-09-06
Organization system

WcaleNieWolny:
This PR implements V0 for the organization system. This is ALMOST ready, however due to the size of this PR rigorous testing have to  be performed to make sure after merging this PR everything will go smoothly.

I am opening this to get feedback from @riderx but there are a couple things missing from this PR as is:

 * <s>Videos showing changes</s>
 * <s>Rigorous testing, I have tested this code, however there are likely some bugs that will be cough with more tests</s>
 * <s>SQL queries to migrate production (So that this can be merged)</s>
 * <s>Clear of `types.ts`, there are a lot of unnecessary changes there</s>

/claim #258

### Summary by CodeRabbit

- New Feature: Introduced a new Vue component `DropdownOrganization.vue` for organization selection and invitation management.
- New Feature: Added a `redirect` property to the `Tab` interface for optional redirection.
- New Feature: Implemented role-based permissions and access control within the organization system.
- New Feature: Added a new Vue component for managing organization information, including changing the organization logo and name.
- New Feature: Added support for toggling between dark and light modes in the UI.
- Refactor: Simplified the template in `InfoRow.vue` by removing unnecessary code.
- Refactor: Updated conditions for deleting records in `SharedUserTable.vue`.
- Refactor: Removed photo handling code from `Account.vue` and moved it to a new service module `photos.ts`.
- Bug Fix: Modified the `useSupabase` function to infer the type of `options` variable.
- Bug Fix: Updated the `main` function in `upload_link/index.ts` to retrieve user ID based on API key and app ID.

PR #405, opened 2023-10-04
Selectable disallow (metadata disallow)

### LINKED ISSUE #360: Allow select for major or minor for security
**Describe the bug/issue**
Currently we check by default if a major bump in version happened to disable update.
This behavior is used for native code change to prevent install update on device who doesn’t have the native code.
Over time I saw many complaints about doing a major bump for something final users don’t care.
Even if the display version can be different in App Store it seems annoying.

So one change who could be nice is to allow capgo user to set how disable work.
- major 
- minor
- version number 
- None

So The current behavior is “major”, minor will be check if the minor coming from device ( 2 in 1.2.3) is equal or above the latest in the channel.

version number will use the min version send as metadata in the upload step( cli should be modified as well), this version has to be displayed in the UI, if this data is missing and this setting is active, it will use it’s own number as min number.

an none will be no check, same as we disable the feature currently.

this settings should be set from the UI or the CLI.

By default the behavior should be major, and migration script should be provided to migrate all users.

WcaleNieWolny:
This PR fixes #360. /claim #360

First of all here is a brief list of changes:
 * Playwright + supabase integration
 * Improve playwright auth
 * Show an error if the channel is misconfigured (metadata is lacking)
image
 * Show the minimal update version if the metadata disallow is on in both channel and bundle
 * Add a switch to select update strategy in channel settings
 * Add a CLI change to make sure that if updates are disabled based on this metadata then require this metadata to be present when uploading (different PR)
 * Add minor disallow (if updating from `1.2.0` -> `1.3.0` this option will prevent such update)
 * Change backend to check metadata and disallow update based on that
 * Update playwright
 * Add tests for all of the features I mentioned (fronend, backend + CLI)
 
At this time I will not be providing a video of the changes as it would be near impossible to show all edge cases in that video. I believe the extensive tests for this feature are enough.


WcaleNieWolny:
The tests failed, I will rerun them as I changed some things

/test all

WcaleNieWolny:
The new tests are failing because of the state of the db when the tests are being run. I have only tested this partially (I haven never ran `all` test before submitting this PR) I am working on improving this.

@ologbonowiwi could you review this PR if you have a sec

WcaleNieWolny:
I have debugged and improved the tests. I think I remove the weird state issues within the new tests.

/test all

locales/en.yml
wesleymatosdev, line 1: nice changes, re-organizing the order of the keys.

one thing, though, `yes` and `no` are the only ones with a single comma (`'`) on the key; is that intentional?

src/components/tables/ChannelTable.vue
```diff
@@ -142,6 +159,28 @@ async function getData() {
...
+      if (channel.version.minUpdateVersion === null) {
+        channel.misconfigured = true
+        anyMisconfigured = true
+      }
+
+      if ((channel.enable_progressive_deploy || channel.enableAbTesting) && channel.secondVersion.minUpdateVersion === null) {
+        channel.misconfigured = true
+        anyMisconfigured = true
+      }
+    }
```
wesleymatosdev, line 180: you could use the `.map` instead of looping through the array twice

something like:
```suggestion
    .map(e => {
      const channel = e as any as typeof element

      if (channel.version.minUpdateVersion === null) {
        channel.misconfigured = true
        anyMisconfigured = true
      }
      if ((channel.enable_progressive_deploy || channel.enableAbTesting) && channel.secondVersion.minUpdateVersion === null) {
        channel.misconfigured = true
        anyMisconfigured = true
      }
      
      return channel
    })
```

wesleymatosdev: when we loop it only once, one thing that we could improve is the conditionals.

we can use names to explain the conditions and set misconfigured only once

an example: 
```typescript
      const minUpdateVersionIsNull = channel.version.minUpdateVersion === null
      const isProgressiveDeployOrAbTesting = channel.enable_progressive_deploy || channel.enableAbTesting
      const secondMinUpdateVersionIsNull = channel.secondVersion.minUpdateVersion === null
      const nameToBeChanged = isProgressiveDeployOrAbTesting && secondMinUpdateVersionIsNull

      if (minUpdateVersionIsNull || nameToBeChanged) {
          channel.misconfigured = true
          anyMisconfigured = true
      }

     return channel
```

src/pages/app/p/[p]/bundle/[bundle].vue
Comment (wesleymatosdev) [inline src/pages/app/p/[p]/bundle/[bundle].vue:1]: there are two console logs (lines 69 and 453); these will be merged, or are they only debug consoles?

tests/auth.setup.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test as setup } from '@playwright/test'
```
wesleymatosdev, line 1: this can bring confusion since it's well spread by the node community to have a `test` key that declares test cases

when I saw `setup` on the code, I went back to the imports to see if it was a method we made and not from the test library. this likely will bring confusion to contributors

```suggestion
import { expect, test } from '@playwright/test'
```

tests/auth.setup.ts
```diff
@@ -0,0 +1,38 @@
...
+  await page.getByRole('button', { name: 'Log in' }).click()
+
+  // Expect the URL to change to the logged in dashboard
+  await expect(page).toHaveURL(`${BASE_URL}/app/home`, { timeout: 60_000 })
+
+  await page.context().storageState({ path: authFile1 })
+})
+
+// This will be usefull some day
+// setup('authenticate as test2', async ({ page }) => {
```
wesleymatosdev, line 24: Since setup is actually `test` from the playwright, you can uncomment this code and use `test.skip`

tests/auth.setup.ts
wesleymatosdev, line 1: why this file is called `.setup` rather than `.test` our `.spec`?

tests/selectable_disallow.spec.ts
wesleymatosdev, line 1: rather than using css locators (`something > other_thing`), you can set data-testid on the component and use the [getByTestId playwright's locator](https://playwright.dev/docs/locators#locate-by-test-id) instead.

It's more reliable and uncoupled since it doesn't depend on how the HTML is structured

tests/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    return supaClient
+  // eslint-disable-next-line n/prefer-global/process
+  supaClient = createClient<Database>(process.env.SUPABASE_URL ?? defaultSupabaseUrl, process.env.SUPABASE_ANON ?? defaultSupabaseAnonKey, options)
+
+  const { error } = await supaClient.auth.signInWithPassword({
+    email: DEFAULT_EMAIL,
+    password: DEFAULT_PASSWORD,
+  })
+
+  expect(error).toBeNull()
```
wesleymatosdev, line 42: Excellent use of expectation here to handle the error!

tests_backend/run_backend_tests.ts
```diff
@@ -24,7 +24,7 @@ let backendType: 'none' | 'local' | 'upstash' | null = null
 const noRedisEnvFilePath = await getEnvFile('none')
 const localRedisEnvFilePath = await getEnvFile('local')
 
-async function getAdminSupabaseTokens(): Promise<{ url: string; serviceKey: string }> {
+async function getAdminSupabaseTokens(): Promise<{ url: string; serviceKey: string; anonToken: string }> {
```
wesleymatosdev, line 27: we could give a better name than `anonToken` here, something like `anonymousToken`, `publicToken`, etc.; something that auto-explains what this thing is

ologbonowiwi (changes requested):
Left some comments but the core stuff looks great, and it's nice that we're adding more tests almost on each PR 🚀

WcaleNieWolny:
I really am trying to reproduce this CI fail, I have change the playwright config. I will test again all and selectable_disallow without the rest. I am not sure why this if failing, I cannot reproduce this fail in my local setup

tests_backend/run_backend_tests.ts
WcaleNieWolny: This is called `supa_anon` in config.js `anon key` in `supabase status`. Anon is the term used officially by supabase

tests/utils.ts
WcaleNieWolny: thanks 😄

locales/en.yml
WcaleNieWolny: I think that the VS code extensions changed something in a weird way. I could fix that, tho it does work as it

src/pages/app/p/[p]/bundle/[bundle].vue
Comment (WcaleNieWolny) [reply src/pages/app/p/[p]/bundle/[bundle].vue:1]: Good catch, removed in the latest commit

tests/auth.setup.ts
WcaleNieWolny: I have followed the [official tutorial](https://playwright.dev/docs/auth) for this. This is not here to test auth, rather to remove the requirement for each invididual test to authenticate. This is here to prevent code duplication and not to test anything

tests/auth.setup.ts
WcaleNieWolny: Again, I have used the [official tutorial(https://playwright.dev/docs/auth) and that is how it has been done there

tests/auth.setup.ts
WcaleNieWolny: I could also remove it. This is here only because I originally  started to work on this file for the [organization pr](https://github.com/Cap-go/capgo/pull/335/) and there having 2 users was required.

This is here so that if anyone wants to add a second user in the tests one day he or she can just uncoment this

tests/selectable_disallow.spec.ts
WcaleNieWolny: That is an interesting thought, I will look into this

WcaleNieWolny:
I am doing the test again. Slowly starting to think that the CI is just slower then my PC causing this to fail

/test selectable_disallow

WcaleNieWolny:
Tests are still failing. I am wondering if this is due to the spec diff between CI/CD and my PC. I have increased the timeout. If it works in the CI/CD then I will think about improving this

/test selectable_disallow

WcaleNieWolny:
> Test timeout of 20000ms exceeded

yeah, this is just slow in the CI/CD

tests_backend/run_backend_tests.ts
wesleymatosdev: Thanks for clarifying!

tests/auth.setup.ts
wesleymatosdev: Oh, I never saw it! I always used `beforeAll` to do this, but knowing another way is good.

No change on this file is needed then, thanks!

WcaleNieWolny:
I think it could be fixed in the latest commit, I have change the state problems

/test selectable_disallow

WcaleNieWolny:
I have changed the dev server in tests. I think I figured out why my tests failed.

/test selectable_disallow

WcaleNieWolny:
FINALY!! After 3 days the tests passed 🥳 

/test all

WcaleNieWolny:
Another day, another attempt. PLEASE just work

/test selectable_disallow

WcaleNieWolny:
PERHAPS THIS WORKS NOW

/test all

WcaleNieWolny:
I have added a fix for a /updates endpoint bug. This bug was not related to my changes

/test all

WcaleNieWolny:
Added a test for the bug fix + fixed the endpoint

/test backend

riderx:
thanks @ologbonowiwi  /tip $20 @ologbonowiwi

PR #418, opened 2023-10-14
Fix perms

WcaleNieWolny:


supabase/migrations/20231014161224_fix_api_key_policy.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+  ALTER POLICY "allow apikey to select" ON "public"."apps" USING (is_allowed_capgkey(((current_setting('request.headers'::text, true))::json ->> 'capgkey'::text), '{all,write,upload,read}'::key_mode[], app_id));
```
wesleymatosdev, line 2: Is it possible to write this in more than one line? It's pretty hard to read everything in one single line

supabase/migrations/20230930062908_drop_constraint.sql
```diff
@@ -1,8 +1,4 @@
 ALTER TABLE devices DROP CONSTRAINT devices_app_id_fkey;
-ALTER TABLE devices DROP CONSTRAINT devices_version_fkey;
-ALTER TABLE stats DROP CONSTRAINT logs_app_id_fkey;
-ALTER TABLE stats DROP CONSTRAINT logs_device_id_fkey;
-ALTER TABLE stats DROP CONSTRAINT logs_version_fkey;
```
wesleymatosdev, line 5: does this change take effect? considering this already ran

supabase/migrations/20230815171919_base.sql
```diff
@@ -1884,9 +1884,6 @@ ALTER TABLE ONLY "public"."deleted_account"
 ALTER TABLE ONLY "public"."devices_override"
     ADD CONSTRAINT "devices_override_pkey" PRIMARY KEY ("device_id");
 
-ALTER TABLE ONLY "public"."channel_devices"
-    ADD CONSTRAINT "channel_devices_pkey" PRIMARY KEY ("device_id");
-
```
wesleymatosdev, line 1889: does this change take effect? considering this already ran

supabase/migrations/20230930062908_drop_constraint.sql
WcaleNieWolny: Martin broke [base.sql](https://github.com/Cap-go/capgo/actions/runs/6512192143/job/17689348153) and the `supabase up` was no longer working (because of the broken base.sql). I fixed the base.sql and then it threw an exception in this migration so I removed the lines that were causing the problems (The problem was that `logs_app_id_fkey` constraint did not exist)

supabase/migrations/20230815171919_base.sql
WcaleNieWolny: See, the thing is that I have not REALLY removed anything. This constraint is duplicated on [this line](https://github.com/Cap-go/capgo/blob/2210c90ae8c9dd957914d658b37e7f6f5cf989a1/supabase/migrations/20230815171919_base.sql#L1872) and [this line](https://github.com/Cap-go/capgo/blob/2210c90ae8c9dd957914d658b37e7f6f5cf989a1/supabase/migrations/20230815171919_base.sql#L1887). Since you cannot have 2 constrains that are exactly the same I had to remove this line

supabase/migrations/20231014161224_fix_api_key_policy.sql
WcaleNieWolny: I do not know, I have copied this code from supabase and have not changed a bit. The truth is that we have even longer lines in [base.sql](https://github.com/Cap-go/capgo/blob/2210c90ae8c9dd957914d658b37e7f6f5cf989a1/supabase/migrations/20230815171919_base.sql#L2247) so I would not risk breaking this

ologbonowiwi (approved):
thanks for the patience answering the comments. 

LGTM

PR #428, opened 2023-10-19
Fix app retention

WcaleNieWolny:


supabase/migrations/20230815171919_base.sql
```diff
@@ -1042,6 +1042,7 @@ BEGIN
     SET deleted = true
     FROM apps, app_versions_meta
     WHERE app_versions_meta.app_id = app_versions.app_id
+    AND app_versions.app_id = apps.app_id
```
wesleymatosdev, line 1045: I know that I've already commented this on other PRs, but this works?

The migration was not ran already?

WcaleNieWolny: It requires manual migration by Martin, no idea if there is a better way.

I tested this change and it fixed the issue

ologbonowiwi (approved):
LGTM

Issue #415, opened 2023-10-13
Add to Cloudflare worker update and stats endpoint


riderx:
Currently, Cloudflare worker is redirecting to Supabase function for update and stats endpoint.
This lead to increase the latency and cost x3.
Cloudflare worker price + Supabase worker price (2 time Cloudflare one)
So the goal is to make a script like we have for Netlify serverless function where we automatically generate the code for Cloudflare workers.
One worker should be used to reduce cost.
Bonus, a rework on the generate Netlify logic could be done to simplify the code:
https://www.npmjs.com/package/hono
could be used on every platform (Supabase function, Cloudflare, Netlify and Netlify edge) to make the code the same

rishi-raj-jain:
Proceeding as talked with @riderx on Discord

riderx:
seems :
https://docs.deno.com/runtime/manual/advanced/deploying_deno/cloudflare_workers could help for cloudflare worker

rishi-raj-jain:
Waiting for Martin to test the latest update in prod

rishi-raj-jain:
Martin verified, onto this one now

WcaleNieWolny:
I guess that since I [create a partial solution](https://github.com/Cap-go/capgo/pull/448) I should announce that I am attempting this

/attempt #415

riderx:
@rishi-raj-jain i made this for the work on the captime repo

riderx:
/tip $30 @rishi-raj-jain thanks for the md file migration !

PR #448, opened 2023-11-04
Denoflare code gen + enviroment variables

WcaleNieWolny:
[Denoflare](https://denoflare.dev) + codegen for it.

Also ENV from denoflare -> `getEnv` fn

/claim #415


WcaleNieWolny:
With the latest commit this can generate the deno flare config. Thus this is ready to be deployed (I leave deployment to @riderx as he knows how to do cloudflare best but the entire edge fn -> cloudflare is done)

Issue #447, opened 2023-11-04
Make sure when bundle version is deleted it's delete from storage.


riderx:
Right now this has many edge cases not handle.
So this should be fixed, tested and having a script to look for all old storage and remove the bundle of version already deleted in the db

WcaleNieWolny:
@mrkirthi-24 Hey, how is your progress?

Can I attempt this?

mrkirthi-24:
@WcaleNieWolny you can go ahead

WcaleNieWolny:
Thanks 😄

/attempt #447

PR #449, opened 2023-11-05
S3 missing supabase version script

WcaleNieWolny:
This script is here to detect and delete ALL versions in R2 that are not present in supabase or are marked in supabase as `deleted`.

It will make a local copy of all versions to delete before performing any destructive action.
The purpose of this script is to de-clutter R2. If at any point the edge fns were down or simply did not trigger properly then unused data is floating in the R2.

Given [this commit](https://github.com/Cap-go/capgo/commit/739e5ffe34b5bee2232050abb3f4447ac9918398) and this script I conciser this to solve #447

I tested my script on X86 linux (gentoo). When running in prod I **STRONGLY RECOMEND** piping the output into a file like this:
```sh
deno run --allow-all supabase/functions/_script/remove_not_used_r2.ts > /tmp/out.txt
```

/claim #447

### DIFF
--- a/supabase/functions/_script/remove_not_used_r2.ts
+++ b/supabase/functions/_script/remove_not_used_r2.ts
@@ -18,6 +18,8 @@ function getEnv(env: string) {
   return Deno.env.get(env) ?? ''
 }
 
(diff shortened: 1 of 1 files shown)


riderx:
I think this script has to take in account also the old way to store bundle.
In supabase storage and r2 without folder.
This require to look at old commit to see how it was done before

WcaleNieWolny:
@riderx I finished the legacy version deleting

Issue #255, opened 2023-07-23
Bundle set to multiple channel are not clear


riderx:
**Describe the bug/issue**
If a specific bundle is associated to multiple channels, the UI for the bundle itself only shows 1 of them which is confusing.

The goal is the show all of them and add an easy way to remove the link directly from the bundle page

WcaleNieWolny:
Please note a couple of things:

A) Tests for this would be nice (we have playwright, take a look at [this](https://github.com/Cap-go/capgo/blob/development/tests/organization_system.spec.ts) and [this](https://github.com/Cap-go/capgo/blob/development/tests_backend/tests/organization/main.ts) for some details with impl)
B) Please base your changes on the [development branch](https://github.com/Cap-go/capgo/tree/development) to avoid merge conflicts

yuvrajjsingh0:
> Please note a couple of things:
> 
> A) Tests for this would be nice (we have playwright, take a look at [this](https://github.com/Cap-go/capgo/blob/development/tests/organization_system.spec.ts) and [this](https://github.com/Cap-go/capgo/blob/development/tests_backend/tests/organization/main.ts) for some details with impl) B) Please base your changes on the [development branch](https://github.com/Cap-go/capgo/tree/development) to avoid merge conflicts

Cool, I will take care of it!

PR #461, opened 2023-12-02
Shows All the channels for bundles and can unlink them easily

yuvrajjsingh0:
/claim #255


WcaleNieWolny:
Looks good, there are no tests but i could add them in the morning

riderx:
Can you share screenshots ?

yuvrajjsingh0:
> Can you share screenshots ?

Screenshot 2023-12-03 032619
On Click on any channel: 
Screenshot 2023-12-03 032802
On unlink:
Screenshot 2023-12-03 032817

WcaleNieWolny:
image

I am not a big fan of the spacing, it just looks out of place. Could you move it to the left so it matches the rest of the table?

Also I found a bug in your code. I enabled AB testing 
image

And set the same bundle for both version A and version B
image

Then I checked the bundle page and only 1 of those channels showed up
image

The desired behavior would be to have `two_default-B, two_default-A` (Order does not matter)

Given this setup I tried to link the same bundle to the `no_access` channel. It worked so I tried to unlink it but unfortunately I was unable to

https://github.com/Cap-go/capgo/assets/50914789/c4037967-d274-4ca7-a857-7d9e4696459a

After disabling AB testing everything worked correctly 

https://github.com/Cap-go/capgo/assets/50914789/73915ddf-4a6a-4da3-be6e-13518b62fd38

yuvrajjsingh0:
I have fixed these bugs, here's the video of same 

https://github.com/Cap-go/capgo/assets/20877577/d84aa36a-4011-40c5-aa49-cdcb4d91e0d8

supabase/functions/_utils/lemon.ts
WcaleNieWolny, line 1: why is this file here?

yuvrajjsingh0: I have removed this file. I was working on some other issue and stashed it

yuvrajjsingh0:
@riderx any update on this?

PR #471, opened 2023-12-13
Migrate capgo from pnpm to bun Fixes #470

### LINKED ISSUE #470: Switch to bun
Switch the whole repo to bun.
CI included

RutamBhagat:
Fixes issue #470 
 /claim #470
Migrate from pnpm to bun


RutamBhagat:
> Migration from pnpm to bun

coud not fighure out what to do with the capacitor.setting.gradle, Podfile and Podfile.lock file and I might need some help with the CI yml files

RutamBhagat:
@riderx  would it be ok if I let the /tests_backend/tests/cli/test_cli.ts file use pnpm instead of bun? Its causing the E2E test to fail

riderx:
I think all the capacitor CLI commands do not support bun for now.
Outside of that things should work, maybe @WcaleNieWolny can help on the test to know if there is a way or not, since he made it.
I'm not sure to understand your ask about:
 capacitor.setting.gradle, Podfile and Podfile.lock

WcaleNieWolny:
> @riderx  would it be ok if I let the /tests_backend/tests/cli/test_cli.ts file use pnpm instead of bun? Its causing the E2E test to fail

PNPM is used in the demo app as far as I know so changing that to bun does not have much sense as as I am concerned

riderx:
@WcaleNieWolny maybe it's late and i'm not able to read anymore, but i don't understand your comment.
just in case i added bun if the bot repo.
Since the direction is to switch to bun mwe could switch demo app as well.

WcaleNieWolny:
> @WcaleNieWolny maybe it's late and i'm not able to read anymore, but i don't understand your comment. just in case i added bun if the bot repo. Since the direction is to switch to bun mwe could switch demo app as well, if that a blocker

As you are aware we have a [test app](https://github.com/Cap-go/demo-app). And in that test app we use `pnpm`
image

The [/tests_backend/tests/cli/test_cli.ts](https://github.com/Cap-go/capgo/blob/main/tests_backend/tests/cli/test_cli.ts) file currently uses `pnpm` and this PR changed that to bun.

As it was said: That change (pnpm -> bun) caused the tests to fail. 

So as for this

> would it be ok if I let the /tests_backend/tests/cli/test_cli.ts file use pnpm instead of bun

Yes, I am 100% fine with this using pnpm because this test should be as simple

Also there [exists a cache step](https://github.com/Cap-go/github-bot/blob/152fdc75b015eebd1c329c4950bfe0ed569ce9c3/.github/workflows/test.yml#L107) that I do not thing was taken into account in this PR

riderx:
@WcaleNieWolny i did check for the cache in bun action repo and the author say basically the cache github is slower than bun so this is useless ^^.
Let's keep pnpm for the demo app, but i believe at one point we need to have all same

RutamBhagat:
> Also there [exists a cache step](https://github.com/Cap-go/github-bot/blob/152fdc75b015eebd1c329c4950bfe0ed569ce9c3/.github/workflows/test.yml#L107) that I do not thing was taken into account in this PR

@WcaleNieWolny I am sorry I am a bit confused, that cache step you mentioned is in github-bot repo and this PR is for the capgo repo.
Also I have reverted the changes made in test_cli.ts file and all E2E tests are now passing except for this protestware warning
image
Let me know if any other changes are required

riderx:
Found some problem in the bun command it was still some dlx command from pnpm mixed with new bun, where it shoud be "bun x"

WcaleNieWolny:
> > Also there [exists a cache step](https://github.com/Cap-go/github-bot/blob/152fdc75b015eebd1c329c4950bfe0ed569ce9c3/.github/workflows/test.yml#L107) that I do not thing was taken into account in this PR
> 
> @WcaleNieWolny I am sorry I am a bit confused, that cache step you mentioned is in github-bot repo and this PR is for the capgo repo. Also I have reverted the changes made in test_cli.ts file and all E2E tests are now passing except for this protestware warning image Let me know if any other changes are required

Hard to say, I never experienced this protestware and the scraper (upstream package of martin's fork) does not have any issue regarding this so it's likely a minor issue. Regardless, thank you for your report

PR #484, opened 2023-12-19
feat: add oauth providers

anikdhabal:
Screenshot 2023-12-19 230458


anikdhabal:
cc @riderx

riderx:
thanks @anikdhabal there issue here, Capgo is also app mobile and adding this require to config it also for mobile app.
so it need mobile code changes with capacitor and test about this for google at least.
Also to be accepted by Apple if we use multiple login option we need Apple one as well.
So i'm not sure i will be able to merge that asap without having to have many support issue

anikdhabal:
> thanks @anikdhabal there issue here, Capgo is also app mobile and adding this require to config it also for mobile app. so it need mobile code changes with capacitor and test about this for google at least. Also to be accepted by Apple if we use multiple login option we need Apple one as well. So i'm not sure i will be able to merge that asap without having to have many support issue

Okay

Issue #526, opened 2024-01-17
As an org member who is in paid plan I still see trial banner.


riderx:
When i switch to org who has paid plan i shouldn't see the trial banner of my other org.
![image](https://github.com/Cap-go/capgo/assets/4084527/21817a33-3a09-45f3-8dd9-786aeb0e3232)

Min2who:
/attempt #526
can i get assigned?

WcaleNieWolny:
Before attempting this, please provide an idea on how you are going to solve this issue. For some context, here is the situation:

Here is the situation:

 - User A invited user B to org
 - User B has purchased a plan
 - User A i still on trial, but owns all the apps
 - User's A trial runs out, now he cannot use capgo as capgo does not use the plan from user B
 
 Here are some problems to consider in your solution:
 - An org might consist of multiple paid users. (Let's say user A invites user C. User C has a paid plan. Now, both user C and B have paid plans. What happens now?)
  - How do you warn user B that joining will cause user A to use his plan?
  - How do we redesign the payment system without making it a total mess to solve this issue?

WcaleNieWolny:
Here is the solution I propose, please note that this is just an idea from someone who implemented the org system. @riderx has the final say in the matter

 - Make the billing system org based. Here is what I mean by that:
 
 - Each org should have it's own billing section. If you are invited to more then 1 org, you can use the `OrganizationDropdown` to change the billing section.
  - A new role should be create. `Super admin`. This will allow for changing the billing (the documentation should be changed accordingly)
  - The billing popout should be removed/not shown from the main dashboard. We can just say `trail over`/`billing issue` in apps table.
  
  - The billing popout should be shown ONLY when you click on the app
  
  Here is what I mean
  
![image](https://github.com/Cap-go/capgo/assets/50914789/1b952611-0845-40a1-b2ca-e0ac8def81ad)

 - The element nr 1 should NEVER be shown on this main dashboard
 - After clicking on the app (`nr 2`) you should be able to see this billing popout (see next screenshot)
 - In the area (`nr 3`) a new column should be created. There we can state the state of the billing (`trail over`/`billing issue`)
 
 Once you click on the app this billing popout should be visible/stay as is

 
![image](https://github.com/Cap-go/capgo/assets/50914789/82d4bcd5-0c41-4f59-a176-4a3cc25c0568)

 - The billing setting should be moved to the orgs settings, instead of the user settings

kremedev:
[...]

PR #561, opened 2024-02-11
Fix: Current tier/trial banner with multiple orgs

WcaleNieWolny:
## Summary

This PR solves the issue https://github.com/Cap-go/capgo/issues/526. It does so by making the following changes:

The banner will NEVER appear on this screen:
image

A banner will appear for the appropriate trial/tier on the app page. What I mean by that is that if user A own app `com.demo.app` and invites user B that owns `com.apple.stock` and user A has a valid plan but user B does not when user B clicks on `com.demo.app` he will see the banner that user A would see.

For the owner it will look something like this:
image

But for the invited user it will look like this:

image

As you can see, there is a difference. This is because the user B (the invited one) CANNOT upgrade the tier of the organization owner. If you click on the information icon you will be informed about it.

image

If, however you still try to upgrade the capgo tier you will be warned that it's only going to affect the apps you own

image

This warning does not appear if you are not invited to any organizations

image

## Screenshots

See above

## Checklist

- [x] My code follows the code style of this project and passes `bun run lint-backend && bun run lint`.
- [x] My change requires a change to the documentation.
- [ ] I have [updated the documentation](https://github.com/Cap-go/website) accordingly. (TODO)
- [x] I have tested my code manually, and I have provided steps how to reproduce my tests 

/claim #526

### DIFF
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -323,7 +323,8 @@ plan-inactive: Plan inactive
 plan-is-the-best-pla: plan is the best plan for your needs.
 plan-maker: Best for small business owners
 plan-member: plan member.
-plan-page-warn: Buying a plan will not affect the organizations your are invited to
+plan-page-warn: Buying a plan will not affect the organizations your are invited to.
+plan-page-warn-2: Learn more about it.
 plan-payasyougo: Best for scalling enterprises
 plan-pricing-plans: Pricing Plans
 plan-progressive-deploy: Progressive deploy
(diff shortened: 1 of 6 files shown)


WcaleNieWolny:
**DO NOT MERGE UNTIL THE DOCUMENTATION CHANGES ARE COMPLETE**

WcaleNieWolny:
Documentation pr: https://github.com/Cap-go/website/pull/178

image


Ready to review @riderx

riderx:
I don’t understand on the home we see only the apps of the org, so why do hide the modal at that time ?

WcaleNieWolny:
> I don’t understand on the home we see only the apps of the org

No we do not - the org switcher does not do ANYTHING unless you are in settings

riderx:
Wait what ?!?
That a huge issue

PR #562, opened 2024-02-13
Fix: Make stats faster

WcaleNieWolny:


supabase/functions/_backend/private/plugins/stats.ts
```diff
@@ -69,6 +69,23 @@ export const jsonRequestSchema = z.object({
   message: INVALID_STRING_DEVICE_ID,
 })
 
+function requestData(c: Context, app_id: string, version_name: string) {
+  const appOwnerRaw = supabaseAdmin(c)
```
riderx, line 73: There are many calls to stats made by plugin not using Capgo cloud, so this will just create 2 requests instead of one in that case.
Not sure it's better approach

supabase/functions/_backend/private/plugins/stats.ts
```diff
@@ -212,7 +204,7 @@ async function post(c: Context, body: AppStats) {
       }, 200)
     }
     rows.push(stat)
-    await Promise.all([sendDevice(c, device).then(() => sendStats(c, rows))])
+    await Promise.all([sendDevice(c, device), sendStats(c, rows)])
```
riderx, line 207: i think the then was for a reason, but i couldn't find why anymore, so we can try

riderx: maybe it was for the stats it was spamming otherwise

supabase/functions/_backend/private/plugins/stats.ts
WcaleNieWolny: I don't think there is a case when you send only 1 request - i checked the old code

riderx:
not require anymore i changed the way we send to clickhouse

PR #575, opened 2024-02-17
Orgs v2

WcaleNieWolny:
## Summary

This PR implements the organization system v2. Here is a more detailed list of what changed:

 - Resource ownership was moved from `user_id` into `org_id`
 - RLS was rewritten to reduce complexity + reflect the change of ownership
 - The `owner` role was removed and replaced with `super_admin`
 - The main app dashboard was changed in the frontend to reflect the change of app
 - Stats display was fixed for orgs users
 - CLI was changed to make sure org users have the same permission as the owner
 - The REST HTTP api was changed to allow org users to access it
 - Some frontend bugs were fixed

## Test plan

This PR does not really have a test plan. I tested this manually but this method was approved by @riderx 

### Frontend changing

https://github.com/Cap-go/capgo/assets/50914789/13a62d0d-f4da-43f3-866d-01ffeec01967

(By default the org with the greatest number of apps is loaded)

### New members panel

image

## Checklist

- [x] My code follows the code style of this project and passes `bun run lint-backend && bun run lint`.
- [ ] I have [updated the documentation](https://github.com/Cap-go/website) accordingly.
- [x] My change has adequate E2E test coverage. (Partial E2E coverage - i changed the old orgs e2e tests to work with this)
- [x] I have tested my code manually, and I have provided steps how to reproduce my tests


WcaleNieWolny:
Ready for review @riderx
Some small changes will still be made but not anything to big. I will send a msg when it's fully ready for a merge

supabase/migrations/20240212005746_seed_db.sql
riderx, line 1: i don't think we should modify the file better to create a new seed file

WcaleNieWolny: I strongly disagree. It will create a dumpster fire in the migration files as the seed is updated fairly regular.  
Also: It's a function that is meant to be **NEVER** user in prod. 

I would drop this seed fn in prod and exclude this migration for ever being applied into prod but that is a future concert.

If this function is supposed to be used only once - during supabase seeding then there really is no point in making sure this function is up to date after the seed operation is complete. 

I would much rather not update the function and have a clean code base then make sure it's always up to date even if it's never going to be used

riderx: agree

PR #576, opened 2024-02-18
Mau fixes v3

WcaleNieWolny:
## Summary

This plan fixed mau - this time for good.

## Test setup

This PR was tested A LOT with A LOT of edge cases. The setup for this looks like this

image

Let me break this down day by day:

 - 18th: 1 new device was created for `com.demo.app` (total mau: 1)
 - 19th: 3 new devices were created for `com.demo.app` (total mau: 4)
 - 20th
   - 2 new devices were created for `com.demo.app` (total mau: 6)
   - 5 new devices were created for `com.demoadmin.app` (total mau: 11)
  - 21th: No new devices were created, total mau: 11
  - 22th: One new device was created for `com.demo.app`
    - total mau: 7 (from `com.demo.app` and 5 from `com.demoadmin.app` = 12
  - 23th: 
    - 1 new device was created for `com.demo.app`
    - 1 new device was created for `com.demoadmin.app`
    - total mau: 14 (8 from `com.demo.app` and 6 from `com.demoadmin.app`
  - 24th: 1 new device was created for `com.demo.app` - total mau: 15
  
  This however is an edge case. Let's assume `22th` is the last day of the users plan. How do we calculate the MAU?
  Obviously we have to take the mau for `com.demo.app` from the 23th and add it with mau for `com.demoadmin.app` from `22th`. 
  
 And this is what the new `get_total_stats_v4` does - exactly that.
 
 Now, I have disabled the graph on the main dashboard
 
image

Let me show you why:

image

Here on the 22th the mau goes down to 7. Why? Because on the graph sees that on the `22th` mau was 7 for `com.demo.app`, it does not see that no new devices were created for `com.demoadmin.app` so the value HAS TO BE the previous one. From the graph's perspective the `com.demoadmin.app` is not in the dataset for this specific day so it does not add it properly. Since mau CANNOT go down I had to disable this graph

 Here is a detailed migration guide for prod:
  - Rename `clickhouse_app_usage` into `clickhouse_app_usage_tmp`
  - Apply the migration that is included in this PR

### DIFF
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -404,6 +404,7 @@ thank-you-for-choosi: Thank you for choosing Capgo !
 the: The
 this-page-will-self-: This page will self refresh
 title: title
+tmp-disabled: (Temporary disabled)
 to: to
 to-live-reload: to live reload.
 to-open-encrypted-bu: 'To open Encrypted bundle use the command:'
--- a/src/components/dashboard/Usage.vue
+++ b/src/components/dashboard/Usage.vue
@@ -111,10 +111,16 @@ loadData()
 <template>
   <div v-if="!noData || isLoading" class="grid grid-cols-12 gap-6 mb-6" :class="appId ? 'grid-cols-16' : ''">
     <UsageCard
-      v-if="!isLoading" id="mau-stat" :limits="allLimits.mau" :colors="colors.emerald"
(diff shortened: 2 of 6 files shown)


WcaleNieWolny:
As for the `mau_final_param` view - it is a parametrized query because previously the filters were added in an incorrect place. In the mau v2 I did not catch that because I never tested with the proper amount of data. Since mau v2 I saw a lot more edge cases and I tested all of them in this pr

PR #655, opened 2024-05-03
🐛 (fix): Fix #642 (Delete account should stop subscription)

### LINKED ISSUE #642: Delete account should stop subscription
**Describe the bug/issue**
I had to refund someone today who deleted his account and received an email he will be charged.
And realized we never remove subscription in that case

WcaleNieWolny:


riderx:
Thanks for the work, before accept I think there is thing to improve:
Check org delete, does more than a check so we should find a better name to reflect that.
This function use get identity, who is not what we want, it should use the user id of the row deleted.
We want the function work in situation where the user self delete his account or when we do it from database ourself.
So please update the function

riderx:
Can we call the function something like,
Delete_empty_org ?
And do the switch to org user instead of users

WcaleNieWolny:
@riderx i changed this trigger to make it more versatile. Can you please review again?

PR #666, opened 2024-05-07
🐛 (fix): Fix count_all_onboarded to work with orgs

### LINKED ISSUE #658: Fix logsnag integration
**Describe the bug/issue**
With the new org system the tracking of onboarding, subscriptions conversion is dead.
We need to fix this tracking is important to understand how to improve

WcaleNieWolny:
fixes #658 

PS: i don't know how to fix the subscription count as i don't know what is broken with counting the subscriptions


riderx:
Currently what is broken is the chart who follow the whole flow of user.
So it should map from event onboarding to subscription.
I think the first event are in CLI so be sure the CLI is also sending the event with the right user ID

riderx:
CleanShot 2024-05-07 at 21 44 56@2x
Here is the chart and the event it watch

riderx:
it's a good fix on it's own

PR #693, opened 2024-05-27
✅ (chore): Test plan calculation math

WcaleNieWolny:


supabase/tests/13_test_plan_math.sql
```diff
@@ -0,0 +1,71 @@
...
+  RETURN NEXT IS(usage.bandwidth_percent, (SELECT CAST ('1.0' AS DOUBLE PRECISION)), 'Bandwidth usage = 1% for "Solo" plan');
+
+  -- Let's now add a second app to this org. 
+  ALTER TABLE app_versions DISABLE TRIGGER force_valid_owner_org_app_versions;
+
+  UPDATE apps set owner_org='046a36ac-e03c-4590-9257-bd6c9dba9ee8' where app_id='com.demoadmin.app';
+
+  UPDATE app_versions set app_id='com.demoadmin.app' where id=9601;
+  INSERT INTO "public"."daily_mau" ("app_id", "mau", "date") VALUES 
+  ('com.demo.app', 10, (now() - interval '1 day')::date);
```
riderx: Should be here the second app I think

WcaleNieWolny: Fixed in the latest commit

Issue #684, opened 2024-05-16
Refactor minUpdateVersion to follow db convention


riderx:
**Describe the bug/issue**
This field is not following the convention and should follow it.
This has to be change in CLI as well
All fields who should be changed:
- minUpdateVersion
- enableNotifications
- optForNewsletters
- legalAccepted
- disableAutoUpdateUnderNative
- enableAbTesting
- secondaryVersionPercentage
- secondVersion
- disableAutoUpdate

Migration should be written, and CLI should be updated to work with both name.
Public API should still return the old name if it was exposed

WcaleNieWolny:
`optForNewsletters`, `enableNotifications` are stored in `auth.users` - they are hard to change without affecting users
![image](https://github.com/Cap-go/capgo/assets/50914789/7ef0144f-59ec-4566-ba95-0563f01964aa)

`legalAccepted` is never used, we can drop it

riderx:
legalAccepted, is required for legal aspect nothing in code require it ^^

riderx:
ok i checked the code and yes only legal is used.
optForNewsletters, enableNotifications can stay the same in auth as they are in json obj, but they have to be changed in users table

PR #735, opened 2024-06-29
Fix db convention

WcaleNieWolny:


WcaleNieWolny:
https://github.com/Cap-go/CLI/pull/304

supabase/migrations/20240629035932_fix_fields_db_convention.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE channels 
+RENAME COLUMN "disableAutoUpdate" TO disable_auto_update;
+
+CREATE PROCEDURE "public"."update_channels_progressive_deploy"()
+    LANGUAGE "plpgsql"
+    AS $$
+BEGIN
+    UPDATE channels
+    SET "secondary_version_percentage" = CASE 
+        WHEN channels."secondVersion" not in (select version from stats where stats.action='update_fail' and 10800 > extract(epoch from now()) - extract(epoch from stats.created_at))
```
riderx: secondVersion is still used here

locales/en.yml
```diff
@@ -106,6 +106,7 @@ button-toggle_dark: Toggle dark mode
 button-toggle_langs: Change languages
 by-clicking-on-them: by clicking on them.
 cancel: Cancel
+canceled-delete: Canceled bundle deletion, cannot delete a linked bundle
```
riderx: Missing translations in other languages

supabase/functions/_backend/public/channel.ts
```diff
@@ -146,8 +146,8 @@ export async function post(c: Context, body: ChannelSet, apikey: Database['publi
     app_id: body.app_id,
     name: body.channel,
     ...(body.public == null ? {} : { public: body.public }),
-    ...(body.disableAutoUpdateUnderNative == null ? {} : { disableAutoUpdateUnderNative: body.disableAutoUpdateUnderNative }),
```
riderx, line 149: the object returned shouldn't change, otherwise the user will have they process broken

supabase/migrations/20240629035932_fix_fields_db_convention.sql
```diff
@@ -0,0 +1,32 @@
+ALTER TABLE app_versions 
+RENAME COLUMN "minUpdateVersion" TO min_update_version;
+
+ALTER TABLE channels
+RENAME COLUMN "disableAutoUpdateUnderNative" TO disable_auto_update_under_native;
+
+ALTER TABLE channels
+RENAME COLUMN "enableAbTesting" TO enable_ab_testing;
+
+ALTER TABLE channels
```
riderx, line 10: missing changes in users tables, as say here https://github.com/Cap-go/capgo/issues/684

supabase/functions/_backend/public/channel.ts
```diff
@@ -76,8 +76,8 @@ export async function get(c: Context, body: GetDevice, apikey: Database['public'
         created_by,
         updated_at,
         public,
-        disableAutoUpdateUnderNative,
-        disableAutoUpdate,
```
riderx, line 80: Still issue here you cannot change what the public API returns

WcaleNieWolny: fixed in the latest commit

supabase/functions/_backend/public/channel.ts
WcaleNieWolny: I don't think I follow. It returns a [BRES/empty response](https://github.com/Cap-go/capgo/blob/0052d865610e243991fd14ba45396917f9c7a5e5/supabase/functions/_backend/public/channel.ts#L209) as this is a POST request. It still accepts the same body parameters as before.

supabase/migrations/20240629035932_fix_fields_db_convention.sql
WcaleNieWolny: it was agreed to postpone this users table change

supabase/migrations/20240629035932_fix_fields_db_convention.sql
WcaleNieWolny: fix in the latest commit

PR #774, opened 2024-08-06
Open api v2

WcaleNieWolny:


cloudflare_workers/index.ts
```diff
@@ -1,4 +1,5 @@
-import { Hono } from 'hono/tiny'
```
riderx, line 1: keep tiny here, OpenAPIHono is not required here

riderx:
fix conflict and import as suggested in comment and you can merge yourself

cloudflare_workers/index.ts
WcaleNieWolny: Yes it IS required. 

I call the following
- `app.openAPIRegistry`
- `app.doc`

PR #777, opened 2024-08-07
🔒️ (feat): Add bundle signing in the backend

WcaleNieWolny:


riderx:
Please make signature return only from a certain version of plugin

WcaleNieWolny:
> Please make signature return only from a certain version of plugin

image

Already doing that

PR #801, opened 2024-08-20
✨ (feat): Partial v3 (WiP)

WcaleNieWolny:


riderx (approved):

supabase/functions/_backend/private/partial_upload.ts
```diff
@@ -0,0 +1,129 @@
...
+    }),
+    hash: z.string().refine(val => hexRegex.test(val), {
+      message: 'The manifest hash does not match the HEX regex (/[0-9a-fA-F]+/)',
+    }),
+  }).array().max(10),
+})
+
+export const app = new Hono()
+
+app.post('/v1', middlewareKey(['all', 'write', 'upload']), async (c: Context) => {
```
riderx: do not use v1, we are going to use headers

WcaleNieWolny: A) You CANNOT route based on headers
B) Going to do that when?

WcaleNieWolny:
Rebased and fixed issues in: https://github.com/Cap-go/capgo/pull/818

PR #802, opened 2024-08-20
✨ (feat): Add a new plan failed popout

### LINKED ISSUE #800: Improve UI payment failed
**Describe the bug/issue**
Currently payment fail doesn’t stand in the UI so user miss it.
This should be in red and bold to make sure user see it :)
IMG_1391

WcaleNieWolny:
image

Fixed #800


riderx:
If display as banner we need a button in it then.
And be careful in mobile app the button shouldn’t be displayed are we are not allowed to do it

WcaleNieWolny:
> we need a button in it then

What button? What would it do?

PR #803, opened 2024-08-20
🐛 (fix): Fix RLS for bundle/channel overwrite

### LINKED ISSUE #790: devices_override and channel_devices lack update RLS policies
Both `devices_override` and `channel_devices` lack update RLS policies.

This causes the following problem:

1. Force device A to version 1.0.1
2. Try to force device A to version 1.0.2
3. Error will happen because you cannot update an overwrite directly

## WARNING

I have manually added an update RLS policy for `devices_override` in the production database

WcaleNieWolny:
Fix #790 

# Requires a manual update to the production db


riderx:
Why it does require manual?

WcaleNieWolny:
Because I have created the following policy manually in order to fix a problem that one used had

image

It requires removing this policy before merging

PR #818, opened 2024-08-27
✨ (feat): Partial upload (v2)

WcaleNieWolny:
Rebased #801 with fixed issues mentioned in #801

### DIFF
--- a/cloudflare_workers/index.ts
+++ b/cloudflare_workers/index.ts
@@ -34,6 +34,7 @@ import { app as latency_postres } from '../supabase/functions/_backend/private/l
 import { app as latency_drizzle } from '../supabase/functions/_backend/private/latency_drizzle.ts'
 import { app as multipart } from '../supabase/functions/_backend/private/multipart.ts'
 import { app as create_device } from '../supabase/functions/_backend/private/create_device.ts'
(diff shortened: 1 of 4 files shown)


riderx:
why do you rebase and recreate a PR each time?

WcaleNieWolny:
I rebased and squashed because I felt like learning rebase.

From my understanding:
1. You rebase to avoid a 3-way merge commit.
2. Rebasing allows you to do a fast-forward merge
3. Squashing helps with the git commit history. It removes nonsensical commits that made sense in the scope of the PR but that do not make sense in the scope of the main branch
4. You NEVER rewrite history for a branch that was already shared with a co-worker, thus I created a new PR

riderx:
By doing that, we lose the full history of the conversation.

The best way to do what you want for me is keep pushing commit on the same branch he resolves your issue and squash to remove the history of unwanted changes.
And this is ok to do it on your current branch without switching as no one else of you should have uncommit changes on this branch.
So people will just force pull if they did pull before.
Please do no create multiple branches and PR for the same task in the future. The history of the conversation is essential to me.

WcaleNieWolny:
So, in practice, you are saying that I CAN squash commits but I just have to force push instead of creating a new branch?

PR #861, opened 2024-10-11
feat: captcha at login

WcaleNieWolny:


riderx:
Make capchat optional if VITE_CAPTCHA_KEY is undefined then do not import and try to run the capchat

WcaleNieWolny:
> Make capchat optional if VITE_CAPTCHA_KEY is undefined then do not import and try to run the capchat

it;s okay - captcha is not possible to configure locally, and I think that supabase will just ignore captcha if you don't configure it. Furthermore, the default captcha key for local is a test only key from turnstile. This key was designed specifically to be a placeholder

PR #883, opened 2024-11-06
Fix retention

WcaleNieWolny:


riderx:
your PR add update_app_versions_retention but i don't see it in the code and i dont see it used also.
Did something is missing?

PR #899, opened 2024-11-26
feat: rate limit the public api

WcaleNieWolny:


cloudflare_workers/api/wrangler.toml
```diff
@@ -44,3 +44,51 @@ d1_databases = [
...
+simple = { limit = 20, period = 10 }
+
+[[unsafe.bindings]]
+name = "API_BUNDLE_DELETE_RATE_LIMITER"
+type = "ratelimit"
+namespace_id = "1002"
+simple = { limit = 20, period = 10 }
+
+[[unsafe.bindings]]
+name = "API_CHANNEL_GET_RATE_LIMITER"
```
riderx, line 61: This belongs to the plugin API wrangler file, not here

riderx:
missing limit on file endpoint as well

WcaleNieWolny:
> missing limit on file endpoint as well

Tell me please, what should be the key for the file endpoint

riderx:
we added in prod device_id for files so you can now use it

riderx:
i think the rate limit is missing wrangler config

WcaleNieWolny:
> i think the rate limit PR is missing wrangler config

it's not. You need to set the rate limits in the code, and than use the newly created script (`scripts/generate_rate_limits.ts`) to generate the rate limits for the wrangler config.

I have shared the command with you privately, but I will share it again here:
```bash
bun run scripts/generate_rate_limits.ts ./cloudflare_workers/plugin/index.ts >> cloudflare_workers/plugin/wrangler.toml
```

riderx:
Why generating them while all the wrangler config is written in the config file?

riderx:
Add also a log in stats about rate limiting, it's important for the user to be able to see it

WcaleNieWolny:
@riderx https://github.com/Cap-go/capgo/pull/899/commits/bc7c56ccfd56eba3546f4d22714b681a96743bc6

In this commit I added a feature where when the device is rate-limited in `/updates` or `/channel_self` the backend will send a stat

WcaleNieWolny:
I will add rate limiting for the files worker

riderx:
2 points:
- i would like to have this merged this week
- rate limit should be disabled properly in case of Supabase or other not supported platform

WcaleNieWolny:
> rate limit should be disabled properly in case of Supabase or other not supported platform

It's never enabled in the first place. If you take a close look at the code, it's only enabled for workers.

riderx:
Can you fix the conflict, and add the file limit as said?
any limit reach should show up on stats table for user to realize it happen

WcaleNieWolny:
I absolutely can add the ratelimit to files, but I would strongly appreciate if responded to my concerns about limiting /files

image

WcaleNieWolny:
> any limit reach should show up on stats table for user to realize it happen

This isn't really possible for `/api`. API doesn't have a device ID, and as such, `/API` cannot report stats

PR #914, opened 2024-12-06
Plans fix

WcaleNieWolny:


WcaleNieWolny:
https://github.com/user-attachments/assets/f5b6c6e4-42d3-40cf-ad7b-b2a3ad5bf870

riderx:
Thanks that great, in the demo i don’t see the upgrade (situation and text)
And the current button doesn’t seems disabled it should be less visible grayed.

WcaleNieWolny:
@riderx this SHOULD be mergable

PR #915, opened 2024-12-07
fix: revert to native

WcaleNieWolny:


riderx:
We need to work on this again after the features for the client

WcaleNieWolny:
What do you want me to do with this PR? I think this is sort of finished. You revered it because you never reviewed this PR, and you merged it by accident. I think that aside from the conflicts, it's finished

PR #925, opened 2024-12-12
Api key limit org

WcaleNieWolny:


riderx (commented):
why the check is not in middleware?

supabase/functions/_backend/utils/downloadUrl.ts
```diff
@@ -78,7 +78,7 @@ export async function getManifestUrl(c: Context, version: {
   }
 
   try {
-    const url = new URL(c.req.url)
+    const url = new URL('http://wolny-ubuntu.local:7887')
```
riderx: this shouldn't be commit

riderx (commented):
i really believe keeping the current key system with just adding org is a bad idea, because it directly block us to do the next step where we want to allow user and apikey to have ressources access with precision.
in the case of just doing org then i want to make sure upload key work correctly and report error correctly when try to do not allowed action

WcaleNieWolny:
> where we want to allow user and apikey to have ressources access with precision

This all sounds great, until you consider the implications of this change. When I was designing the organization's system, I did think about this. Going one or even two levels down in the "precision access" shouldn't be difficult. If you want to limit a user to a specific app within the organization, we already have the code to do that. 

If you want to limit an API key to a specific app, then we have a huge problem.

If you really, really want to have precise access, I think it's important to consider the drawbacks and the advantages before implementing it.

Advantages:
- Precise resource access is important for enterprise customers.
- Improved user experience with upload API keys, et cetera
Disadvantages:
- Development time. Developing an RBAC-like system would require weeks. It would be a complete rewrite of the RLS, orgs, and API keys. Testing would take months. The complexity far exceeds everything I have ever worked on so far, including orgs V1

riderx:
What about keeping the rights system we have and extend it ?
Like allow rights as a list and orgID as a value of it ?
Then later we would add specific rights, like edit channel, edit org, read log or whatever?

WcaleNieWolny:
At some point, you will realize that the mess we created is untenable. I am sort of against adding this into prod. In a year, all of this code will be legacy code. The more difficult and complex the permission system is, the harder it will be to maintain it.

WcaleNieWolny:
We have a somewhat good balance between complexity and functionality right now. This PR tips this balance slightly without flipping over the scale.

The more complex the security system, the more likely it is we screw it up.

riderx:
I agree with you that our current system is not great, open to enterprise customer is our target so we will need to improve this at one point.
let's accept this like that ( after new year ) and we look at that later on

riderx (approved):

supabase/functions/_backend/utils/downloadUrl.ts
riderx: still here

PR #935, opened 2024-12-16
chore: remove old fields

WcaleNieWolny:


riderx:
test will fail if you not update them :)

PR #940, opened 2024-12-25
feat: stats via backend

WcaleNieWolny:


riderx:
First, I would like to thank you for the work, I find that we have ease to work and I see the progress you had in understanding my unprescise requirement and shaping your work.  I have some design questions and some proposal about this PR  Current local seems different that main with demo data. Please investigate why all chart seems the same where this was not the case before 
why is checkApikeyAuth inner declared? We never do this, is a bad practice in javascript that why eslint complain, please do not suppress it unless you justify this choice, plus we already have a middleware function who does can do this check why not use it? 
d3 is import with * we avoid doing that for unless there are no other ways.  Please justify if it that mandatory. the best would be to do a separate import to be sure it's not even fechted in Deno. Endpoint are not declared in Cloudflare and Netlify, please add them. statistics endpoint is plural where we use singular name, please update it/ 
for demo purpose, seed data should be changed to have meanfull mobile stats chart in local. 
Moving getDateRange outside loadData means if the user never keeps Capgo open more than a day,he will never data, please justify this choice.  The goal of this front change is to make it display only not transform.
Convert function and zero filling should be done server side and be able to switch off by param url.

WcaleNieWolny:
> I see the progress you had in understanding my unprescise requirement and shaping your work

Thank you for saying that; however, the requirements you gave me were very precise.

> Current local seems different that main with demo data

Do you mind asking how you test locally? From my experience, this PR is almost untestable locally. I would put this PR to preprod and then compare the graphs.

Also, as we have established privately, the current storage bandwidth graphs have terrible rounding errors. Naturally, this means that in this PR these graphs will differ slightly, but the general curve and shape should remain consistent. Also, if you have a specific app for which you observed this behavior, please share the app ID privately.

> why is checkApikeyAuth inner declared? [...], is a bad practice in javascript that why eslint complain

ESlint doesn't complain. Deno does, and since I use the deno plugin in VSCode and this warning annoyed me, I supressed it.
With that said, I do agree that the entire auth system for the new endpoints implemented in this PR is quite badly written.
Originally, I did not plan for the usage of these API endpoints on the frontend. I also believe that this might be the first time we have an endpoint that accepts both APIKEY and access_token (JWT). There are a lot of improvements that could be made to it, but I decided that as long as it is functional, it's good enough. I prioritized usefulness over niceness, just as you taught me. 

> d3 is import with * we avoid doing that for unless there are no other ways.  Please justify if it that mandatory

I don't think it is. I will change it.

> Endpoint are not declared in Cloudflare and Netlify, please add them.

Will do.

> statistics endpoint is plural where we use singular name, please update it

I sort of don't agree with this. I like the plural, as the endpoint clearly provides MULTIPLE statistics. (mau, bandwidth, storage, bundle_usage). Also, I wanted to distance this endpoint from `/stats` (the plugin endpoint for sending stats to capgo)



> for demo purpose, seed data should be changed to have meanfull mobile stats chart in local

How? I would really appreciate your guidance on this. I have blindly copied the bundle_usage generation code from frontend to backend, and I have not really investigated how it works. Would you be able to assist me in completing this task?

> Moving getDateRange outside loadData means if the user never keeps Capgo open more than a day,he will never data, please justify this choice

Fixed in https://github.com/Cap-go/capgo/pull/940/commits/3f936ed4f822fa3d869877929b738057c451cc5b

> The goal of this front change is to make it display only not transform.
Convert function and zero filling should be done server side and be able to switch off by param url.

I STRONGLY disagree with this. There are a million ways a consumer of the `/statistics` API may go about drawing a graph. Perhaps they use a JS library like D3, or as our frontend does — Chart.js. Perhaps they use a Python library.
It's alright if the consumer of the API must do some transforming in order to fit the data into their use case. The only real transformation that I am doing is transforming storage and bandwidth from bytes to GB. I am also adding colors into the dataset for `bundle_usage`. Again, this shouldn't be done on the backend side, as this is specific to our frontend.

Transformations are inevitable. They are required to provide a coherent experience from all consumers of the `/statistics` API.

PR #950, opened 2025-01-01
feat: drop legacy triggers

### LINKED ISSUE #937: remove auto_owner_org_by_app_id
**Describe the bug/issue**
This function is against the philosophy of Capgo, RLS should never modify row unless its field always set by db like updated_at.
This should be removed in favor of always providing it explicitly

WcaleNieWolny:
Fixes #937


WcaleNieWolny:
Fixes #937

PR #952, opened 2025-01-04
feat: base_disable_auto_update_on_native

WcaleNieWolny:


WcaleNieWolny:
### **WARNING** The migrations for D1 and supabase are ALREADY deployed in prod !!!!

riderx:
Why d1 and supabase are already deploy?
This PR will not be merged as this should be in the product and not optional

PR #958, opened 2025-01-05
fix: disableAutoUpdate use native instead of version_name

WcaleNieWolny:


riderx:
Can you check the test ? and add a test for this maybe?

PR #961, opened 2025-01-05
chore: resolve TODOs

WcaleNieWolny:


supabase/functions/_backend/utils/stripe.ts
```diff
@@ -111,11 +111,11 @@ export async function createCheckout(c: Context, customerId: string, reccurence:
       name: 'auto',
     },
     tax_id_collection: { enabled: true },
-    // TODO: find why this is not working as expected
```
riderx, line 114: did you tested it?

riderx: @WcaleNieWolny

PR #966, opened 2025-01-12
feat: fork ratelimit to log instead of blocking wih 429

WcaleNieWolny:


riderx:
we drop rate limit for now

PR #996, opened 2025-01-29
feat(WiP): tokens

WcaleNieWolny:


riderx:
@WcaleNieWolny i think we will go for credits instead of token

riderx:
Another thing
CleanShot 2025-06-06 at 10 35 57
In expo they do have the same price for all plan but lower plan have higher price first, they sub the plan number to the price to now the tier we should do the same

riderx:
To me all plan should give credits, no more different price structure.
That means we could should allow to not pay any plan and just buy some credit for tiny apps

PR #1371, opened 2026-01-06
fix: add RLS policy for service_role on stripe_info table

riderx:
## Summary
- Adds RLS policy allowing `service_role` full access to `stripe_info` table
- Fixes Stripe webhooks failing to update customer subscription data

## Problem
The `stripe_info` table has RLS enabled but only had a SELECT policy for org members. When Stripe webhooks attempt to update subscription information via `supabaseAdmin()` (which uses `service_role`), the UPDATE operations were being blocked by RLS.

## Solution
Added a new policy `Allow service_role full access to stripe_info` that grants service_role full access to the table.

## Changes
- `supabase/migrations/20260106181654_stripe_info_service_role_rls_bypass.sql` - New RLS policy
- `supabase/tests/26_test_rls_policies.sql` - Updated test expectations


riderx:
Closing - the RLS policy change was likely not the correct fix. service_role should already bypass RLS for direct table operations. Need to investigate the actual root cause.

PR #1430, opened 2026-01-13
Delta manifest: reduce update response sizes by 75-85%

riderx:
## Summary (AI generated)

Implement server-side manifest delta computation to reduce response sizes by 75-85% for apps with large manifests. Only returns files that are new or changed compared to device's current version using an optimized NOT EXISTS subquery.

## Motivation (AI generated)

Large apps (e.g., Codere with 3,253 files, ~1.4MB response) were returning full manifests even though devices already had most files from the previous version. Delta manifests reduce bandwidth and improve update performance.

## Test Plan (AI generated)

- [x] All existing manifest update tests pass (4 tests)
- [x] All existing general update tests pass (32 tests)
- [x] Linting passes (eslint)
- [x] Backwards compatibility: gracefully falls back to full manifest for first installs, unknown versions, and missing old versions
- [x] Delta filtering uses existing database index for performance

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Delta manifest support: updates return only changed or new files when updating from a prior version, with a full-manifest fallback if no prior version is found.
  * Version-name lookup API: resolve a version name to its numeric ID (returns null for builtin/unknown or on failure).
  * Early no-new-version handling: detects and returns a no-new-version response earlier in the update flow.

* **Tests**
  * Added end-to-end tests covering delta manifest scenarios (unchanged, changed, new files, missing old versions).

### DIFF
--- a/tests/updates-manifest.test.ts
+++ b/tests/updates-manifest.test.ts
@@ -310,7 +310,7 @@ describe('delta manifest scenarios', () => {
     await resetAppDataStats(DELTA_APPNAME)
   })
 
-  it('returns delta manifest with only changed/new files', async () => {
+  it.concurrent('returns delta manifest with only changed/new files', async () => {
     // Request update from old version (1.359.0) to new version (1.0.0)
     // Should only return file_b.js (changed hash) and file_c.js (new file)
     // file_a.js should be excluded because it has the same hash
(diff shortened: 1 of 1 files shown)


riderx:
this can't be done yet we need to allow it first in the plugin code

Issue #1667, opened 2026-02-24
security issues retribution


riderx:
**Describe the bug/issue**

riderx:
/tip $70 @confirmation_url for Open redirect vulnerability via confirmation_url

riderx:
/tip $100 cyberboy225588-coder for Potential Privilege Escalation via SECURITY DEFINER function apply_usage_overage

riderx:
/tip $70 @Cyberheroes1010 for EXIF METADATA IN App informations Image Upload

riderx:
/tip $100 @cyberboy225588-coder for Potential Privilege Escalation via SECURITY DEFINER function apply_usage_overage

riderx:
/tip @Cyberheroes1010 $50 for 2FA missing recuperation method

lucky7323:
## Security Vulnerability: RBAC Role Binding Privilege Persistence After Member Removal

### Summary
When a member is removed from an organization via the member deletion endpoint, their `role_bindings` entries are **not cleaned up**, allowing the removed user to retain RBAC permissions.

### Affected Code
**`supabase/functions/_backend/public/organization/members/delete.ts`**

The `deleteMember` function only deletes from `org_users`:
```typescript
const { error } = await supabase
  .from('org_users')
  .delete()
  .eq('user_id', userData.id)
  .eq('org_id', body.orgId)
```

No corresponding cleanup of `role_bindings` is performed.

### Impact
In organizations with RBAC enabled (`use_new_rbac = true`):

1. Admin removes a member from the organization
2. The `org_users` row is deleted
3. The user's `role_bindings` rows **remain in the database**
4. `rbac_has_permission()` checks `role_bindings` directly (not `org_users`), so the removed user **still passes permission checks**

This means a removed member can continue to access org resources (apps, channels, bundles) as if they were still a member, as long as RBAC mode is active.

### Root Cause
- No foreign key relationship between `org_users` and `role_bindings` (no CASCADE)
- No database trigger on `org_users` DELETE to clean up `role_bindings`
- The deletion endpoint only handles `org_users`, not RBAC artifacts

### Reproduction Steps
1. Enable RBAC for an organization (`use_new_rbac = true`)
2. Invite a user and assign them a role (e.g., `org_admin`)
3. Verify the user has entries in both `org_users` and `role_bindings`
4. Remove the user via the member deletion API
5. Verify `org_users` entry is gone but `role_bindings` entry persists
[...]

PR #1702, opened 2026-02-26
fix(security): sanitize SQL interpolation in Cloudflare Analytics Engine queries

artylobos:
## Summary

Fixes SQL injection vulnerabilities in Cloudflare Analytics Engine queries (GHSA-f83x-p28r-pf74).

User-controlled string parameters were interpolated directly into SQL query strings without sanitization in `readDevicesCF` and `readStatsCF`, allowing an attacker to break out of string literals and inject arbitrary SQL.

## Changes

- Added `escapeSqlString()` utility function that escapes single quotes and backslashes for safe SQL string interpolation
- Applied `escapeSqlString()` to all user-controlled values interpolated into Analytics Engine SQL queries

### `readDevicesCF`
- `params.deviceIds` — single value and IN-list interpolation
- `params.search` — search filter interpolation (both with and without deviceIds)
- `params.version_name` — version filter interpolation
- `params.cursor` — cursor-based pagination (timestamp and device_id components)

### `readStatsCF`
- `params.deviceIds` — single value and IN-list interpolation (the IN-list was critically missing quotes entirely)
- `params.actions` — single value and IN-list interpolation
- `params.search` — search filter interpolation (both with and without deviceIds)

## Critical Fix

The `readStatsCF` IN-list for `deviceIds` was particularly dangerous — values were joined without any quoting at all, making injection trivial.

/claim #1667

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL query sanitization across analytics, reporting, and device/app data retrieval to reduce injection risk and improve reliability.
  * Hardened internal query construction for admin metrics, usage/bandwidth reports, and search/filtering; no changes to public APIs or visible functionality.

### DIFF
--- a/supabase/functions/_backend/utils/cloudflare.ts
+++ b/supabase/functions/_backend/utils/cloudflare.ts
@@ -12,7 +12,7 @@ import { getEnv } from './utils.ts'
 
 /** Escape a value for safe interpolation into an Analytics Engine SQL string. */
 function escapeSqlString(value: string): string {
-  return value.replace(/'/g, "''").replace(/\\/g, '\\\\')
+  return value.replace(/'/g, '\'\'').replace(/\\/g, '\\\\')
 }
 
 // type is require for the bindings no interface


artylobos:
Hi @riderx, thanks so much! 🙏

Just a small note - looks like the tip command had a typo (`&150` instead of `$150`). The bot needs the dollar sign format:

`/tip $150 @artylobos`

Thanks again for the recognition!

artylobos:
Hi @riderx, just a friendly reminder - the tip command had a small typo (ampersand instead of dollar sign). Could you please re-run it? Thanks!

PR #1724, opened 2026-02-28
fix(security): RBAC security audit fixes

Dalanir:
## Summary (AI generated)

Deep security audit of the RBAC system uncovering 6 vulnerabilities, each fixed in a separate commit:

1. **SQL injection in `getStoreAppByIdCF`** — delete as it's not used anywhere
2. **Privilege escalation in `/private/role_bindings`** — users with `org.update_user_roles` could assign roles above their own rank; added `priority_rank` check
3. **`is_assignable` bypass via PostgREST** — RLS INSERT policy didn't check `is_assignable`, allowing direct inserts of non-assignable roles (e.g. `platform_super_admin`)
4. **Orphaned app/channel bindings on member removal** — `delete_org_member_role` only deleted org-level bindings; now cascades all scopes
5. **Last super_admin unprotected via direct DELETE** — added BEFORE DELETE trigger on `role_bindings` to prevent deleting the last `org_super_admin`
6. **Hashed API key lookup failure** — `rbac_check_permission_direct` used `WHERE key = p_apikey` which silently fails for hashed keys; switched to `find_apikey_by_value()`

## Motivation (AI generated)

The RBAC system is well-architected overall, but several edge cases allowed bypasses through direct PostgREST access or inconsistent enforcement between SQL functions and the application layer.

These fixes close privilege escalation vectors that could allow org members to grant themselves admin access, prevent org lockout scenarios, and ensure hashed API keys get proper RBAC permissions. Critical for enterprise customers relying on fine-grained access control.

## Test Plan (AI generated)

- [ ] Verify `getStoreAppByIdCF` still returns correct results with parameterized query
- [ ] Test that a non-super_admin cannot assign `org_super_admin` via POST `/private/role_bindings`
- [ ] Test that direct PostgREST INSERT into `role_bindings` with a non-assignable role is rejected
- [ ] Test that `delete_org_member_role` removes app/channel bindings too
- [ ] Test that deleting the last super_admin binding is blocked by the trigger
- [ ] Test that hashed API keys get RBAC permissions resolved correctly
- [ ] Run `bun test:backend` to verify no regressions

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Enforced privilege checks to prevent creating, updating, or deleting role assignments beyond a caller’s allowed rank; stricter API key creation limits to prevent privilege escalation.

* **Bug Fixes**
  * Cascaded removal of member role bindings; protections to prevent deleting the last org super-admin.

* **Security**
  * Stronger insert-time access controls and enhanced permission checks, including hashed API-key support.

* **Tests**
  * Updated RBAC and org-member tests to validate trigger-driven bindings and deletion behavior.


supabase/functions/_backend/private/role_bindings.ts
```diff
@@ -1,6 +1,6 @@
 import type { Context } from 'hono'
 import type { MiddlewareKeyVariables } from '../utils/hono.ts'
-import { and, eq } from 'drizzle-orm'
+import { and, desc, eq, sql } from 'drizzle-orm'
```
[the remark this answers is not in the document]
Dalanir: Fixed

supabase/migrations/20260228000000_role_bindings_rls_assignable.sql
```diff
@@ -0,0 +1,23 @@
+-- Add is_assignable check to role_bindings INSERT RLS policy
+-- Without this, a direct PostgREST INSERT could bypass the endpoint's is_assignable check
+
+DROP POLICY IF EXISTS "role_bindings_insert" ON "public"."role_bindings";
+
+CREATE POLICY "role_bindings_insert" ON "public"."role_bindings" FOR INSERT TO "authenticated" WITH CHECK (
```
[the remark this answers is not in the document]
Dalanir: All existing role_bindings policies (SELECT, INSERT, UPDATE, DELETE) in prod.sql use only TO "authenticated" — none include anon. This is consistent: role_bindings operations are admin-level and should not be accessible to anonymous users.

The finding is incorrect. API key auth in this codebase works through the authenticated role with JWT, not the anon role. The get_identity_org_appid functions used in the policy already handle API key resolution within the authenticated context. Adding anon would be a security regression.

No change needed — the policy is correct as-is.

supabase/migrations/20260228000000_role_bindings_rls_assignable.sql
```diff
@@ -0,0 +1,23 @@
...
+  -- The role must be assignable
+  (EXISTS (
+    SELECT 1 FROM "public"."roles" r
+    WHERE r.id = "role_bindings"."role_id" AND r.is_assignable = true
+  ))
+  AND
+  (EXISTS ( SELECT 1
+   FROM ( SELECT "auth"."uid"() AS "uid") "auth_user"
+  WHERE ("public"."is_admin"("auth_user"."uid") OR (("role_bindings"."scope_type" = "public"."rbac_scope_org"()) AND "public"."check_min_rights"("public"."rbac_right_admin"(), "auth_user"."uid", "role_bindings"."org_id", NULL::character varying, NULL::bigint)) OR (("role_bindings"."scope_type" = "public"."rbac_scope_app"()) AND (EXISTS ( SELECT 1
```
WcaleNieWolny, line 15: Why do we use `SELECT "auth"."uid"() AS "uid")` and 

`"public"."is_admin"("auth_user"."uid"`

This makes zero sense to me?

supabase/migrations/20260228172309_fix_rbac_test_compatibility.sql
```diff
@@ -0,0 +1,60 @@
...
+  END IF;
+
+  -- Allow deletion if the org itself is being deleted (CASCADE scenario)
+  SELECT EXISTS(
+    SELECT 1 FROM public.orgs WHERE id = OLD.org_id
+  ) INTO v_org_exists;
+
+  IF NOT v_org_exists THEN
+    RETURN OLD;
+  END IF;
```
WcaleNieWolny, line 37: Perhaps I do not know enough about postgres, but I am not sure if this will work?

Its worth checking if the Postgres will actually mark the `orgs` as deleted when this trigger execute

I think logically, it doesn't work because on `org` deletion the org is deleted last, after all the objects that depended on it have already been deleted, but I might be wrong

supabase/migrations/20260228000100_delete_member_cascade_bindings.sql
```diff
@@ -0,0 +1,61 @@
...
+
+  -- Get org owner to prevent removing the last super admin
+  SELECT created_by INTO v_org_created_by
+  FROM public.orgs
+  WHERE id = p_org_id;
+
+  -- Prevent removing the org owner
+  IF p_user_id = v_org_created_by THEN
+    RAISE EXCEPTION 'CANNOT_CHANGE_OWNER_ROLE';
+  END IF;
```
WcaleNieWolny, line 24: This is new, the old system doesn't do this - transferring ownership of an organization is a valid use case. For example, if you own a business that creates mobile apps for others, when you finish working on an app, you might want to transfer its ownership to your client, so that he can handle billing. There are many reasons why you would want to transfer the ownership

Dalanir: I think you're reading this out of its context. It's just preventing the owner to delete himself during the process of deleting members of an org. It won't be played if you're transferring your org to a new owner.

PR #1728, opened 2026-03-01
fix: remove auto-open of demo onboarding modal

ToriChanIntegration:
The demo modal no longer auto-opens when users have no apps. Users can still trigger it explicitly via `?show-onboarding-demo=1` query param.

## Release Notes

* **Bug Fixes**
  * Updated demo onboarding display to only appear when explicitly triggered via query parameter, removing unintended automatic displays based on app count and other conditions.


ToriChanIntegration:
This change has already been implemented in PR #1733 (merged 2026-03-02), which removes the auto-open demo onboarding modal and simplifies `DemoOnboardingGate.vue` to only show when explicitly triggered via `?show-onboarding-demo=1`. Closing as duplicate.

PR #1913, opened 2026-04-15
[codex] Fix IdP-initiated SSO org provisioning

riderx:
## Summary (AI generated)

- provision SSO org membership from a shared frontend helper used by both `/sso-callback` and the auth guard
- provision IdP-initiated SSO sessions before org onboarding redirects, and preserve the existing merged-account sign-out flow
- harden `/private/sso/provision-user` so invite-only memberships are promoted to real memberships and merge flows fail when the org link cannot be guaranteed
- add frontend and backend regression coverage for IdP-initiated SSO, invite promotion, merge membership linking, and non-SSO onboarding fallback

## Motivation (AI generated)

Some customers entering through IdP-initiated SSO were bypassing the `/sso-callback` page, so the org-linking call never ran. The auth guard then saw no selectable org and redirected the user to create an organization even though their managed SSO org already existed. The backend also treated `invite_*` org memberships as completed membership and could silently continue a merge flow without proving the canonical account was linked to the managed org.

This removes a broken first-login path for SSO customers, which reduces onboarding friction for enterprise users and lowers support load on managed SSO setups. It also makes SSO account linking safer by preventing silent partial merges that would strand users without access to their organization.

## Test Plan (AI generated)

- [x] `bunx eslint src/services/ssoProvisioning.ts src/composables/useSSOProvisioning.ts src/modules/auth.ts tests/auth-sso-provisioning.unit.test.ts tests/sso.test.ts supabase/functions/_backend/private/sso/provision-user.ts`
- [x] `bunx vitest run tests/auth-sso-provisioning.unit.test.ts`
- [x] `bun run typecheck`
- [x] `bun run supabase:with-env -- bunx vitest run tests/sso.test.ts`

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Centralized SSO provisioning service and automatic SSO account linking during sign-in, including promotion of invite-only memberships to active roles.

* **Bug Fixes**
  * Improved provisioning error handling to ensure correct navigation outcomes (continue, redirect to login when accounts are merged, or abort) and more reliable membership updates.

* **Tests**
  * Added unit and integration tests validating SSO provisioning flows, membership promotion, and navigation behavior.

### DIFF
--- a/src/composables/useSSOProvisioning.ts
+++ b/src/composables/useSSOProvisioning.ts
@@ -1,6 +1,6 @@
 import type { Session } from '@supabase/supabase-js'
 import { ref } from 'vue'
-import { defaultApiHost, useSupabase } from '~/services/supabase'
+import { provisionSsoUser } from '~/services/ssoProvisioning'
 
 export function useSSOProvisioning() {
   const isProvisioning = ref(false)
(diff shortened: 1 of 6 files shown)


riderx:
@coderabbitai review

riderx:
Addressed the auth-guard review note in 1a3b43591: provisioning errors now abort navigation instead of falling through to org onboarding, and the unit test suite covers that branch.

riderx:
Addressed the merged-session sign-out edge case in 4a318b1b7: the auth guard now aborts navigation if sign-out fails, and the unit tests cover that path.

PR #1944, opened 2026-04-24
fix: guard app ownership transfers

riderx:
## Summary (AI generated)

- add a Postgres trigger guard that blocks direct `owner_org` reassignment on transfer-owned tables
- route legitimate ownership changes through `public.transfer_app()` by setting an internal transfer-only guard flag
- add a regression test covering the raw PostgREST patch attempt and the approved transfer RPC path

## Motivation (AI generated)

The repository advisory [GHSA-v9jp-r5wh-qqcp](https://github.com/Cap-go/capgo/security/advisories/GHSA-v9jp-r5wh-qqcp) shows that callers could patch `public.apps.owner_org` directly through PostgREST, bypass `transfer_app()`, and leave related rows in the previous organization. This change restores the single-owner invariant by making raw ownership changes fail unless they happen inside the dedicated transfer workflow.

This closes a cross-org access bug on a security-sensitive ownership boundary. It reduces the risk of customers exposing version-layer data to the wrong organization and avoids trust damage from inconsistent app ownership state.

## Test Plan (AI generated)

- [x] `bun run supabase:with-env -- bunx vitest run tests/app-transfer-security.test.ts`
- [x] `bun typecheck`

Generated with AI

### DIFF
--- a/supabase/migrations/20260424090125_protect_owner_org_transfer_path.sql
+++ b/supabase/migrations/20260424090125_protect_owner_org_transfer_path.sql
@@ -1,0 +1,163 @@
+CREATE OR REPLACE FUNCTION public.guard_owner_org_reassignment()
+RETURNS trigger
+LANGUAGE plpgsql
+SET search_path = ''
+AS $$
+BEGIN
+  IF NEW.owner_org IS DISTINCT FROM OLD.owner_org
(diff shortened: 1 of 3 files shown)


riderx:
@coderabbit review

PR #1974, opened 2026-04-27
fix(api): keep anon CLI app-list helper grants

riderx:
## Summary (AI generated)

- add a temporary migration that preserves anonymous execute on the helper functions used by the published CLI `app list` flow
- document why the legacy CLI still depends on these helper grants
- extend the SQL privilege regression test to keep the anon grants covered and to exercise an anonymous API-key `apps` read

## Motivation (AI generated)

The currently published CLI still authenticates through legacy anonymous PostgREST helper calls and then issues a direct `GET /rest/v1/apps` request with the `capgkey` header. When these helper grants drift or get tightened, CLI users regress from working API-key auth to opaque `Apps not found` / auth failures.

This restores short-term CLI reliability for existing customers without forcing an immediate CLI upgrade. It reduces support load around broken API-key workflows and buys time to ship the longer-term CLI-side migration to RBAC-aware wrappers.

## Test Plan (AI generated)

- [x] Run `bun run typecheck`
- [ ] Run `bun scripts/supabase-worktree.ts test db`
- [ ] Verify `bunx @capgo/cli app list` succeeds against a local stack after the legacy anon grants are present

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI app list now correctly returns apps for anonymous users using legacy API-key header authentication (temporary backend permission adjustment).
* **Tests**
  * Expanded test coverage for the header-based identity flow and validated app query behavior with valid vs. invalid API-key headers.

### DIFF
--- a/supabase/migrations/20260427175506_temporary_cli_apps_list_anon_helper_grants.sql
+++ b/supabase/migrations/20260427175506_temporary_cli_apps_list_anon_helper_grants.sql
@@ -2,9 +2,42 @@
 --
 -- The currently published CLI still does legacy anonymous PostgREST auth checks
 -- before issuing a direct `GET /rest/v1/apps` request with the `capgkey`
--- header. The `apps` RLS path resolves API-key identity through the helper
--- functions below, so anonymous execute on them remains required until the CLI
--- repo switches `app list` to the RBAC-aware wrappers.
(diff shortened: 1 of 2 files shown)


riderx:
@coderabbitai review

PR #2053, opened 2026-05-06
[codex] Add random sticky rollouts

riderx:
## Summary (AI generated)

- Added channel-scoped random sticky progressive rollouts backed by Cloudflare Cache decisions.
- Added rollout and auto-pause configuration to channels plus app-level rollout counters for the fast /updates path.
- Wired rollout controls through the public channel API, CLI, SDK/MCP schemas, and the web channel page.
- Added rollout-aware update selection that fetches only the selected bundle manifest.
- Added existing-cron auto-pause evaluation using rollout install/fail stats and configurable confidence thresholds.

## Motivation (AI generated)

Capgo needs gradual update releases without changing the updater, storing rollout membership per device, or adding hot-path primary database work. This lets teams expose a bundle to a percentage of devices, observe failures, and pause or rollback before a full promotion.

Progressive rollout controls reduce production blast radius for Capgo customers and make larger customers safer to onboard. Keeping apps with no active rollout on the current fast query path protects update endpoint latency and infrastructure cost.

## Screenshots (AI generated)

Rollout disabled, compact enable row:

Rollout disabled desktop

Rollout disabled mobile

Rollout enabled:

Rollout enabled desktop

## Test Plan (AI generated)

- [x] bun run supabase:db:reset
- [x] bun lint
- [x] bun lint:backend
- [x] bun typecheck
- [x] bun test:backend tests/stats.test.ts tests/rollout.test.ts tests/updates.test.ts
- [x] bun run cli:check
- [x] git diff --check
- [x] Headless Playwright screenshot capture for desktop and mobile rollout controls

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Progressive channel rollouts: percentage targets, stable vs rollout targets, enable/disable, pause/resume, promote, rollback; rollout-aware CLI/UI/API controls and debug tooling.
  * Auto-pause automation with configurable failure-rate, confidence, window, cooldown and actions.
  * Channel-scoped version data and per-channel version usage tracking.

* **Backend / Automation**
  * Database schema and migration updates, scheduled jobs, rollout decision engine, and rollout-aware request paths.

* **Tests**
  * New and expanded tests covering rollout decisions, auto-pause policy, and rollout-aware flows.

* **Documentation**
  * New localization keys for rollout and auto-pause UI.

---

> [!NOTE]
> **High Risk**
> Changes live update/release paths (channel config, stats cohorts, scheduled auto-pause/rollback) and bundle deletion guards—mistakes can affect which builds devices receive or halt rollouts incorrectly.
> 
> **Overview**
> Adds **progressive rollouts** on channels: a stable bundle stays live while a separate **rollout target** is exposed to a configurable percentage, with pause/resume, promote-to-stable, rollback, and optional **auto-pause** from install/fail stats.
> 
> **CLI & automation:** `bundle upload` can set rollout targets (`--rollout`, bps, cache TTL); `channel set` and MCP/SDK gain matching rollout and auto-pause flags. Unlink/delete flows now treat **stable and rollout** links separately (explicit Supabase FK selects, shared `bundleLinkedChannels` helpers).
> 
> **Backend:** Channel create/update APIs accept rollout and auto-pause fields; bundle delete blocks versions linked as stable or rollout. Stats/version usage record **effective channel** for install/fail (for auto-pause); Analytics Engine queries can filter by channel. New **`cron_rollout_auto_pause`** trigger evaluates policies and pause/rollback/notify.
> 
> **Dashboard:** Channel detail page rollout UI and bundle tables respect rollout linkage.
> 
> **Misc:** Schema/replicate/types for new channel columns; i18n strings; TUI workflow R2 upload uses `continue-on-error` and documents the secret key name fix; Yarn PnP path for updater version detection in `init`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00b0dc2606c7ca3daf4b24df90a72c2477d98f82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


adamsardo (commented):
I found one rollout-cohort edge case that looks worth locking down before merge.

In `supabase/functions/_backend/utils/rollout.ts`, cached selected devices are returned at lines 160-168 before the current rollout percentage is checked at lines 170-178. That means if an operator lowers a rollout from e.g. 50% to 0% while leaving `rollout_enabled` true, any device with a cached `selected: true` decision continues receiving `rollout_version` until the cache TTL expires. With the default TTL, that can keep a supposedly 0% rollout active for up to 30 days for previously selected devices.

The existing `returns stable at 0 percent` test only covers the no-cache path, and the paused/disabled tests exercise separate controls. I would either move the `percentageBps <= 0` check before `cached?.selected`, or only honor a cached selected decision when the current percentage is still positive. At minimum, please add a regression case with `cachePayload.selected = true` and `rolloutPercentageBps: 0` that expects `selected: false`.

adamsardo (commented):
I found one more authorization edge case around rollout targets.

`supabase/functions/_backend/public/channel/post.ts` still gates the whole endpoint with only `checkPermission(c, 'app.create_channel', { appId: body.app_id })`, then allows `rolloutVersion`, `rollback`, and `promoteToStable` to change `channels.rollout_version` / `channels.version` before calling `updateOrCreateChannel`, which uses the admin client. Setting `rollout_version` with `rollout_enabled=true` is effectively a partial bundle promotion because selected devices can receive that version immediately, but this path does not require `channel.promote_bundle` for the target channel.

There is a similar database-side gap: the migration adds `refresh_channel_rollout_id` as `BEFORE UPDATE OF rollout_version`, but it only refreshes rollout metadata. It does not enforce the same promotion permission that #2067 adds for `channels.version`, so direct channel updates can still move the rollout target without a `channel.promote_bundle` check.

I would treat any transition that sets, clears via rollback, or promotes from `rollout_version` as deployment control and gate it with `channel.promote_bundle` for the specific channel. Concretely: resolve the existing channel id before applying rollout fields, require `checkPermission(c, 'channel.promote_bundle', { appId: body.app_id, channelId })` when `rolloutVersion`, `rollback`, or `promoteToStable` is present, and add/extend the DB trigger so `UPDATE OF rollout_version` fails closed just like `UPDATE OF version` when the requester lacks channel promotion rights. Add a regression with a channel-level deny override where stable channel setting updates still work, but rollout target changes are rejected.

riderx:
Addressed the remaining rollout review feedback in 67b7582d0:\n\n- Cached selected rollout decisions are no longer honored when the current rollout percentage is 0%, with a regression test.\n- Rollout target changes now require channel.promote_bundle in the public channel API and direct rollout_version updates are guarded in the DB trigger, with API and direct-update regressions.\n\nCI is green and CodeRabbit review is complete.

mySebbe (commented):
﻿The latest backend/DB changes now treat `rollout_version` as a promote-bundle operation, but this screen still enables several rollout target actions from `canUpdateChannelSettings`. In particular, `rollbackRollout()` calls `saveChannelChanges({ rollout_version: null, ... })`, and `saveChannelChanges()` only switches to `canPromoteBundle` when the patch contains `version`, not when it contains `rollout_version`. A user with `channel.update_settings` but without `channel.promote_bundle` can therefore click rollback and hit the DB trigger's `NO_RIGHTS` path instead of being blocked cleanly in the UI.

Can we treat `rollout_version` changes as promote-bundle changes in `saveChannelChanges()` and gate the set-target/promote/rollback buttons with `canPromoteBundle` as well? Pause/resume and percentage/settings-only edits can still use `channel.update_settings`; the target-changing actions should match the permission enforced by `refresh_channel_rollout_id` and `public/channel/post.ts`.

KeRoY1412 (commented):
I think there is one remaining permission bypass through the delayed auto-pause rollback path.

The PR now treats direct `rollout_version` changes as `channel.promote_bundle` operations, but `auto_pause_action = 'rollback'` and its trigger settings are still writable as normal channel settings. A user with `channel.update_settings` but without `channel.promote_bundle` can configure auto-pause to rollback a rollout later, and the cron runs through `supabaseAdmin`, so the DB trigger that protects direct `rollout_version` updates will not stop that service-role update.

This is especially sharp because `auto_pause_failure_rate_bps`, `auto_pause_min_attempts`, and `auto_pause_min_failures` all allow `0`; with no attempts, `evaluateAutoPausePolicy()` computes `lowerBoundBps = 0`, so threshold `0` can trigger a rollback on the next cron pass.

I’d either gate `auto_pause_action: 'rollback'` and rollback-capable threshold changes behind `channel.promote_bundle`, or make the cron refuse rollback unless the configuration was last authorized by someone with promote rights. A regression with `channel.update_settings` allowed and `channel.promote_bundle` denied would cover this.

digzrow-coder (commented):
I think the percentage decrease path is still sticky in a way operators will not expect after the 0% fix. `resolveRolloutDecision()` now checks `percentageBps <= 0` before honoring a cached selection, but any positive decrease still returns the cached rollout decision first. For example, a device cached at `selected: true, percentage_bps: 5000` continues to receive the rollout when the channel is lowered to 1000 bps, because lines 170-178 return `cached_selected` before comparing the cached percentage with the current percentage.

That means reducing a troubled rollout from 50% to 10% does not actually shrink the exposed cohort for previously selected devices until the cache TTL expires; with the default TTL that can be 30 days. The operator-visible `rollout_percentage_bps` no longer matches runtime exposure unless they know to pause, disable, or set exactly 0.

I would treat cached selected decisions like cached rejections when the current percentage is below the cached percentage: either re-evaluate them against the reduced percentage, or invalidate/write a new unselected payload when `percentageBps < cached.percentage_bps`. A regression with `cachePayload.selected = true`, `cachePayload.percentage_bps = 5000`, and `rolloutPercentageBps = 1000` should assert that the device is not automatically kept selected just because the old cache says so.

PR #2067, opened 2026-05-07
[codex] Fix channel promotion permission bypass

riderx:
## Summary (AI generated)

- Enforce `channel.promote_bundle` with the target `channelId` in `PUT /bundle`.
- Add a database trigger to block direct `channels.version` updates when channel promotion is denied.
- Update frontend promotion and rollback affordances to use channel-scoped permission checks.
- Restore the secure header-bound `get_accessible_apps_for_apikey_v2` wrapper after a main-branch migration dropped it, keeping the current database tests and production schema aligned.
- Add regression tests for API-key promotion denial and direct database update denial.

## Motivation (AI generated)

GHSA-633x-9x77-77gf identified an authorization bypass where channel permission overrides could be skipped by checking promotion permission at app scope only. The fix makes the target channel part of the authorization decision everywhere promotion is initiated.

This protects customers using channel-specific RBAC overrides from unauthorized bundle promotion, preserving deployment controls and reducing the risk of unintended releases. The restored API-key app-listing wrapper also keeps existing CLI compatibility intact while remaining bound to the request `capgkey` header.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun lint:backend`
- [x] `bun typecheck`
- [x] `bun run supabase:with-env -- bunx vitest run tests/bundle-error-cases.test.ts tests/bundle.test.ts tests/rbac-permissions.test.ts`
- [x] `bun scripts/supabase-worktree.ts test db supabase/tests/49_test_apikey_oracle_rpc_permissions.sql`
- [x] `git diff --check`

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced per-channel permission checks for promotions, rollbacks and deployments across UI and backend.
  * Channel lists/choosers show only promotable channels; attempts on non-promotable channels are blocked with a notification.
  * Deployment banner and visibility now depend on actual deployable targets; deployment targets are filtered and operations abort if none remain.
  * Editing bundle metadata is blocked when not permitted.
  * Database guard prevents unauthorized channel version updates and surfaces permission-denied errors.

* **Tests**
  * Added RBAC tests covering channel-level permission overrides and promotion denial scenarios.


adamsardo (commented):
Found one cross-flow regression to handle before merge.

The new trigger correctly treats every `channels.version` change as `channel.promote_bundle`, but the bundle deletion path still only gates on app-scoped `bundle.delete`. In `deleteBundle()`, any linked channels are collected into `unlink`, and `unlinkChannels()` then does one bulk `channels.update({ version: unknownVersion.id }).in("id", ...)` for all of them.

With the new trigger, an app admin who has `bundle.delete` but has a channel-level deny override for one linked channel can get through the delete confirmation flow and then have the unlink update fail with `PERMISSION_DENIED_CHANNEL_PROMOTE_BUNDLE`. That leaves the bundle undeleted even though the UI already allowed the delete action. It also means one denied channel blocks unlinking every other linked channel in the bulk update.

A safer fix would be to reconcile the delete flow with the new channel gate: either require `channel.promote_bundle` on every linked channel before presenting deletion as available, or split/filter the unlink targets and show a clear blocker for channels the user cannot promote. A regression test with a user that has `bundle.delete` plus a `channel.promote_bundle` deny override on a linked channel would cover this case.

PR #2075, opened 2026-05-07
fix(api): allow scoped keys past cli warnings

riderx:
## Summary (AI generated)

- Updated `get_organization_cli_warnings` so app-scoped API keys are not blocked by an org-level read warning gate.
- Added a regression test proving an app-scoped API key can upload while receiving no org CLI warnings.

## Motivation (AI generated)

The upload command calls the org CLI warning RPC before the actual app-scoped upload permission check. New API keys can be valid for app administration or upload without org-level read access, so the warning RPC was incorrectly failing before upload could use the correct permission path.

This restores bundle uploads for customers using scoped API keys in CI/CD while preserving existing plan and upload permission checks. It reduces false authorization failures for the new API key model.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bunx vitest run tests/rbac-permissions.test.ts`
- [x] Pre-commit hook: `bun run cli:build && vue-tsc --noEmit`

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * CLI warnings refined: scoped/app API keys without org read access no longer trigger read-related fatal warnings; uploads remain permitted when appropriate.
  * Storage-limit warnings now clearly indicate when uploads will be blocked for orgs lacking storage entitlement while other services remain active.

* **Tests**
  * Added tests confirming scoped keys with restricted org read rights produce no CLI warnings and still allow uploads.


adamsardo (commented):
Found one authorization edge case worth tightening before merge.

`get_organization_cli_warnings` now returns an empty warning array whenever `check_min_rights(... read ...)` is false but the supplied API key exists and is not expired. That fixes the app-scoped upload case, but it also means any unrelated active API key can call this RPC for an org it has no read/upload relationship to and avoid the previous fatal `API key does not have read access to this organization` response.

The regression test covers a scoped key whose `limited_to_orgs` includes the target org, so it does not catch the unrelated-key path. A safer boundary would be to skip the fatal warning only when the key is non-expired and is plausibly scoped to the requested org, e.g. `orgid = ANY(api_key.limited_to_orgs)` or one of `api_key.limited_to_apps` resolves to an app owned by `orgid`; otherwise keep returning the fatal warning. That preserves app-scoped CI uploads without silently treating a wrong valid key as acceptable for this org.

grantf04:
I think this can accidentally remove the only storage-limit enforcement for app-scoped uploads. `get_organization_cli_warnings()` is where the fatal storage warning is generated when `storage` is not on a good plan, but this patch returns early for any valid scoped API key without org read access. The later `POST /bundle` path appears to check `app.upload_bundle`, HTTPS URL format, encryption policy, duplicate version, and then inserts `app_versions`; I do not see an equivalent `is_paying_and_good_plan_org_action(... storage ...)` check there. So an app-scoped key can now avoid the fatal storage warning and still create the bundle metadata for an external upload URL.

Could we either keep the storage-plan warning/check even when skipping the org-read warning, or add the same storage entitlement check to the upload/bundle creation path? A regression with storage over limit + app-scoped key + no org read would make this clear.

KCDaemon (changes requested):
Rechecked the latest head (5f70410) and the authorization/plan boundary concerns are still present in the current SQL. `get_organization_cli_warnings()` returns early whenever org `read` is denied and the supplied API key merely exists and is not expired. It does not prove that the key is scoped to the requested org, nor that one of its app scopes belongs to that org.

That keeps the unrelated-active-key path open: a valid key for some other org can avoid the fatal `API key does not have read access to this organization` warning for this org. The new regression only covers the intended app-scoped key case, so it does not protect that boundary.

The same early return also still skips the storage-plan warning path. If the later bundle creation path is meant to be the storage entitlement gate, this PR should add/point to that enforcement and cover it; otherwise app-scoped uploads without org read can bypass the fatal storage warning generated here.

`git diff --check origin/main...origin/pr-2075` passes locally, but I would keep this blocked until the early-return is limited to keys actually related to the requested org/app and the storage-limit behavior is covered by a regression or enforced later in the upload path.

RobinWitch (changes requested):
I found another blocker on the current head, separate from the existing API-key boundary comments.

This PR adds a new migration with timestamp `20260507171200`, but current `main` already has later migrations that redefine the same `public.get_organization_cli_warnings(uuid,text)` function: `20260511151503_fix_get_organization_cli_warnings_rbac.sql`, `20260516151507_fix_cli_warnings_app_scoped_apikeys.sql`, and `20260526133000_migrate_apikeys_to_v2_timestamp_fix.sql`.

Introducing this older-numbered file now makes the migration history non-append-only. On an existing environment that has already applied the later May 2026 migrations, this file is still pending and would either be rejected by the migration-order guard or run after the newer fixes, replacing the current `cli_check_permission` / app-scoped fallback implementation with the older `check_min_rights` version from this PR.

I think this needs to move into a new migration timestamped after the latest migration on `main`, and the function body should be based on the current `main` definition rather than the stale May 7 version. That preserves append-only migration order and avoids reverting the later RBAC warning fixes.

Evidence I checked:
- PR file: `supabase/migrations/20260507171200_skip_cli_warning_read_fatal_for_scoped_keys.sql`
- Current main already defines the same function in `20260511151503_fix_get_organization_cli_warnings_rbac.sql`, `20260516151507_fix_cli_warnings_app_scoped_apikeys.sql`, and `20260526133000_migrate_apikeys_to_v2_timestamp_fix.sql`.
- `supabase/schemas/prod.sql:6735-6816` shows the current function uses `public.cli_check_permission(...)` with app-scoped fallback, which is not the body introduced by this PR's older migration.

PR #2080, opened 2026-05-08
[codex] Fix RBAC API key password policy gate

riderx:
## Summary (AI generated)

- Fixed `check_min_rights` so RBAC-managed API keys can be resolved before password-policy and 2FA gates run.
- Added a pgTAP regression for an anon `apps` RLS read using an RBAC v2 API key on an org with password policy enabled.

## Motivation (AI generated)

RBAC v2 API keys have `mode = NULL`, so `get_identity_org_appid()` can return `NULL` before `rbac_check_permission_direct()` resolves the key. That made password-policy enforcement reject valid CLI/API-key requests with `CHECK_MIN_RIGHTS_PASSWORD_POLICY_ENFORCEMENT`.

This restores CLI/API-key access for organizations using RBAC v2 and password policies, avoiding false authorization failures for valid automation keys.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun scripts/supabase-worktree.ts test db supabase/tests/49_test_apikey_oracle_rpc_permissions.sql`
- [ ] Full pgTAP suite: blocked by existing PostgreSQL segfault in `supabase/tests/17_test_prevent_admin_privilege_escalation.sql` during `SELECT my_tests()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened API key authorization to enforce org-level password policies and 2FA gates so access is denied when requirements aren't met.
  * Improved handling of org/app/channel scoping for permission checks.

* **Tests**
  * Expanded RBAC v2 test coverage for API key scenarios, including compliant and stale password-policy states.

* **UI**
  * Fixed metric formatting in organization credits display.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2080)


adamsardo (commented):
I reviewed the RBAC API-key/password-policy path. The implementation direction makes sense, but I think this needs one negative regression before merge.

The new `check_min_rights` branch intentionally skips its local 2FA/password-policy block for API-key-only calls (`v_apikey IS NOT NULL AND user_id IS NULL`) and relies on `rbac_check_permission_direct` to enforce the API-key owner's password policy later. The added pgTAP case only proves the happy path where `user_password_compliance` already contains the matching policy hash.

Could you add a denial assertion for the same RBAC v2 API key after removing/staling that compliance row? For example, with the same `capgkey`, delete the `user_password_compliance` row or change `password_policy_config`, then assert the app RLS query returns `0` rows and/or `public.check_min_rights('read', NULL, org_id, app_id, NULL)` returns false. That would lock down the actual security invariant this patch is trying to preserve, not just the allowed path.

riderx:
Addressed the review by adding the negative RBAC v2 API-key/password-policy regression: after deleting the owner's user_password_compliance row, the same capgkey now asserts both the app RLS query returns 0 rows and check_min_rights returns false. CI is green on the latest commit.

PR #2083, opened 2026-05-08
fix(updates): gate strategy against installed bundle

riderx:
## Summary (AI generated)

- Compare major/minor/patch channel update strategies against the installed bundle version instead of the native baseline version.
- Add a regression test for the minor strategy case where installed bundle `1.4.0` would otherwise update to target bundle `2.4.0` when native baseline is already `2.4.0`.

## Motivation (AI generated)

The minor strategy was using `version_build` as the comparison baseline. That lets a device with installed bundle `1.4.0` and native build `2.4.0` receive target bundle `2.4.0`, even though the installed bundle crosses a major boundary. Strategy gates should use the currently installed bundle (`version_name`) because that is the version being updated from.

This prevents apps from receiving OTA updates outside the configured channel strategy, reducing unexpected cross-major rollouts for customers who use minor or patch update gates.

## Test Plan (AI generated)

- [x] `bun run lint:backend`
- [x] `bun run test:updates`
- [x] `git diff --check origin/main...HEAD`

### DIFF
--- a/cli/src/init/updater.ts
+++ b/cli/src/init/updater.ts
@@ -20,6 +20,7 @@ interface PackageJsonDependencies {
   dependencies?: Record<string, unknown>
   devDependencies?: Record<string, unknown>
   optionalDependencies?: Record<string, unknown>
+  workspaces?: unknown
 }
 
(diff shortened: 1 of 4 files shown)


riderx:
@coderabbitai review

PR #2101, opened 2026-05-10
feat(admin): add churn chart unit toggle

riderx:
## Summary (AI generated)

- Added a `$` / `%` toggle to the admin revenue churn chart.
- Kept the existing lost MRR view and added a churn-rate view using lost MRR divided by previous-day MRR.
- Extended the admin global stats trend query to return previous-day MRR for the chart calculation.

## Motivation (AI generated)

- Admins need to inspect churn as absolute lost revenue or as a normalized percentage from the same chart.

- Makes churn analysis faster by avoiding separate calculations outside the dashboard.
- Helps revenue review distinguish high-dollar churn from high-rate churn movement.

## Test Plan (AI generated)

- [x] `bunx eslint --no-ignore src/pages/admin/dashboard/revenue.vue supabase/functions/_backend/utils/pg.ts`
- [x] `bun lint`
- [x] `bun typecheck`
- [x] `git diff --check origin/main...HEAD`
- [x] `bun run build`

## Summary by CodeRabbit

* **New Features**
  * Admin revenue dashboard: toggle churn display between lost-revenue dollars ($) and churn percentage (%) with updated chart formatting.
  * Added a Plan Conversion Rate chart showing per-plan conversion trends alongside overall conversion.
  * Trend data now supports previous-period MRR to power accurate churn-rate calculations and per-plan breakdowns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2101)


SpeedyArt (commented):
I think the new plan-rate series are using the wrong denominator.

The backend now returns only `previous_mrr`, and the frontend computes each plan line as `churn_revenue_<plan> / previous_mrr`. But the numerator is plan-specific lost MRR while `previous_mrr` is total prior MRR, so the labeled `Solo Churn (%)`, `Maker Churn (%)`, etc. are not actually plan churn rates. For example, if Solo lost $10 against $100 of prior Solo MRR while total prior MRR was $1,000, the chart would show 1% instead of the plan’s 10% churn rate.

`global_stats` already has separate plan revenue columns (`revenue_solo`, `revenue_maker`, `revenue_team`, `revenue_enterprise`; those are ARR, so the per-plan MRR denominator would be prior-day revenue / 12). Could we return previous per-plan MRR denominators and use those for the plan series, or else relabel these lines as each plan’s lost MRR share of total prior MRR if that is the intended metric? A small regression with two plans that have very different prior MRR would catch the distinction.

SpeedyArt:
Thanks, the latest commit looks like it addresses the denominator issue I raised: the backend now returns per-plan `previous_mrr_*` values and the UI uses those for the plan-rate series instead of dividing each plan by total previous MRR.

PR #2102, opened 2026-05-10
[codex] Select org plan by active user count

riderx:
## Summary (AI generated)

- Added a required active-user tier choice to organization creation using plan-limit stops: 2K, 10K, 100K, and 1M+.
- Create organizations with the matching initial plan instead of always defaulting to Solo.
- Updated the Solo MAU limit to 2K in seed data and migration-backed backend data.
- Enterprise-gated features now require an active succeeded subscription, so pending onboarding plan rows do not unlock SSO.
- Refreshed the PR screenshot after the UX redesign.

## Motivation (AI generated)

New organizations should start on a plan that matches their current scale, and the choice should be clear enough that users do not have to understand plan internals.

This reduces under-provisioned org creation, makes upgrade intent visible earlier, keeps larger customers from starting in the wrong plan by default, and avoids giving Enterprise feature access from a self-reported onboarding estimate before payment approval.

## Screenshot (AI generated)

Organization plan selector

## Test Plan (AI generated)

- [x] bun lint
- [x] bun lint:backend
- [x] bun typecheck
- [x] bun run build
- [x] bun run supabase:db:reset
- [x] bun scripts/supabase-worktree.ts test db supabase/tests/00-supabase_test_helpers.sql supabase/tests/08_plan_functions.sql supabase/tests/13_test_plan_math.sql
- [x] bun run supabase:with-env -- bunx vitest run tests/organization-api.test.ts
- [x] bun run supabase:with-env -- bunx vitest run tests/sso.test.ts tests/organization-api.test.ts
- [x] Headless Playwright screenshot for the org creation selector
- [x] Headless mobile viewport overflow check

## Summary by CodeRabbit

* **New Features**
  * Organization onboarding now includes a step to estimate and select your user scale
  * Your initial plan is automatically assigned based on the estimated user count you select during onboarding
  * Starting plan selection is displayed in the onboarding summary

* **Updates**
  * Solo plan user limit increased to 2,000 monthly active users

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2102)


riderx:
@coderabbitai review

mySebbe (commented):
I think this changes a feature gate, not just onboarding plan selection. `estimatedMau` is client-controlled up to `1_000_000`, and `createPendingStripeInfo()` writes the selected plan into `stripe_info.product_id` while also marking the pending row `is_good_plan: true`. With the seeded plan table, that max value selects `Enterprise`.

Existing Enterprise-only paths, for example SSO provider creation, call `requireEnterprisePlan()`, which resolves the current plan through `get_current_plan_name_org`. That RPC only joins `orgs.customer_id -> stripe_info.product_id -> plans.name`; it does not require `stripe_info.status = 'succeeded'` or any other paid/approved subscription state. So a new authenticated user can create a trial org with a large `estimatedMau`, have this pending row identify the org as Enterprise, and pass Enterprise-gated checks before payment/approval.

I’d keep the onboarding size estimate separate from the privileged `product_id`, or make Enterprise feature gates require an active/approved Enterprise subscription instead of trusting the pending onboarding plan row.

riderx:
Addressed the review feedback in `c91c99c`:

- Enterprise gates now require an active succeeded `stripe_info` row with `is_good_plan = true`, so pending onboarding-selected plans no longer unlock SSO or other Enterprise-only feature checks.
- Added a regression test that creates a pending Enterprise onboarding org and verifies SSO provider creation returns `enterprise_plan_required`.
- Guarded plan MAU parsing in the onboarding selector so invalid plan data falls back instead of sending an invalid `estimatedMau`.

CI is green after the fix: both backend test runs and both Playwright runs passed.

PR #2105, opened 2026-05-10
fix: avoid logging signed bundle URLs

TATANKA97:
## Summary

- Stops backend logs from storing full presigned bundle/download URLs.
- Keeps the existing update response behavior unchanged while retaining non-sensitive log context.
- Adds a regression test to ensure presigned URL tokens are not written through `cloudlog`.

## Security note

This avoids exposing time-limited bundle access URLs in operational logs. The PR intentionally keeps the public description high level.

## Test plan

- [x] `.\node_modules\.bin\vitest.cmd run tests\download-url-logging.unit.test.ts`
- [x] `git diff --check`

/claim #1667

## Summary by CodeRabbit

* **Security**
  * Enhanced logging security by preventing sensitive authentication credentials from being exposed in system logs while maintaining visibility into download URL operations.

* **Tests**
  * Added unit tests to verify secure logging behavior for download URL functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2105)

### DIFF
--- a/supabase/functions/_backend/utils/downloadUrl.ts
+++ b/supabase/functions/_backend/utils/downloadUrl.ts
@@ -67,7 +67,7 @@ export async function getBundleUrl(
   if (getRuntimeKey() !== 'workerd') {
     try {
       const signedUrl = await s3.getSignedUrl(c, r2_path, EXPIRATION_SECONDS)
-      cloudlog({ requestId: c.get('requestId'), message: 'getBundleUrl', signedUrl })
+      cloudlog({ requestId: c.get('requestId'), message: 'getBundleUrl', hasDownloadUrl: Boolean(signedUrl), source: 's3' })
       const url = signedUrl
       // Since it's signed url we cannot add extra query params like checksum and device id
       // TODO: switch to our own file endpoint instead of direct s3 signed url
--- a/supabase/functions/_backend/utils/update.ts
+++ b/supabase/functions/_backend/utils/update.ts
@@ -259,7 +259,6 @@ export async function updateWithPG(
     })
   }
 
-  // cloudlog(c.get('requestId'), 'signedURL', device_id, version_name, version.name)
   if (version_name === version.name) {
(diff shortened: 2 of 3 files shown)


zinc-builds (approved):
## 🔍 Security Review — Hermes Agent

**Verdict: ✅ Approved** — Clean fix preventing credential leakage in logs.

### What the PR fixes
Signed S3 URLs were being logged verbatim in `cloudlog()` calls. These URLs contain temporary credentials (`X-Amz-Signature`, `X-Amz-Credential`) that grant access to private bundles. Logging them leaks download credentials to anyone with log access.

### ✅ Good patterns
1. **Boolean presence check**: Replacing `signedUrl` with `hasDownloadUrl: Boolean(signedUrl)` preserves observability (you can see if a URL was generated) without leaking the credential
2. **Manifest count tracking**: Added `manifestEntries: manifest.length` to the "New version available" log, compensating for the lost URL detail
3. **Test coverage**: The unit test mocks `getSignedUrl` and verifies the log output contains `hasDownloadUrl: true` and `source: 's3'` instead of the raw URL
4. **Dead code removal**: Cleaned up a commented-out `cloudlog` call

### ⚠️ Minor note
The error case on line 447 already uses `hasDownloadUrl: Boolean(signedURL)` — good consistency. The `getBundleUrl` function on line 70 now logs `source: 's3'` — consider using the same pattern across all log messages for consistency (though not blocking).

e4792200-stack (approved):
Reviewed the signed URL logging hardening. The log payloads that previously risked carrying presigned bundle URLs now keep only `hasDownloadUrl`/metadata, while `signedURL` remains used for the actual client response path. I also checked the branch for remaining `cloudlog` usages around `signedURL`/`signedUrl`; the remaining matches are not logging the raw URL. The regression test now asserts the specific `getBundleUrl` log payload rather than only the aggregate calls, and CI is green. No blockers from me.

WcaleNieWolny:

PR #2106, opened 2026-05-10
Harden website preview DNS IP validation

popka13221:
## Summary

- moved website preview DNS IP classification into a dependency-free utility
- blocks additional non-public DNS answers before server-side preview fetches
- adds coverage for multicast, reserved, documentation, 6to4, ORCHID, IPv6 documentation/multicast/link-local, and IPv4-mapped loopback forms
- parses IPv6 special ranges by hextet/mask instead of relying on string prefixes
- adds focused unit tests for the SSRF guard helper

/claim #1667

## Security impact

The `private/website_preview` endpoint validates a user-supplied website URL before fetching HTML and icons server-side. It already blocks localhost, direct IP literals, private IPv4 ranges, and some IPv6 ranges.

This patch closes gaps where a DNS response could resolve to non-public but previously unclassified ranges such as:

- IPv4 multicast/reserved: `224.0.0.0/4`, `240.0.0.0/4`, `255.255.255.255`
- 6to4 relay anycast: `192.88.99.0/24`
- IPv6 multicast, documentation, ORCHID, deprecated site-local, and link-local `fe80::/10` ranges
- IPv4-mapped IPv6 loopback forms such as `::ffff:7f00:1`

That keeps website preview behavior aligned with its existing rule: preview targets must point to public hosts.

## Residual risk

This is DNS preflight hardening. It does not fully bind the runtime `fetch()` connection to the validated DNS answer. A complete SSRF defense for DNS rebinding or split-horizon DNS would need connection-time egress enforcement, a trusted fetch proxy, or another runtime-supported way to reject private destinations at connect time.

## Tests

- `npx --yes vitest run tests/website-preview-security.unit.test.ts --config /dev/null`
- `npx --yes -p typescript tsc --noEmit --ignoreConfig --allowImportingTsExtensions --moduleResolution bundler --module esnext --target es2022 --skipLibCheck supabase/functions/_backend/utils/ip.ts`

## Notes

- No production endpoints were tested.
- No private data, credentials, or accounts were accessed.
- This is a defensive code-review patch, not an exploit submission.

## Summary by CodeRabbit

* **New Features**
  * Enhanced IP address validation to more accurately classify private and public addresses, with comprehensive support for IPv4, IPv6, and IPv4-mapped IPv6 address formats, including improved detection of special-use and reserved address ranges.
* **Tests**
  * Added comprehensive unit tests validating IP address classification across diverse address types and network ranges.
* **Refactor**
  * Improved internal code organization by consolidating IP validation logic into dedicated utilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2106)


mySebbe (commented):
﻿The new IP classifier helps for the DNS answers we inspect, but the fetch is still not bound to those validated answers. `getPublicHostnameValidationError()` resolves the hostname through Cloudflare DoH and accepts it when those A/AAAA answers are public, then `fetchValidatedUrl()` calls `fetch(currentUrl)` with the original hostname. That second request performs its own resolution at connection time, so an attacker-controlled hostname can return a public address to Cloudflare's resolver and a private/link-local address to the worker fetch resolver, or rebind between validation and fetch. The redirect loop repeats the same pre-check/fetch pattern for each hop.

For a website preview SSRF fix, the validation needs to be enforced on the address actually used for the outbound connection. Otherwise a host such as an attacker domain with TTL 0/split-horizon DNS can pass lines 172-180 and still make line 237 connect to `127.0.0.1`, `169.254.169.254`, or another internal address. Consider routing the fetch through a resolver/egress layer that rejects private targets at connect time, or otherwise binding the request to the validated IP while preserving host/SNI semantics.

SpeedyArt (commented):
I think there is still a material SSRF gap here that this hardening does not close yet.

The guard validates DNS answers by querying Cloudflare DoH in `getPublicHostnameValidationError()`, but the subsequent `fetch(currentUrl, ...)` does its own resolution/connect step. That means the allow/deny decision is not bound to the address the runtime actually connects to. A host can present public answers to the preflight resolver while the fetch path later resolves to, or is routed to, a non-public address; redirects repeat the same pattern. In that case the new `isPrivateIp()` coverage is bypassed even though the PR claims DNS-resolved non-public hosts are blocked.

For this endpoint I would avoid treating resolver preflight as the enforcement boundary. The safer fix is to make the connect path use the same vetted address, or put the website-preview fetch behind an egress control/proxy that rejects private, link-local, loopback, metadata, and reserved destinations at connection time. If that is not possible in this runtime, the PR should at least call out the residual DNS rebinding/TOCTOU risk so the bounty fix is not considered complete.

mySebbe (commented):
﻿`isPrivateIpv6()` still only rejects addresses whose text starts with `fe80:`, but IPv6 link-local is the whole `fe80::/10` block (`fe80` through `febf`). A DNS AAAA answer like `fe90::1` or `febf::1` is non-public/link-local, but it currently falls through this guard and is accepted by the website preview public-host check.

For this SSRF hardening path, please parse the first hextet and reject `(firstHextet & 0xffc0) === 0xfe80`, or use a CIDR-aware IP parser, and add regression cases for `fe90::1` and `febf::1`.

popka13221:
Thanks for the review. I pushed an update that narrows the IPv4 192.0.0.0 check to 192.0.0.0/24, adds a regression case for 192.0.1.1, and parses the first IPv6 hextet to cover the full fe80::/10 link-local range including fe90::1 and febf::1.\n\nOn the DNS rebinding / TOCTOU point: agreed. This PR is DNS preflight hardening and does not fully bind fetch() to the validated answer. I added a Residual risk section to the PR description to make that clear. A complete fix likely needs connection-time egress enforcement or a trusted fetch proxy that rejects private/link-local/reserved destinations at connect time.

KeRoY1412 (commented):
I think one more non-public IPv6 range is still missing from this denylist: deprecated site-local `fec0::/10`.

The updated parser covers `fe80::/10` link-local, ULA (`fc00::/7`), multicast, documentation, ORCHID, and 6to4, but `fec0::1` has first hextet `0xfec0`; it does not match `(firstHextet & 0xFFC0) === 0xFE80`, so `isPrivateIp('fec0::1')` returns `false`. IANA’s IPv6 address-space registry lists `fec0::/10` as reserved/deprecated, formerly site-local, so it should not be accepted as a public website-preview target.

For this SSRF guard I’d reject `(firstHextet & 0xFFC0) === 0xFEC0` as well, and add regression cases for `fec0::1` and `feff::1`.

popka13221:
Thanks, good catch. I pushed a follow-up that rejects deprecated site-local fec0::/10 via the same hextet/mask parser and added regression cases for fec0::1 and feff::1.

zinc-builds (approved):
## Security Review -- Hermes Agent

**Verdict: Approved** -- Solid SSRF prevention hardening with good refactoring.

### Improvements over the old code
1. **Regex pre-check on IPv4**: Validates format before splitting -- prevents NaN on malformed input
2. **Additional private ranges**: Added multicast (a >= 224), 6to4 relay (192.88.99.x), and 0.0.0.0/8
3. **IPv6 hextet parsing**: Uses numeric parsing instead of naive prefix-matching
4. **Code extraction**: Moving to shared utils/ip.ts enables reuse

### Suggestion (non-blocking)
The IPv6 bitmask check is correct -- consider adding a brief comment explaining it for readability.

supabase/functions/_backend/utils/ip.ts
```diff
@@ -0,0 +1,94 @@
...
+    || (Number.isFinite(firstHextet) && (firstHextet & 0xFE00) === 0xFC00)
+    || (Number.isFinite(firstHextet) && (firstHextet & 0xFF00) === 0xFF00)
+    || firstHextet === 0x2002
+    || (firstHextet === 0x2001 && secondHextet === 0x0DB8)
+    || (firstHextet === 0x2001 && secondHextet === 0x0002)
+    || (firstHextet === 0x2001 && Number.isFinite(secondHextet) && (secondHextet & 0xFFF0) === 0x0010)
+  ) {
+    return true
+  }
+  if (normalized.startsWith('::ffff:')) {
```
xordanx: `isPrivateIpv6()` still only detects IPv4-mapped addresses when the DNS answer is rendered with the compressed `::ffff:` prefix. Valid IPv4-mapped literals are not required to use that spelling, so forms like `0:0:0:0:0:ffff:7f00:1` or `0000:0000:0000:0000:0000:ffff:127.0.0.1` skip this branch and currently fall through as public.

Because this helper is the SSRF preflight for AAAA answers, the `::ffff:0:0/96` range should be detected after parsing/canonicalizing the full IPv6 address, then the low 32 bits should be passed through `isPrivateIpv4()`. I would add regression cases for both expanded hex and expanded dotted-quad forms so private IPv4 targets cannot bypass the mapped-address check by presentation format.

popka13221:
Thanks, addressed. I replaced the loose IPv6 prefix parsing with a small fail-closed hextet parser, kept valid compressed public IPv6 accepted, and added regression cases for malformed IPv6 inputs plus a public compressed IPv6 address.

e4792200-stack (changes requested):
The IP range hardening itself looks directionally useful, especially the numeric IPv6 parsing and new coverage for mapped IPv4/link-local/site-local/documentation ranges. I would not merge this revision yet because the current required `Run tests` check is red.

The failing job is the backend lint step on `supabase/functions/_backend/utils/ip.ts`: `e18e/prefer-array-fill` flags the constant callback at the zero-fill expression near the end of `parseIpv6Hextets`. Replacing `Array.from({ length: compressed.length === 2 ? missingCount : 0 }, () => "0")` with the project-preferred `.fill("0")` form should clear that failure.

I also verified the later commit now covers the earlier `fe80::/10` and `fec0::/10` gaps in the tests. Once lint is green, this will be much easier to approve.

popka13221:
Fixed the lint failure from the latest CI run () and pushed the update. Local targeted checks still pass: website preview security unit test, TypeScript check for , and .

popka13221:
Fixed the lint failure from the latest CI run (`e18e/prefer-array-fill`) and pushed the update. Local targeted checks still pass: website preview security unit test, TypeScript check for `utils/ip.ts`, and `git diff --check`.

e4792200-stack (approved):
Thanks for the follow-up. I rechecked the latest commit (`b520dae`): the lint issue I requested changes for is resolved, the IPv6 range regressions remain covered, and the PR description now clearly scopes this as DNS preflight hardening with the DNS rebinding/TOCTOU residual risk called out.

The full GitHub checks are now green as well (`Run tests`, Playwright, SonarCloud, Socket, dead-code check). No further blockers from me.

skotchdopolepet:
Review note: this improves the range classifier, but the fetch path still has a validation/use split. `getPublicHostnameValidationError()` resolves the host through Cloudflare DoH, then `fetch(currentUrl)` resolves the hostname again through the runtime resolver. A hostile hostname can answer with a public address during the validation lookup and a private/link-local address during the actual fetch, so the SSRF protection still depends on DNS stability across two independent resolutions.

A stronger follow-up would be to fetch the already-validated IP directly while preserving the original `Host` header/SNI where the runtime supports it, or proxy resolution/fetch through one controlled component that enforces the same resolved address. At minimum, I would document this residual risk so this PR is not treated as fully closing DNS rebinding for website previews.

michael-schvarcz:
Strong improvement over the original `website_preview.ts` private-IP checks — the move to `utils/ip.ts` with hextet-and-mask parsing is the right call, and the IPv6 coverage is now correct for the actual prefix lengths instead of leaning on `startsWith` strings (which silently miscount for the longer prefixes like `fec0::/10`). The PR body also honestly flags the residual rebinding risk, which I appreciate.

Few things worth tightening before this merges.

### 1. Webhook validation should now use the same library

`utils/ip.ts` is materially stricter than the `isPrivateIp` still living inside `utils/webhook.ts` — the webhook path is missing 6to4 anycast (`192.88.99.0/24`), multicast/class-E (`224.0.0.0/3`), TEST-NET-1, deprecated site-local IPv6, ORCHID, and the IPv4-mapped hex-pair form. After this PR, the website preview is well-defended and webhook delivery is not. Worse, `website_preview.ts` still imports `getWebhookUrlValidationError` from `webhook.ts` for the URL-shape check, so the two SSRF paths are now structurally entangled but inconsistent in what they block.

Suggested follow-up (could be in this PR or a quick next one): replace the inline `isPrivateIpv4`/`isPrivateIpv6`/`isPrivateIp` in `webhook.ts` with imports from `utils/ip.ts`, and have `getWebhookUrlValidationErrorAsync` use the same helper. Otherwise it's only a matter of time before a webhook target points at `192.88.99.x` or a `::ffff:0a00:0001` and someone has to re-discover this.

### 2. NAT64 `64:ff9b::/96` isn't blocked

`isPrivateIpv6` correctly covers ULA, link-local, deprecated site-local, multicast, documentation, ORCHID, benchmarking, and 6to4. But the well-known NAT64 prefix `64:ff9b::/96` (RFC 6052) isn't in the list. An attacker who can get the resolver to return a NAT64 address can embed an arbitrary IPv4 — *including any private one* — and slip through, because the embedded `10.x.x.x` is encoded in the low 32 bits of the IPv6 address and `isPrivateIpv6` never looks at them.

Concrete test that currently passes ("public") but shouldn't:

```ts

expect(isPrivateIp('64:ff9b::a00:1')).toBe(true) // currently returns false

```

Suggested fix: add `firstHextet === 0x0064 && secondHextet === 0xFF9B` to the block list, OR (better) extract the low-32-bit IPv4 from any `64:ff9b::/96` address and recurse through `isPrivateIpv4`. The recurse form also defends against the local-network-translation prefix `64:ff9b:1::/48`.

### 3. The `192.88.99.0/24` check uses `octets[2]` where the previous lines use `c`

Minor consistency — `const [a, b, c] = octets` is destructured at the top, but several lines mix `c` and `octets[2]`:

```ts

|| (a === 192 && b === 88 && octets[2] === 99)   // uses octets[2]

…

|| (a === 192 && b === 0 && c === 0)              // uses c

|| (a === 192 && b === 0 && octets[2] === 2)      // uses octets[2]

|| (a === 198 && b === 51 && octets[2] === 100)   // uses octets[2]

|| (a === 203 && b === 0 && octets[2] === 113)    // uses octets[2]

```

Easy ✋ — pick one. `c` reads slightly better in the boolean chain.

### 4. The comment on the TEST-NET block is slightly misleading

```ts

// Reserved TEST-NET ranges are also non-public for server-side fetch paths.

|| (a === 192 && b === 0 && c === 0)              // 192.0.0.0/24 — IETF Protocol Assignments, NOT a TEST-NET

|| (a === 192 && b === 0 && octets[2] === 2)      // 192.0.2.0/24 — TEST-NET-1

|| (a === 198 && b === 51 && octets[2] === 100)   // TEST-NET-2

|| (a === 203 && b === 0 && octets[2] === 113)    // TEST-NET-3

```

The first line under the comment is the IETF Protocol Assignments block (RFC 6890 § 2.2.5), not a TEST-NET. Worth either splitting it under its own comment or relabeling the section "RFC 6890 special-use ranges."

### 5. Test gaps

The IPv6 test list is thorough. A few I'd add:

- **NAT64**: `64:ff9b::a00:1` (see #2) — single highest-value addition

- **Legacy IPv4-compatible IPv6**: `::1.2.3.4` and `::a00:1` — current code returns `true` (rejected) because the mixed-mode address falls through `parseIpv6Hextets`'s `value.includes('.') → null` early-return. Pin the behavior with a test so a future cleanup of the parser doesn't accidentally unlock it.

- **Loopback in hex-pair form via `::ffff:`**: `::ffff:127.0.0.1` is covered; `::ffff:7f00:1` is covered. Worth also pinning `::ffff:c0a8:101` for 192.168.1.1 — different from anything currently exercised.

- **Globally routable IPv6 negative**: only Cloudflare and Google DNS are tested as public; worth one or two outside of the well-known DNS hosts to make sure no implementation detail accidentally classifies them as private.

### 6. Style nit: `Number.isFinite` vs `Number.isNaN`

```ts

const [high, low] = parts.map(part => Number.parseInt(part, 16))

if (!Number.isFinite(high) || !Number.isFinite(low) || high < 0 || high > 0xFFFF …)

```

`parseInt('xyz', 16)` returns `NaN`, and `Number.isFinite(NaN)` is `false`, so the check works. But `Number.isNaN(x)` reads more directly for the case you're actually trying to catch. The `high < 0` guard is also unreachable since `parseInt` with a positive radix can only return `NaN` or a non-negative integer. Pure style, take it or leave it.

---

Excellent direction overall — turning this into a shared util is exactly what should have happened the first time, and now you've done it for two endpoints. Closing the webhook side too (#1) is the highest-leverage follow-up.

digzrow-coder:
The IP range hardening is good, but `fetchPublicUrl()` still has the same DNS TOCTOU window as the webhook path. `getPublicHostnameValidationError()` resolves A/AAAA through Cloudflare DoH, then `fetch(currentUrl, { redirect: 'manual' })` performs a separate runtime DNS resolution for the actual connection. A rebinding hostname can answer the validation lookup with a public IP and then answer the fetch resolver with `127.0.0.1`, RFC1918, or metadata/link-local addresses.

Because redirects are already handled manually and each `Location` is revalidated, the missing piece is binding validation to the connection that gets opened. I would add a regression where the validation resolver sees a public address but the delivery/fetch resolver reaches a private address, and either route these requests through an egress proxy that enforces peer IPs or use a runtime mechanism that connects only to the validated address while preserving the original Host/SNI semantics.

PR #2111, opened 2026-05-10
Fix update statistics card overflow

popka13221:
## Summary

- fixes the update statistics card header overflow by keeping the header focused on the title/total
- moves the requested/installed/failed legend into the chart content area
- renders compact legend numbers while preserving full values in title/aria labels
- keeps the chart area constrained with `min-h-0`/flex sizing so the legend does not push content outside the card

Fixes #1790

## Tests

- `git diff --check`
- targeted template sanity check with Node

## Notes

- Local full app build was not run because this worktree does not have the repository package manager/runtime installed (`bun` is unavailable locally).
- No production endpoints were accessed.

### DIFF
--- a/src/components/dashboard/UpdateStatsCard.vue
+++ b/src/components/dashboard/UpdateStatsCard.vue
@@ -122,6 +122,15 @@ const hasData = computed(() => effectiveTotalUpdates.value > 0 || isDemoMode.val
 
 const PAGE_SIZE = 1000
 
+const compactNumberFormatter = new Intl.NumberFormat(undefined, {
+  maximumFractionDigits: 1,
+  notation: 'compact',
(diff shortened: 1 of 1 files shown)


e4792200-stack (approved):
Reviewed the single-component diff for the stats-card overflow fix. The change keeps the `ChartCard` total/evolution rendering intact, moves only the per-action legend into the chart content area, and the `min-h-0`/`flex-1` wrapper preserves the chart area sizing rather than letting the legend force the card taller. The compact legend keeps full values available through `aria-label`/`title`, so the abbreviated display does not drop precision for assistive/hover use.

I also checked `git diff --check` locally and the PR checks are green. No blockers from my side.

PR #2113, opened 2026-05-10
fix: restrict admin credit grants to internal API

jonit-dev:
## Summary

Restricts the admin credit grant endpoint to internal operational callers and removes the browser-facing grant form from the platform admin credits dashboard.

- Move `POST /private/admin_credits/grant` behind `middlewareAPISecret`
- Keep platform-admin credit dashboard behavior read-only: search, balances, analytics, and history
- Remove grant-form state/actions from the Vue admin page
- Add a regression test proving JWT-authenticated platform-admin requests do not reach credit grants

## Security rationale

`AGENTS.md` says platform-admin capabilities must remain read-only except impersonation, and must not mutate billing state. Credit grants mutate billing/accounting state through `top_up_usage_credits`, so this path should be internal API-secret-only rather than reachable through platform-admin JWT/API-key middleware.

Related to the rewarded security review thread: #1667

## Test plan

- `git diff --check`
- `/home/joao/.bun/bin/bun test tests/admin-credits-auth-boundary.unit.test.ts`
- `/home/joao/.bun/bin/bun lint`

Note: I did not run the full `tests/admin-credits.test.ts` integration suite locally because the Supabase stack was not running on `127.0.0.1:54321` in this environment.

## Summary by CodeRabbit

* **New Features**
  * Added Organization Credit Lookup feature allowing admins to search and view organization credit balances directly.

* **Changes**
  * Removed the admin credit grant form interface from the dashboard.
  * Updated credits section UI labels, descriptions, and related helper text.
  * Credit grants are now processed through internal mechanisms only.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2113)


e4792200-stack (changes requested):
Thanks for tightening this boundary. I verified the route/UI direction and the CI is green, but I found one blocker before merge.

New grants created by `POST /private/admin_credits/grant` now write `source_ref` with `{ granted_via: "internal_api", org_name }`, but `/grants-history` still fetches only manual grants where `source_ref->admin_user_id` is not null. Because this PR removes `admin_user_id` from the new `source_ref`, internal grants created after this change will disappear from the read-only grants history dashboard even though the page still advertises grant history.

Please update the history query to include the new `granted_via = internal_api` shape while preserving legacy `admin_user_id` rows. A happy-path test using the internal `apisecret` header would also cover the only allowed caller and catch this kind of regression. I also checked the patch with `git show --check`; no whitespace issues.

digzrow-coder (commented):
The grants-history endpoint still filters on the old UI shape, so the new internal grants disappear from the read-only admin page. `/grant` now writes `source_ref = { granted_via: 'internal_api', org_name: ... }` with no `admin_user_id`, but `/grants-history` keeps `.not('source_ref->admin_user_id', 'is', null)`. After this change, any credits granted through the only remaining mutation path are excluded from the "Recent Admin Grants" table, so admins can no longer audit the grant history from this page. The filter needs to include the new internal grant shape, or be changed to match the intended manual-credit sources.

KCDaemon (changes requested):
Rechecked latest head (f7503e1). Moving the grant endpoint behind the internal API secret closes the user-JWT mutation path, but the successful grant is no longer attributable to any actor.

Before this PR, the grant wrote `admin_user_id` into `source_ref` and logged `adminUserId`. Now every grant records only `{ granted_via: "internal_api", org_name }`, and the success/failure logs also omit who or which internal job/operator initiated the credit mutation. For a financial credit grant, a shared secret should authenticate the caller, but it should not be the only audit identity.

Please require an internal actor/source identifier (for example service name, job id, or operator id) and persist it in `source_ref` and logs, then add a regression asserting grant history retains that attribution for internal grants. Without that, admins can see balances changed but cannot audit who/what granted the credits.

radiantjade:
I think this leaves the read-only history path out of sync with the new grant writer.

`POST /grant` now records manual grants with `source_ref = { granted_via: 'internal_api', org_name: ... }`, but `GET /grants-history` still filters with `.not('source_ref->admin_user_id', 'is', null)`. Any grant created through the new internal-secret path therefore has `source = 'manual'` but no `admin_user_id`, so it is excluded from the history endpoint and the admin Credits page never shows the new grants it is supposed to review.

A quick repro from the code path is: call `/private/admin_credits/grant` with a valid `apisecret`; the inserted grant uses the new `internal_api` source_ref; then `/private/admin_credits/grants-history` filters it out. I would update the history query to include the new internal marker, for example `source = manual` plus either the old `admin_user_id` marker or `source_ref->>granted_via = internal_api`, and add a positive regression with the API secret grant followed by `grants-history` returning that row. That would also restore coverage for the successful grant path, which this PR currently removes from the test suite.

PR #2116, opened 2026-05-11
fix: disable Drizzle query logging by default

ocepkejepintu-droid:
## Summary

- Disable Drizzle SQL query logging by default for shared backend DB clients.
- Keep query logging available as an explicit opt-in argument for focused debugging.

## Motivation

The shared `getDrizzleClient` helper currently enables Drizzle logging for every backend caller. That can put SQL text and parameters into application logs across normal request paths. Query logging is useful for local debugging, but it should not be the default for production-facing shared clients.

## Business Impact

This reduces accidental sensitive-data exposure in logs with a very small runtime change. Existing callers keep using the same helper and can still opt into query logging when needed.

## Test Plan

- [x] `bunx eslint supabase/functions/_backend/utils/pg.ts`
- [x] `git diff --check`
- [x] GitHub checks: Run tests, Run Playwright tests, Socket Security, SonarCloud, CodSpeed

## Checklist

- [x] Kept the change scoped to the shared Drizzle client helper.
- [x] Preserved opt-in query logging for focused debugging.
- [x] Confirmed no migration or documentation update is required for this behavior change.

## Summary by CodeRabbit

* **Chores**
  * Updated database logging configuration for backend operations, reducing verbosity by default while maintaining the ability to enable detailed logging when needed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2116)

### DIFF
--- a/supabase/functions/_backend/utils/pg.ts
+++ b/supabase/functions/_backend/utils/pg.ts
@@ -373,8 +373,8 @@ export function getPgClient(c: Context, readOnly = false) {
   return pool
 }
 
-export function getDrizzleClient(db: ReturnType<typeof getPgClient>) {
-  return drizzle({ client: db, logger: true })
+export function getDrizzleClient(db: ReturnType<typeof getPgClient>, logger = false) {
+  return drizzle({ client: db, logger })
 }
 
 // Helper to extract detailed error information from pg errors


e4792200-stack (approved):
Reviewed the hardening change. The old helper always instantiated Drizzle with `logger: true`, and this PR flips only the shared default while leaving an explicit opt-in parameter for focused debugging. I checked the current `getDrizzleClient(...)` call sites and they use the default helper form, so this should reduce query/parameter log exposure broadly without forcing behavior changes at call sites. CI is green; no blockers from me.

WcaleNieWolny:

PR #2117, opened 2026-05-11
[codex] Fix admin global stats date joins

ToriChanIntegration:
## Summary (AI generated)

- Guard the admin global stats previous-day join so `date_id` is only cast when it matches `YYYY-MM-DD`.
- Filter malformed `global_stats.date_id` rows out of the trend query instead of letting one bad row empty the chart response.

## Motivation (AI generated)

Admin charts started returning `success: true` with an empty `global_stats_trend` dataset after the recent release added a `date_id::date` self-join. Because `date_id` is stored as text, any non-date historic row can make the query throw, and the helper catches that error by returning an empty array.

Restores admin dashboard trend charts used to monitor platform health, revenue, usage, and conversion metrics without requiring a production data cleanup first.

## Test Plan (AI generated)

- [x] `bun test tests/admin-stats.unit.test.ts`
- [x] `bun lint`
- [x] `bun lint:backend`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed global statistics trending calculations to ensure only properly formatted dates are processed, improving data accuracy and preventing calculation errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2117)

### DIFF
--- a/supabase/functions/_backend/utils/pg.ts
+++ b/supabase/functions/_backend/utils/pg.ts
@@ -1270,8 +1270,15 @@ export async function getAdminGlobalStatsTrend(
         COALESCE(NULLIF(to_jsonb(gs) ->> 'builder_active_paying_clients_60d', '')::int, 0)::int AS builder_active_paying_clients_60d,
         COALESCE(NULLIF(to_jsonb(gs) ->> 'live_updates_active_paying_clients_60d', '')::int, 0)::int AS live_updates_active_paying_clients_60d
       FROM global_stats gs
-      LEFT JOIN global_stats prev ON prev.date_id = (gs.date_id::date - 1)::text
-      WHERE gs.date_id <= ${endDateOnly}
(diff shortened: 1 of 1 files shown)


jihadMo (approved):
Reviewed the current patch after the date validation update. The important part for me is that the previous unsafe `gs.date_id::date` cast path is gone: the join now computes the previous date only after the `YYYY-MM-DD` shape check and the `to_date`/`to_char` round-trip confirms it is a real calendar date. The `WHERE` clause applies the same guard before the later trend filtering, so malformed `date_id` rows should be skipped instead of crashing the admin stats query.

The remaining `gs.date_id <= ${endDateOnly}` comparison is still a text comparison, but after the round-trip validation every surviving value is normalized `YYYY-MM-DD`, so lexical order matches chronological order. CI is green on the PR. No blockers from me.

PR #2120, opened 2026-05-11
fix: avoid logging full client config fallback

ocepkejepintu-droid:
## Summary

- Stop printing the full local client config object when remote config fetch falls back.
- Keep fallback behavior unchanged and retain enough status context for debugging.

## Motivation

The fallback path currently logs the whole local config object in the browser console. That object includes deployment configuration such as the Supabase URL and anon key. The anon key is intended for client use, but printing the full config is unnecessary and makes logs noisier than needed.

## Test Plan

- [x] `bunx eslint src/services/supabase.ts`
- [x] `git diff --check`

## Checklist

- [x] Behavior unchanged: local config is still used on failed remote config fetch.
- [x] Fallback logs no longer include the full config object.
- [x] Change kept scoped to client config fallback logging.

/claim #1667

### DIFF
--- a/src/services/supabase.ts
+++ b/src/services/supabase.ts
@@ -57,7 +57,7 @@ export async function getRemoteConfig() {
   try {
     const response = await fetch(`${defaultApiHost}/private/config`)
     if (!response.ok) {
-      console.log('Local config', localConfig)
+      console.log('Using local config fallback', { status: response.status })
       stripeEnabled.value = localConfig.stripeEnabled ?? stripeEnabled.value
       return localConfig as CapgoConfig
     }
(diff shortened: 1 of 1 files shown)


jihadMo (approved):
Reviewed this scoped hardening patch. Both fallback paths that previously printed the full `localConfig` object now log only fallback status/context, while the behavior stays the same: `stripeEnabled` is still updated from the local config and the same local config is returned when remote config fetch fails or returns a non-OK response.

This keeps the useful signal that the fallback path was used without putting the Supabase URL/anon key and any future config fields into routine browser logs. Checks on the PR are green. No blockers from me.

WcaleNieWolny:

PR #2121, opened 2026-05-11
fix: redact private endpoint log payloads

ocepkejepintu-droid:
## Summary
- stop logging full request bodies for private Stripe portal, upload-link, and invite endpoints
- keep metadata-only diagnostics for request tracing
- avoid recording Stripe customer rows in routine logs

## Motivation
This is a small public hardening follow-up for #1667. These endpoints are authenticated/private and the existing logs included request payloads or customer row data that are not needed for normal debugging.

## Business Impact
- Lowers accidental sensitive-data exposure in logs
- Keeps behavior and API responses unchanged
- Preserves enough log signal for request tracing

## Test Plan
- `bunx eslint supabase/functions/_backend/private/stripe_portal.ts supabase/functions/_backend/private/upload_link.ts supabase/functions/_backend/private/invite_new_user_to_org.ts`
- `git diff --check`

## Checklist
- [x] No behavioral/API contract changes
- [x] Focused security hardening
- [x] Local lint passes

### DIFF
--- a/supabase/functions/_backend/private/invite_new_user_to_org.ts
+++ b/supabase/functions/_backend/private/invite_new_user_to_org.ts
@@ -85,7 +85,13 @@ async function validateInvite(c: Context, rawBody: any) {
   }
 
   const body = validationResult.data
-  cloudlog({ requestId: c.get('requestId'), context: 'invite_new_user_to_org validated body', body })
+  cloudlog({
+    requestId: c.get('requestId'),
+    context: 'invite_new_user_to_org validated body',
+    orgId: body.org_id,
+    inviteType: body.invite_type,
+    hasCaptchaToken: Boolean(body.captcha_token),
+  })
 
   const authorization = c.get('authorization')
   if (!authorization)
--- a/supabase/functions/_backend/private/invite_new_user_to_org.ts
+++ b/supabase/functions/_backend/private/invite_new_user_to_org.ts
@@ -161,7 +161,7 @@ async function validateInvite(c: Context, rawBody: any) {
 
 app.post('/', middlewareAuth, async (c) => {
   const rawBody = await parseBody<any>(c)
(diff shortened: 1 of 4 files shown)


janishahn:
This patch still leaves the same private-endpoint log exposure in nearby routes:

- `supabase/functions/_backend/private/invite_new_user_to_org.ts` still logs the full validated invite body in `validateInvite`, including invitee email/name/role and `captcha_token` when present.
- `supabase/functions/_backend/private/stripe_checkout.ts` still logs the full checkout request body and later logs the loaded org object containing the Stripe `customer_id`.

If this PR is meant to close private endpoint payload logging, those should be redacted in the same pass. Otherwise invite/captcha data and Stripe customer identifiers can still land in routine worker logs after this merges.

ocepkejepintu-droid:
Addressed in 6772337ff. I extended the same metadata-only logging approach to the nearby invite and checkout routes:

- invite validation now logs org id, invite type, and captcha-token presence only, not email/name/role/captcha body values
- stripe checkout now logs org id, recurrence, and optional-field presence only, not the full checkout body
- org lookup logging now records only org id and whether a customer id exists, not the Stripe customer id value

Re-ran:

`bunx eslint supabase/functions/_backend/private/invite_new_user_to_org.ts supabase/functions/_backend/private/stripe_checkout.ts supabase/functions/_backend/private/stripe_portal.ts supabase/functions/_backend/private/upload_link.ts`
`git diff --check`

jihadMo (approved):
Reviewed the current version after the follow-up commit. The earlier gap in nearby private endpoints looks addressed now: invite validation logs only org/invite metadata plus captcha-token presence, Stripe checkout logs org/recurrence/optional-field presence instead of the full request body, and the org lookup log no longer includes the Stripe customer id value.

The original portal/upload-link changes also preserve behavior while dropping raw request payloads. The PR checks are now green, including the main test job and Playwright. No blockers from me.

WcaleNieWolny:

PR #2124, opened 2026-05-11
Block deleted bundle reads from cache

slashdevcorpse:
## Summary

- Block app-scoped attachment reads when the requested path matches a soft-deleted `app_versions.r2_path` or a persisted manifest asset path.
- Run the deleted-bundle check before cache reads and cache-to-R2 restore logic.
- Keep private download links from resolving deleted bundles.
- Add indexes for deleted bundle path lookups and persisted manifest path lookups.
- Add regression coverage for deleted bundle paths, persisted manifest asset paths, and non-deleted cached bundle reads.

/claim #1667

## Test plan

- [x] GitHub CI passed on head `0374e6c`.
- [x] `bunx vitest run tests/files-app-read-guard.unit.test.ts`
- [x] `bunx vitest run tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/download-link-deleted-bundle.unit.test.ts`
- [x] `bunx eslint supabase/functions/_backend/files/files.ts supabase/functions/_backend/private/download_link.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/download-link-deleted-bundle.unit.test.ts`
- [x] `git diff --check`

## Screenshots

Not applicable; backend-only change.

## Checklist

- [x] My code follows the code style of this project and passes the targeted lint checks listed above.
- [x] I have tested my code manually, and I have provided steps how to reproduce my tests.

### DIFF
--- a/supabase/functions/_backend/files/files.ts
+++ b/supabase/functions/_backend/files/files.ts
@@ -363,8 +363,15 @@ async function assertReadableAppScopedAttachment(c: Context, fileId: unknown): P
         FROM public.app_versions
         WHERE owner_org = $1
           AND app_id = $2
-          AND r2_path = $3
-          AND COALESCE(deleted, false) = true
+          AND deleted = true
+          AND (
+            r2_path = $3
+            OR EXISTS (
+              SELECT 1
+              FROM unnest(manifest) AS manifest_entry
+              WHERE manifest_entry.s3_path = $3
+            )
+          )
         LIMIT 1
       `,
       [scopedPath.owner_org, scopedPath.app_id, fileId],
--- a/supabase/functions/_backend/files/files.ts
+++ b/supabase/functions/_backend/files/files.ts
@@ -356,6 +356,23 @@ async function assertReadableAppScopedAttachment(c: Context, fileId: unknown): P
     if (!app || app.owner_org !== scopedPath.owner_org) {
       quickError(404, 'not_found', 'Not found')
     }
+
+    const deletedBundlePath = await pgClient.query<{ id: number }>(
+      `
+        SELECT id
+        FROM public.app_versions
+        WHERE owner_org = $1
+          AND app_id = $2
+          AND r2_path = $3
+          AND COALESCE(deleted, false) = true
+        LIMIT 1
+      `,
+      [scopedPath.owner_org, scopedPath.app_id, fileId],
+    )
+
+    if (deletedBundlePath.rows.length > 0) {
+      quickError(404, 'not_found', 'Not found')
+    }
   }
   finally {
     await closeClient(c, pgClient)
--- a/supabase/functions/_backend/files/files.ts
+++ b/supabase/functions/_backend/files/files.ts
@@ -368,8 +368,9 @@ async function assertReadableAppScopedAttachment(c: Context, fileId: unknown): P
             r2_path = $3
             OR EXISTS (
               SELECT 1
-              FROM unnest(manifest) AS manifest_entry
-              WHERE manifest_entry.s3_path = $3
+              FROM public.manifest m
+              WHERE m.app_version_id = public.app_versions.id
+                AND m.s3_path = $3
             )
           )
         LIMIT 1
--- a/supabase/functions/_backend/private/download_link.ts
+++ b/supabase/functions/_backend/private/download_link.ts
@@ -44,6 +44,7 @@ app.post('/', middlewareAuth, async (c) => {
     .select('*, owner_org ( created_by )')
     .eq('app_id', body.app_id)
     .eq('id', body.id)
+    .eq('deleted', false)
     .single()
 
   const ownerOrg = bundle?.owner_org.created_by
--- a/supabase/migrations/20260511012637_add_app_versions_deleted_r2_path_index.sql
+++ b/supabase/migrations/20260511012637_add_app_versions_deleted_r2_path_index.sql
@@ -1,3 +1,6 @@
 CREATE INDEX IF NOT EXISTS "idx_app_versions_deleted_r2_path"
 ON "public"."app_versions" USING "btree" ("owner_org", "app_id", "r2_path")
 WHERE ("deleted" = true);
+
+CREATE INDEX IF NOT EXISTS "idx_manifest_app_version_id_s3_path"
+ON "public"."manifest" USING "btree" ("app_version_id", "s3_path");
--- a/supabase/migrations/20260511012637_add_app_versions_deleted_r2_path_index.sql
+++ b/supabase/migrations/20260511012637_add_app_versions_deleted_r2_path_index.sql
@@ -1,0 +1,3 @@
+CREATE INDEX IF NOT EXISTS "idx_app_versions_deleted_r2_path"
+ON "public"."app_versions" USING "btree" ("owner_org", "app_id", "r2_path")
+WHERE ("deleted" = true);
--- a/tests/download-link-deleted-bundle.unit.test.ts
+++ b/tests/download-link-deleted-bundle.unit.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, it, vi } from 'vitest'
+import { createAllCatch, createHono } from '../supabase/functions/_backend/utils/hono.ts'
+import { version } from '../supabase/functions/_backend/utils/version.ts'
+
+const eqMock = vi.fn()
+const singleMock = vi.fn()
+const getBundleUrlMock = vi.fn()
+
+vi.mock('../supabase/functions/_backend/utils/discord.ts', () => ({
+  sendDiscordAlert500: () => Promise.resolve(),
+  sendDiscordAlert: () => Promise.resolve(),
+}))
+
+vi.mock('../supabase/functions/_backend/utils/downloadUrl.ts', () => ({
+  getBundleUrl: getBundleUrlMock,
+  getManifestUrl: vi.fn(() => []),
+}))
+
+vi.mock('../supabase/functions/_backend/utils/rbac.ts', () => ({
+  checkPermission: vi.fn(() => Promise.resolve(true)),
+}))
+
+vi.mock('../supabase/functions/_backend/utils/supabase.ts', () => ({
+  supabaseClient: vi.fn(() => ({
+    from: vi.fn(() => ({
+      select: vi.fn(() => ({
+        eq: eqMock,
+      })),
+    })),
+  })),
+}))
+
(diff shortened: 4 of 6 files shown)


lyndon050516 (commented):
I would not merge this without a matching migration for the new `app_versions` lookup.

I checked the current schema and migrations: `app_versions` has separate indexes for `owner_org`, `app_id`, and `deleted`, plus `deleted = false` partial indexes for other paths, but I do not see any index involving `r2_path` or a `deleted = true` partial index. Since this guard runs in `assertReadableAppScopedAttachment` before serving cached content, every app-scoped attachment read now pays for this query before it can use the cache.

A narrow partial index would keep the security guard cheap and scoped to the rows it cares about, for example:

```sql
CREATE INDEX IF NOT EXISTS idx_app_versions_deleted_r2_path
  ON public.app_versions (owner_org, app_id, r2_path)
  WHERE deleted = true;
```

If the migration runner supports concurrent index creation, `CONCURRENTLY` would also be worth considering. The key point is covering `(owner_org, app_id, r2_path)` for deleted rows so cache hits do not become repeated table scans as `app_versions` grows.

slashdevcorpse:
Addressed the index and coverage feedback.

Current head `cec23dc` now:
- checks deleted `r2_path` and stored manifest asset paths before cache reads or cache-to-R2 restore
- filters deleted bundles in `/private/download_link`
- adds the partial deleted-path index plus regression coverage for deleted zip paths, deleted manifest asset paths, non-deleted cached reads, and private download links

CI is green on the current head.

SpeedyArt (commented):
I think the manifest-asset branch of this guard can miss the normal persisted state.

`on_version_update` copies `record.manifest` into rows in `public.manifest`, updates `manifest_count`, and then clears `app_versions.manifest` with `.update({ manifest: null })`. Runtime manifest download paths are also built from `public.manifest` (`download_link.ts` -> `.from('manifest')`, plus the update queries join `schema.manifest`). So after the trigger has run, the deleted version's asset paths usually live in `public.manifest`, not in the `app_versions.manifest` array this query unnests.

That means a cached direct read for a deleted manifest asset can still get `rows.length === 0` here and be served from cache, even though the PR now blocks `r2_path` bundle reads and private download-link lookup. I would change the `EXISTS` branch to check the persisted table, for example `EXISTS (SELECT 1 FROM public.manifest m WHERE m.app_version_id = app_versions.id AND m.s3_path = $3)`, and add a regression where `app_versions.manifest` is null but `public.manifest` contains the deleted asset path. Since this now runs before cache hits, it probably also wants an index covering the manifest lookup such as `(app_version_id, s3_path)` or `(s3_path, app_version_id)`.

slashdevcorpse:
Pushed `0374e6c` to address the read-guard review feedback.

Changes:
- The deleted-asset guard now checks persisted `public.manifest` rows instead of the cleared `app_versions.manifest` array.
- Added `idx_manifest_app_version_id_s3_path` alongside the deleted `app_versions.r2_path` partial index.
- Updated the regression to cover persisted manifest asset paths and keep the positive cached-read case.

Validation:
- `bunx vitest run tests/files-app-read-guard.unit.test.ts`
- `git diff --check`

mingisrookie (approved):
Reviewed current head `0374e6c` with focus on the deleted-bundle cache-read guard and the two earlier review concerns. The files worker now checks the primary DB before any cached/R2 response is served, and the guard covers both `app_versions.r2_path` and persisted `public.manifest.s3_path` rows for deleted versions. The private `/download_link` path also filters `deleted = false`, so it no longer mints bundle/manifest links for soft-deleted rows.

I also checked the performance feedback path: the migration adds the deleted `app_versions(owner_org, app_id, r2_path)` partial index and the `manifest(app_version_id, s3_path)` index used by the `EXISTS` branch.

- `bun x eslint supabase/functions/_backend/files/files.ts supabase/functions/_backend/private/download_link.ts tests/download-link-deleted-bundle.unit.test.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts`
- `bun x vitest run tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/download-link-deleted-bundle.unit.test.ts` → 3 files / 8 tests passed
- `git diff --check 3d184f8..HEAD`

No remaining touched-file blocker from me.

digzrow-coder:
The manifest-file cache guard still has a hole after the normal deleted-version cleanup runs. `assertReadableAppScopedAttachment()` only blocks a delta object when it can join the requested `s3_path` back through `public.manifest` to a deleted `app_versions` row. But the existing delete path calls `deleteManifest()` when a version is soft-deleted, and that deletes the `manifest` rows before/while deleting the R2 objects. If a delta object was already in the edge cache, the later `/files/read/attachments/orgs/.../delta/...` request no longer has any `manifest` row to match, so the new query returns empty and the handler can still serve the stale cached object.

The new tests cover the direct bundle `r2_path` case, but not this manifest-row-removed case. To actually close deleted partial reads, the guard needs a durable tombstone/path index that survives manifest row cleanup, or the delete flow needs to preserve enough path metadata until caches are guaranteed cold. A regression should seed/cache a manifest attachment path, soft-delete the version so its manifest row is removed, then assert the cached `/delta/...` read is still blocked.

KCDaemon (changes requested):
Rechecked latest head (0374e6c). The new cache guard still depends on a live public.manifest row to recognize deleted manifest assets:

```sql
EXISTS (SELECT 1 FROM public.manifest m WHERE m.app_version_id = public.app_versions.id AND m.s3_path = $3)
```

But the normal delete flow removes manifest rows during deleteManifest()/version cleanup before or while R2 objects are deleted. Once that row is gone, a cached delta asset request no longer matches the guard, so `/files/read/attachments/.../delta/...` can still serve stale cached bytes. The added regression only covers the persisted-manifest-row case, not the row-removed-after-soft-delete case.

To close this, persist a durable tombstone/path index that survives manifest cleanup, or preserve enough path metadata until cache/R2 deletion is guaranteed. A regression should seed/cache a manifest asset, run the delete flow so the manifest row is removed, then assert the cached attachment read remains blocked. The PR is currently merge-conflicted (DIRTY); keep blocked until this path is covered.

austin33133-maker:
The DB-side guard work here looks solid -- the manifest persistence concern @digzrow-coder and @KCDaemon raised is the biggest miss to address before merging. Two additional gaps worth considering since they sit *upstream* of this guard and aren't covered by the new tests.

**1. Edge / CDN caching can serve deleted bundle bytes despite the DB guard**

`assertReadableAppScopedAttachment` runs inside the files worker, which is great for first-fetch paths. But two upstream layers can return a cached copy without invoking the worker:

- **Cloudflare R2 + the worker's own `cache.match()` paths.** If the same `fileId` was successfully fetched before deletion, Workers Cache (and any CDN sitting in front) may have a 200 response keyed on the URL. `assertReadableAppScopedAttachment` only protects the slow path. A client hitting the cached path won't run this check at all.
- **Pre-existing browser caches**, especially when the underlying R2 object was served with a long `Cache-Control: public, max-age=...` and no per-org ETag/varying header. The browser revalidates against the URL, not against the DB.

For SaaS billing/customer-data bundles this is the exact "deleted but still accessible" scenario the PR is trying to close. Suggested follow-ups:

- After `app_versions.deleted = true`, enqueue `caches.default.delete(<canonical url>)` for every variant URL that targets the deleted `r2_path` and persisted `manifest.s3_path`.
- Add a `Cache-Control: private, must-revalidate` (or short max-age) to app-scoped attachment responses so the browser doesn't keep a long-lived copy past deletion.
- Add a regression test that primes the worker cache for a path, soft-deletes, and asserts the next request returns 404 -- not the cached 200.

**2. Signed download links already issued continue to resolve after deletion**

`download_link.ts` now adds `.eq('deleted', false)` so **new** signed URLs can't be minted for a deleted bundle. But signed URLs minted *before* deletion keep working for the remainder of their TTL because the resolver they point to doesn't recheck `app_versions.deleted`.

If TTLs are short (minutes) the exposure window is small; if TTLs are hour/day-scale (common for big bundle downloads) the deletion is essentially advisory until expiry.

Two options, low to high effort:

- Cheap: keep the new check, but also wire the soft-delete write to bump a per-app/per-version `signing_epoch` column and include it in the signed payload, then have the resolver reject if `epoch < app_version.signing_epoch`.
- Cheaper still: shorten TTL on `download_link` issuance and document the window.

A test that issues a signed URL, soft-deletes the version, and asserts the URL returns 404 would make either fix verifiable.

---

Neither blocks the *direct-fetch* fix that's the meat of this PR, but both undercut the security claim if left out. Happy to send a follow-up PR with the cache-invalidate path + a regression test if there's interest.

digzrow-coder:
I could not push to the original `slashdevcorpse/capgo` PR branch from this GitHub account, so I rebased the fix onto current `main` and pushed it here:

- Branch: https://github.com/digzrow-coder/capgo/tree/codex/capgo-deleted-bundle-read-guard
- Fork PR with exact diff: https://github.com/digzrow-coder/capgo/pull/1
- Head commit: `f64ebc3b7` (`Persist deleted manifest path tombstones`)

This addresses the remaining blocker from @KCDaemon: the guard no longer depends only on a live `public.manifest` row. Before `deleteManifest()` removes manifest rows, it writes durable `deleted_app_version_manifest_paths` tombstones keyed by `(app_version_id, s3_path)`, and the attachment read guard checks those tombstones before cache/R2 reads. That keeps cached manifest/delta asset requests blocked after the normal manifest cleanup path has removed the original rows.

Validation on the rebased branch:
- `bun x vitest run tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/download-link-deleted-bundle.unit.test.ts tests/on-version-update-cleanup.unit.test.ts` -> 4 files / 13 tests passed
- `bun x eslint supabase/functions/_backend/files/files.ts supabase/functions/_backend/triggers/on_version_update.ts supabase/functions/_backend/private/download_link.ts tests/download-link-deleted-bundle.unit.test.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/on-version-update-cleanup.unit.test.ts`
- `git diff --check`

I also tried opening a replacement PR from `digzrow-coder/capgo` to `Cap-go/capgo`, but GitHub returned `digzrow-coder does not have the correct permissions to execute CreatePullRequest`. If a maintainer can either allow a PR from that fork or apply/cherry-pick `f64ebc3b7`, this should remove the stale-cache blocker and the merge conflict at the same time.

PR #2125, opened 2026-05-11
fix: redact Stripe customer log metadata

jihadMo:
## Summary

- stop logging raw customer email/name values during Stripe customer creation
- replace the full Stripe customer object in the stripe_info creation log with metadata-only fields
- keep Stripe customer creation, pending-customer fallback, and stripe_info writes unchanged

## Security rationale

This is a small public hardening follow-up for #1667. The existing logs include billing identity data (`email`, `name`) and the full Stripe customer object when creating org billing records. Those values are not needed for routine request tracing and can be reduced to presence flags and stable internal IDs.

## Test plan

- `git diff --check`

Not run locally: `bunx` is not installed on this Windows environment, and the checkout does not have `node_modules`, so lint/type checks are left to CI.

/claim #1667

## Release Notes

* **Chores**
  * Improved internal observability through restructured logging payloads to capture key system identifiers more efficiently.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2125)

### DIFF
--- a/supabase/functions/_backend/utils/stripe.ts
+++ b/supabase/functions/_backend/utils/stripe.ts
@@ -708,7 +708,14 @@ export interface StripeCustomer {
 }
 
 export async function createCustomer(c: Context, email: string, userId: string, orgId: string, name: string) {
-  cloudlog({ requestId: c.get('requestId'), message: 'createCustomer', email, userId, orgId, name })
+  cloudlog({
+    requestId: c.get('requestId'),
+    message: 'createCustomer',
+    userId,
+    orgId,
+    hasEmail: Boolean(email),
+    hasName: Boolean(name),
+  })
   const baseConsoleUrl = (getEnv(c, 'WEBAPP_URL') || '').replace(TRAILING_SLASHES_REGEX, '')
   const metadata: Record<string, string> = {
     user_id: userId,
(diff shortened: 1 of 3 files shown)


mingisrookie (commented):
This removes the raw email/name/customer object, but the replacement logs still keep stable billing/account identifiers in plaintext.

`createCustomer()` now logs raw `userId` and `orgId` in both the normal and "no stripe key" paths, and `createStripeCustomer()` logs `orgId: org.id` plus the Stripe product id. Those are less sensitive than email/name, but they are still durable identifiers that tie a billing-customer creation event back to a specific user/org in retained logs.

If this is meant to be metadata-only logging for the Stripe customer flow, I would mirror the pattern used in the other redaction PRs: `hasUserId`, `hasOrgId`, `hasCustomerId`, maybe `planName`/`hasStripeProductId`, but not the raw ids themselves. A small unit test that serializes these three cloudlog payloads would catch the remaining identifier exposure.

jihadMo:
Thanks, good point. I pushed 6fd851b9d to keep the Stripe customer flow logs metadata-only.

Changes:
- `createCustomer` and the no-Stripe-key path now log `hasUserId`/`hasOrgId` flags instead of raw IDs.
- `createStripeCustomer` now logs `planName` plus `hasPlanId`/`hasStripeProductId`/`hasCustomerId`/`hasOrgId`, without raw org/product/plan/customer IDs.
- Added a regression test that serializes the three cloudlog payload shapes and verifies the raw email/name/user/org/customer/product/plan identifiers are absent.

Verification:
- `vitest run tests/stripe-customer-log-redaction.unit.test.ts` -> 1 passed
- `eslint supabase/functions/_backend/utils/stripe.ts supabase/functions/_backend/utils/supabase.ts tests/stripe-customer-log-redaction.unit.test.ts`
- `git diff --check`

mingisrookie (approved):
Rechecked head `6fd851b9d` after your follow-up. The specific logging concern I raised is resolved: `createCustomer` and the no-Stripe-key branch now log only presence booleans via `getCreateCustomerLogMetadata()`, and `createInfo` logs plan/customer/org presence metadata instead of the full `selectedPlan`, `customer`, customer id, org id, email, or user id.

Local verification passed:

```text
bun x eslint supabase/functions/_backend/utils/stripe.ts supabase/functions/_backend/utils/supabase.ts tests/stripe-customer-log-redaction.unit.test.ts
bun x vitest run tests/stripe-customer-log-redaction.unit.test.ts  # 1 test passed
git diff --check 3d184f8..HEAD
```

WcaleNieWolny:

PR #2126, opened 2026-05-11
fix(api): avoid logging device request bodies

HappyReni:
## Summary

- Replace Device API request-body logging with metadata-only request logs.
- Move device log context into a small helper.
- Add a regression test that confirms app id, device id, channel, and API key values are not retained in `cloudlog` calls.

## Motivation

The public Device API currently logs the raw request body for set/get/delete operations. Those values are not needed for routine request tracing and can include stable device identifiers or private channel names. This keeps behavior unchanged while reducing unnecessary sensitive data in operational logs.

## Test Plan

- `./node_modules/.bin/vitest run tests/device-request-logging.unit.test.ts`
- `./node_modules/.bin/eslint supabase/functions/_backend/public/device/index.ts supabase/functions/_backend/public/device/logging.ts tests/device-request-logging.unit.test.ts`
- `git diff --check`

/claim #1667

### DIFF
--- a/supabase/functions/_backend/public/device/index.ts
+++ b/supabase/functions/_backend/public/device/index.ts
@@ -9,7 +9,7 @@ import { buildRateLimitInfo } from '../../utils/rateLimitInfo.ts'
 import { backgroundTask } from '../../utils/utils.ts'
 import { deleteOverride } from './delete.ts'
 import { get } from './get.ts'
-import { logDeviceRequestContext } from './logging.ts'
+import { logDeviceRateLimitRecordError, logDeviceRequestContext } from './logging.ts'
 import { post } from './post.ts'
 
 async function assertDeviceIPRateLimit(c: Context, appId: string) {
@@ -57,7 +57,7 @@ async function recordDeviceRateLimitSafely(
       await recordChannelSelfRequest(c, body.app_id, body.device_id, operation, rateLimitChannel)
     }
     catch (error) {
-      cloudlog({ requestId: c.get('requestId'), message: `Failed to record device ${operation} rate limit`, app_id: body.app_id, device_id: body.device_id, error })
+      logDeviceRateLimitRecordError(c, operation, body, error)
     }
   }
   if (body.app_id) {
--- a/supabase/functions/_backend/public/device/index.ts
+++ b/supabase/functions/_backend/public/device/index.ts
@@ -9,23 +9,8 @@ import { buildRateLimitInfo } from '../../utils/rateLimitInfo.ts'
 import { backgroundTask } from '../../utils/utils.ts'
 import { deleteOverride } from './delete.ts'
 import { get } from './get.ts'
+import { logDeviceRequestContext } from './logging.ts'
 import { post } from './post.ts'
-
-function logDeviceRequestContext(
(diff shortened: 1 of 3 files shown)


lyndon050516 (commented):
One remaining gap in the same logging surface: `recordDeviceRateLimitSafely` still logs raw `app_id` and `device_id` when rate-limit recording fails.

That is a less common path than the normal request log, but it preserves the same stable identifiers this PR is trying to keep out of operational logs. I would suggest reusing `getDeviceRequestLogMetadata(body)` there as well, or logging only `hasAppId` / `hasDeviceId` / `hasChannel` plus the operation and error. That keeps the error diagnostic useful without retaining device/app values during datastore or rate-limit outages.

The new unit test currently asserts the direct request logger only, so it would not catch this remaining error-path exposure.

HappyReni:
Thanks for catching that remaining error path. Addressed in 8a08f3640 by routing the rate-limit failure log through the same metadata-only helper and adding a regression assertion for that path. Local focused test/lint passed, and the PR checks are green.

mingisrookie (commented):
One rate-limit path still looks like it can retain the same raw device request identifiers after this change.

`assertDeviceOperationRateLimit()` now still logs the limited request with raw `app_id`, `device_id`, and `channel`, and then calls `simpleRateLimit({ app_id: body.app_id, device_id: body.device_id, ... })`. `simpleRateLimit()` itself `cloudlog`s `moreInfo` before throwing, so a throttled request can still put the raw app/device identifiers into logs even though the normal `device ${operation} request` log was converted to metadata.

A concrete trigger would be repeated device `set/get/delete` requests until `isChannelSelfRateLimited(...)` returns limited. I would suggest using `getDeviceRequestLogMetadata(body)` in this branch too, and avoiding raw `app_id`/`device_id` in the `simpleRateLimit` `moreInfo` (or returning only non-sensitive rate-limit metadata). A unit test that stubs the operation limiter as limited would cover the branch that the new logging-helper tests do not exercise.

lyndon050516:
Thanks for addressing the rate-limit logging path from my review. If Capgo is rewarding actionable security/privacy review findings on this batch, a GitHub/Algora tip to @lyndon050516 would be appreciated. No private payment details needed here; Algora/Stripe can handle that privately if you choose to tip.

darshan-Jahagirdar:
There is still an unredacted logging path for the same request fields this PR is trying to remove.

`logDeviceRequestContext()` now records only metadata, but `assertDeviceOperationRateLimit()` still calls `cloudlog` with raw `app_id`, `device_id`, and `channel` when `isChannelSelfRateLimited()` returns limited. The new test treats values like `com.private.app`, `device-secret-123`, and `beta-private` as data that must not appear in logs, but this rate-limited branch will still emit them under load/abuse.

Please route that branch through the same metadata shape, for example `getDeviceRequestLogMetadata(body)`, instead of logging the body fields directly.

riderx:
Logs are important for debug and are unique to our plugin this cannot be used for anything else.

Stop inventing yourself bullshit here.

Security report are to be reported to us not open PR this break the security good practice and is not ok

WcaleNieWolny:

PR #2128, opened 2026-05-11
fix: redact Stripe webhook event logs

jihadMo:
## Summary

- Replace full Stripe webhook event logging with metadata-only event logs.
- Apply the same redaction to the fallback unsupported-event error path.
- Add regression coverage proving customer/source payload values are not retained in log calls.

## Security rationale

Stripe webhook events can contain customer profile fields, owner email addresses, metadata, and other billing context that is not needed for routine observability. Logging only event type/id/object metadata keeps debugging signal while reducing sensitive data exposure.

## Test plan

- `git diff --check`
- `npx.cmd vitest run tests/stripe-event-logging.unit.test.ts` (not completed locally: checkout is missing `vite`/project dependencies)
- `npx.cmd eslint supabase/functions/_backend/utils/stripe_event.ts tests/stripe-event-logging.unit.test.ts` (not completed locally: checkout is missing `@antfu/eslint-config`)

/claim #1667

### DIFF
--- a/supabase/functions/_backend/utils/stripe_event.ts
+++ b/supabase/functions/_backend/utils/stripe_event.ts
@@ -104,6 +104,20 @@ function invoiceUpcoming(event: Stripe.InvoiceUpcomingEvent, data: Database['pub
   return data
 }
 
+function getStripeEventLogMetadata(event: Stripe.Event) {
+  const stripeObject = event.data?.object as { object?: unknown } | undefined
+
+  return {
+    eventId: event.id,
+    eventType: event.type,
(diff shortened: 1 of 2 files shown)


mingisrookie (approved):
Reviewed the Stripe event logging redaction. The patch replaces the full serialized Stripe event and fallback `Other event` payload with bounded metadata (`eventId`, type, object type, API version, livemode, timestamp, previous-attributes presence). I checked the remaining `cloudlog`/`cloudlogErr` calls in the touched file; customer/email/body objects still flow into the processing logic but are no longer passed to logging. No blocker from me.

WcaleNieWolny:

PR #2129, opened 2026-05-11
fix: stop logging organization admin identifiers

janishahn:
## Summary

- Stop logging invitee email/permission values from the organization members invite flow.
- Stop logging organization/user context from the organization deletion permission check.

## Motivation

Both logs are client-side debug breadcrumbs and are not needed for normal behavior. They can retain organization identifiers, user identifiers, current organization details, invitee email addresses, and role choices in browser logs.

## Test Plan

- [x] `git diff --check`
- [ ] `bunx eslint src/pages/settings/organization/Members.vue src/stores/organization.ts` (not run locally: `bunx` is not installed in this environment)

## Checklist

- [x] Change kept to removing debug logs only.
- [x] Invite and delete organization behavior unchanged.
- [x] No private technical details included.

/claim #1667

### DIFF
--- a/src/pages/settings/organization/Members.vue
+++ b/src/pages/settings/organization/Members.vue
@@ -677,8 +677,6 @@ async function showInviteModal() {
 }
 
 async function sendInvitation(email: string, type: Database['public']['Enums']['user_min_right'] | string): Promise<boolean> {
-  console.log(`Invite ${email} with perm ${type}`)
-
   const orgId = currentOrganization.value?.gid
   if (!orgId) {
     toast.error('Organization ID not found.')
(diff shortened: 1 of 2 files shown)


mingisrookie (commented):
One small cleanup gap remains in the same invitation flow: `handleSendInvitationOutput()` still does `console.log('Output: ', output)` immediately after the invite RPC returns.

Today `output` appears to be an enum-like status string, so this is much less sensitive than the removed `Invite ${email} with perm ${type}` log. But it is still an active debug log in the organization-member invite path, and if the RPC output ever grows to include more diagnostic context it would reintroduce browser-console retention in the same area this PR is cleaning up.

I would remove this line as part of the same pass (or cover the member-invite file with a simple no-`console.log` regression if the project has one), while keeping the `console.error`/`console.warn` operational error paths separate.

WcaleNieWolny:

PR #2130, opened 2026-05-11
fix(rbac): enforce API key expiration policy in permission checks

teixr12:
## Summary

Fixes an authorization gap where an API key created without `expires_at` could keep passing RBAC direct permission checks after its org later enables `require_apikey_expiration`.

## Root Cause

The API-key create/update trigger enforces expiration policy for new writes, but `rbac_check_permission_direct()` and `rbac_check_permission_direct_no_password_policy()` only checked whether a key was expired. They did not reject existing non-expiring keys when the target org starts requiring expiration.

## Fix

- Adds a runtime org-policy check in both RBAC direct permission functions after API-key identity/org resolution and before scope/permission dispatch.
- Denies non-expiring keys for orgs with `require_apikey_expiration = true`.
- Logs `RBAC_CHECK_PERM_APIKEY_EXPIRATION_REQUIRED` using key id/org/app/channel metadata only, without logging secret key material.
- Adds regression coverage for an existing scoped non-expiring key that is allowed before the policy flip and denied by both direct permission functions after the flip.

## Test plan

- `git diff --check`
- `bash scripts/check-supabase-migration-order.sh`
- Regression coverage added in `tests/rbac-permissions.test.ts`.
- Attempted targeted local Vitest with `bun run supabase:with-env -- bunx vitest run tests/rbac-permissions.test.ts`, but this checkout does not have `bun` or installed project dependencies. GitHub CI is expected to run the backend suite for the pushed branch.

## Screenshots

N/A. This is a backend SQL authorization change with regression test coverage and no frontend/CLI UI changes.

## Checklist

- [ ] My code follows the code style of this project and passes `bun run lint:backend && bun run lint`.

Refs #1667
/claim #1667

## Summary by CodeRabbit

* **New Features**
  * Orgs can enforce API-key expiration windows: expired keys, keys lacking required expirations, or keys whose expiration exceeds org limits are denied. Org scoping and 2FA requirements are enforced before granting permissions. Webhook endpoints now return 401 with an explicit "expiration_exceeds_max" error when a key’s expiry exceeds org limits.

* **Tests**
  * Added tests covering API-key expiration enforcement and webhook rejection behavior when org expiry enforcement is toggled.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2130)


teixr12:
Follow-up after the latest push:

- Mirrored channel override handling in `rbac_check_permission_direct_no_password_policy()`, including user/group deny precedence and API-key override checks.
- Updated the regression coverage to use dedicated org/app data with app-scoped role bindings instead of mutating shared fixtures.
- Preserved the existing webhook API-key policy contract by using a service-role policy lookup after local API-key org-scope validation, so non-expiring keys still return the explicit `org_requires_expiring_key` 401 response.

CI is green on the latest commit: backend tests, Playwright, benchmarks, dead-code, Socket, Sonar, and CodeRabbit all passed.

SpeedyArt (commented):
I think the runtime policy check still only covers the `require_apikey_expiration` half of the org API-key expiration policy.

The migration reads only `orgs.require_apikey_expiration` inside both direct RBAC functions and rejects only `expires_at IS NULL`. But the existing create/update trigger also enforces `orgs.max_apikey_expiration_days`: if an org later sets or lowers that max, an already-issued key with `expires_at` far beyond the new max will keep passing `rbac_check_permission_direct()` until its old expiration date.

That leaves the same retroactive-policy gap this PR is closing for non-expiring keys, just for long-lived expiring keys. I would load `max_apikey_expiration_days` alongside `require_apikey_expiration` and deny when `v_api_key.expires_at > now() + make_interval(days => max_apikey_expiration_days)`. A regression can mirror the new policy-flip test: create a scoped key expiring well beyond the future max, flip `max_apikey_expiration_days` to a shorter value, and assert both direct RBAC functions deny it.

mingisrookie (commented):
One remaining policy surface looks missed: the TS org-policy helper still only enforces `require_apikey_expiration`, not the max-expiration half that this PR now adds to the SQL RBAC checks.

Evidence: `apikeyHasOrgRightWithPolicy()` now deliberately calls `checkApikeyMeetsOrgPolicy()` via `supabaseAdmin(c)`, but that helper still selects only `require_apikey_expiration` and returns valid once the key has any `expires_at`. Callers such as the webhook permission path (`assertWebhookOrgPolicy()` before `hasOrgRightApikey()`/`check_min_rights`) rely on this helper as the API-key org-policy gate, so an org-scoped key whose `expires_at` is beyond a newly lowered `max_apikey_expiration_days` can still pass those non-`rbac_check_permission_direct()` surfaces.

Could we load `max_apikey_expiration_days` in `checkApikeyMeetsOrgPolicy()` and mirror the same `expires_at > now + max` denial used in the migration? A small regression through one helper caller (webhooks or org audit) would cover the gap that the new `rbac-permissions` tests do not exercise.

KCDaemon (approved):
Rechecked the latest head (7f7523f) against the policy gaps raised earlier. The current patch now enforces both halves of the org API-key expiration policy at runtime: non-expiring keys are denied when required, and overlong keys are compared against the key's created_at + max_apikey_expiration_days window in both the SQL RBAC functions and the TS org-policy helper.

I also checked the no-password direct path and the new regressions: the channel deny override case is covered, the overlong-key RBAC flip test covers both direct functions, and webhook listing now returns expiration_exceeds_max through the TS policy helper. GitHub checks are green, merge state is clean, and `git diff --check origin/main...origin/pr-2130` passes locally. No remaining blocker from the issues I reviewed.

austin33133-maker:
The runtime denial for `max_apikey_expiration_days` looks great overall, but I want to flag two consistency points between the SQL and TS implementations that I think could lead to the same key being allowed by one layer and denied by the other.

**1. `max_apikey_expiration_days = 0` is treated differently by SQL vs TS**

In `supabase/migrations/20260511203500_enforce_rbac_apikey_expiration_policy.sql` the SQL guard fires whenever `v_org_max_apikey_expiration_days IS NOT NULL`:

``sql
IF v_org_max_apikey_expiration_days IS NOT NULL
  AND v_api_key.expires_at IS NOT NULL
  AND (
    v_api_key.created_at IS NULL
    OR v_api_key.expires_at > v_api_key.created_at + make_interval(days => v_org_max_apikey_expiration_days)
  )
THEN ... deny
``

So if an admin sets `max_apikey_expiration_days = 0`, SQL ends up denying every key whose `expires_at > created_at` -- i.e. all expiring keys.

The TS counterpart in `supabase/functions/_backend/utils/supabase.ts` does the opposite:

``ts
if (org.max_apikey_expiration_days && key.expires_at) {
  // ... compute and compare
}
``

`0 && anything` is `0` (falsy), so the route-level check **silently skips** when max is 0. Net effect: the same key is accepted at the HTTP route boundary and then rejected at the RLS layer (or vice-versa, depending on the call path), producing inconsistent 200/401 responses for the same caller.

Two ways to reconcile:
- If 0 is meant to be "no max" / disabled, change the SQL to `v_org_max_apikey_expiration_days IS NOT NULL AND v_org_max_apikey_expiration_days > 0` and add a `CHECK` constraint on the column.
- If 0 is meant to be "deny everything with an expiration", change the TS to `org.max_apikey_expiration_days != null && key.expires_at` and document the semantic.

Either way worth a regression test that sets `max_apikey_expiration_days = 0` and asserts both layers agree on the verdict.

**2. Date arithmetic uses local-timezone JS on the TS side, UTC interval on the SQL side**

``ts
const maxDate = new Date(createdAt)
maxDate.setDate(maxDate.getDate() + org.max_apikey_expiration_days)
``

`Date#getDate()` / `setDate()` operate in the runtime's local timezone. PostgreSQL's `make_interval(days => N) + timestamptz` is timezone-stable. For most cases these agree, but two pathological cases:
- Server in a non-UTC TZ across a DST transition can shift `maxDate` by 1 hour, which matters for keys whose `expires_at` sits within an hour of the boundary.
- `created_at` strings parsed at the very end of a UTC day can flip to the next/previous local day before the `+N days` math, off-by-one-ing the verdict.

Supabase Edge Functions normally run in UTC so this is largely theoretical today, but the safer formulation is timezone-agnostic:

``ts
const maxDate = new Date(createdAt.getTime() + org.max_apikey_expiration_days * 86_400_000)
``

...and add a unit test that pins the system timezone to something non-UTC (e.g. `TZ=America/New_York`) for `checkApikeyMeetsOrgPolicy`.

**3. Minor: `created_at IS NULL` mapped to `expiration_exceeds_max`**

If `key.created_at` is null or unparseable, the TS returns `error: 'expiration_exceeds_max'`. That code's message ("API key expiration exceeds this organization's maximum allowed validity window") will mislead the consumer. A dedicated `invalid_key_timestamps` code (or reusing `org_requires_expiring_key` since these are inherently unverifiable keys) would be friendlier. SQL emits the same code in `pg_log` so the log message has the same drift.

Happy to send a tiny follow-up PR with the test cases if (1) reaches consensus.

PR #2131, opened 2026-05-11
fix: redact plugin request logs

skotchdopolepet:
/claim #1667

## Summary
- replace raw `/updates` and `/channel_self` plugin request body logs with metadata-only summaries
- carry the same metadata-only summary into the downstream update utility log so the parsed update body is not retained there either
- keep operational fields such as app id, platform, plugin version, native version, and boolean presence flags
- avoid retaining raw device IDs, custom IDs, key IDs, channel names, or default channel names in routine cloud logs

## Test
- `npx --yes vitest@4.1.5 run tests/plugin-request-logging.unit.test.ts --config <temp minimal config>`
- `git diff --check`
- `npx --yes esbuild@0.27.2 supabase/functions/_backend/utils/update.ts --bundle --platform=node --format=esm --packages=external --outfile=NUL`

I used a minimal Vitest config locally because this environment does not have Bun installed.

### DIFF
--- a/supabase/functions/_backend/plugins/channel_self.ts
+++ b/supabase/functions/_backend/plugins/channel_self.ts
@@ -14,6 +14,7 @@ import { sendNotifOrgCached } from '../utils/notifications.ts'
 import { sendNotifToOrgMembersCached } from '../utils/org_email_notifications.ts'
 import { closeClient, deleteChannelDevicePg, getAppByIdPg, getAppOwnerPostgres, getChannelByNamePg, getChannelDeviceOverridePg, getChannelsPg, getCompatibleChannelsPg, getDrizzleClient, getMainChannelsPg, getPgClient, setReplicationLagHeader, upsertChannelDevicePg } from '../utils/pg.ts'
 import { convertQueryToBody, makeDevice, parsePluginBody } from '../utils/plugin_parser.ts'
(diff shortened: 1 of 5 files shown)


mingisrookie (commented):
There are still a couple of rate-limit paths that bypass the new plugin request summary and can log the original request values.

In both `plugins/updates.ts` and `parseChannelSelfPluginRequest()` in `plugins/channel_self.ts`, the `isLimited(c, body.app_id)` branch still returns `simpleRateLimit(body)`. `simpleRateLimit()` logs its `moreInfo`, so a globally limited app request would still write the full plugin body (`device_id`, `custom_id`, `key_id`, channel/defaultChannel, etc.) to logs despite the normal request log now using `summarizePluginRequestForLog()`.

There is a similar per-device set limiter later in `channel_self.ts`: the limited branch logs raw `app_id`, `device_id`, and `channel`, then calls `simpleRateLimit({ app_id: bodyParsed.app_id, device_id: bodyParsed.device_id, ... })`, which logs those raw values again.

I would reuse `summarizePluginRequestForLog(body)` (or a smaller rate-limit metadata object) for these `simpleRateLimit` calls and add a test that forces `isLimited` / `isChannelSelfRateLimited` to return limited. The new unit test only covers the helper output directly, so it would not catch these remaining route-level logging paths.

WcaleNieWolny:

PR #2132, opened 2026-05-11
fix: stop returning api keys in statistics errors

janishahn:
## Summary

- replace the raw API key value in /statistics/org invalid-key error details with the API key id
- keep the existing 401 behavior for out-of-scope org/app-limited keys
- add a regression assertion that the rejected statistics response does not contain the plain statistics API key or a moreInfo.data payload

## Security impact

The previous invalid-key branches could put a valid API key into both the client-visible error JSON and the structured error log. This keeps enough non-secret context for debugging without returning the credential.

/claim #1667

## Test plan

- [x] git diff --check
- [ ] bun run supabase:with-env -- bunx vitest run tests/statistics.test.ts (local bun runtime is unavailable in this checkout environment)

### DIFF
--- a/supabase/functions/_backend/public/statistics/index.ts
+++ b/supabase/functions/_backend/public/statistics/index.ts
@@ -909,12 +909,12 @@ app.get('/org/:org_id', async (c) => {
   }
   if (auth.authType === 'apikey' && auth.apikey!.limited_to_orgs && auth.apikey!.limited_to_orgs.length > 0) {
     if (!auth.apikey!.limited_to_orgs.includes(orgId)) {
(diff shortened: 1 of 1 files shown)


mingisrookie (approved):
Reviewed the statistics API-key error redaction. The two org-stats API-key scope failures no longer return `auth.apikey.key` and now use the database id instead, so the raw secret is not put into the `quickError` moreInfo for these paths. The surrounding permission/payment/statistics behavior is unchanged. No blocker from me; a small regression test around a scoped API key hitting the wrong org/app scope would be a useful follow-up.

WcaleNieWolny:

PR #2134, opened 2026-05-11
fix(db): restrict apikey table management RLS

hieu-lee:
## Summary

- Restrict direct `public.apikeys` SELECT/DELETE RLS to JWT callers or unrestricted `all` API keys by reusing `get_identity_for_apikey_creation()`
- Keep the generated schema dump aligned with the new migration
- Extend the API-key security pgTAP test so an org/app-limited key cannot enumerate or delete sibling API-key rows through direct table access

/claim #1667

## Motivation

The Edge API-key management endpoints already reject org/app-limited API keys. The table policies still allowed direct PostgREST SELECT/DELETE for any owner key with broad enough legacy mode, which left a bypass around those endpoint checks.

## Test Plan

- [x] `git diff --check`
- [x] Static assertion check: `53_test_apikey_creation_security.sql` now has 10 planned assertions and 10 assertion calls
- [ ] Supabase pgTAP not run locally because this laptop cannot run Docker

## Checklist

- [x] Security-sensitive RLS change is limited to API-key table SELECT/DELETE management policies
- [x] Schema dump is aligned with the migration
- [x] Regression coverage documents the limited-key direct table bypass
- [x] No application API behavior is intentionally changed beyond closing the direct PostgREST bypass


hieu-lee:
Leaving the policy target as `TO "anon", "authenticated"` intentionally. API-key requests in this codebase use the anon Supabase key plus a `capgkey` header via `supabaseApikey()`, so authenticated-only RLS would also block unrestricted API-key management. The guard here is `get_identity_for_apikey_creation()`: it still allows JWT callers and unrestricted `all` keys, but rejects org/app-limited keys.

PR #2135, opened 2026-05-11
fix(frontend): avoid client-side replication secret

hieu-lee:
## Summary

- Remove the `VITE_REPLICATION_API_SECRET` fallback from the admin replication dashboard
- Always authenticate browser replication status requests with the current admin Supabase JWT
- Keep the backend `apisecret` path untouched for server-side/internal callers

/claim #1667

## Motivation

`VITE_` variables are bundled into client JavaScript. If a replication API secret is configured there, any browser that can fetch the admin bundle can recover and reuse it. The page already supports admin JWT authorization, so the browser should not ship a reusable shared secret.

## Test Plan

- [x] `git diff --check`
- [x] `bunx eslint src/pages/admin/dashboard/replication.vue`

### DIFF
--- a/src/pages/admin/dashboard/replication.vue
+++ b/src/pages/admin/dashboard/replication.vue
@@ -99,25 +99,19 @@ const checkedAt = computed(() => {
   return formatLocalDateTime(data.value.checked_at)
 })
 
-const internalReplicationSecret = import.meta.env.VITE_REPLICATION_API_SECRET as string | undefined
-
 async function loadReplicationStatus() {
   isLoading.value = true
   errorMessage.value = null
 
   try {
-    const headers: Record<string, string> = {}
-    if (internalReplicationSecret) {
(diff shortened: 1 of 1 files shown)


e4792200-stack (approved):
Reviewed this focused frontend hardening change. The admin replication dashboard no longer reads `VITE_REPLICATION_API_SECRET` or sends an `apisecret` from the browser; it now always derives the current Supabase session and calls `/replication` with the user JWT.

I also checked the backend `/replication` access path: JWT callers still go through `getClaimsFromJWT` plus the `is_platform_admin` RPC, while the `apisecret` branch remains available for internal/server callers. The default-branch code search only showed the Vite replication secret in this dashboard file, and this patch removes that reference. CI is green, including `Run tests`, Playwright, SonarCloud, Socket, and dead-code check. No blockers from me.

PR #2136, opened 2026-05-11
fix: redact private stats request logs

skotchdopolepet:
## Summary
- replace raw `/private/stats` and `/private/stats/export` request-body logging with metadata-only summaries
- avoid logging raw device filters, search text, export filenames, order entries, or action values before the app log-read permission check completes
- add focused unit coverage for the stats request log summarizer

## Test
- `npx --yes vitest@4.1.5 run tests/private-stats-request-logging.unit.test.ts --config $tmp`

/claim #1667

### DIFF
--- a/supabase/functions/_backend/private/stats.ts
+++ b/supabase/functions/_backend/private/stats.ts
@@ -9,6 +9,7 @@ import { toCsv } from '../utils/csv.ts'
 import { parseBody, simpleError, useCors } from '../utils/hono.ts'
 import { middlewareV2 } from '../utils/hono_middleware.ts'
 import { cloudlog } from '../utils/logging.ts'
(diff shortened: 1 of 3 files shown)


mingisrookie (commented):
I think one sensitive field is still leaking in the new metadata helper: `summarizePrivateStatsRequestForLog()` returns `app_id: body?.appId`, and `getValidatedStatsRequestBody()` logs that summary before the `checkPermission(c, 'app.read_logs', { appId: body.appId })` gate runs.

That means an unauthorized caller can still cause arbitrary raw app identifiers to be written to the `post private/stats body` / `post private/stats/export body` logs. Since the PR summary says these request logs should be metadata-only and avoid raw request-body logging before the permission check, `appId` should probably be treated like the other filters here (for example `has_app_id` or an app-id length/presence flag) rather than logging the raw value.

WcaleNieWolny:

PR #2137, opened 2026-05-11
Reduce Stripe helper log exposure

ocepkejepintu-droid:
## Summary
- replace raw Stripe price/subscription-item log payloads with compact summaries
- stop routine Stripe customer creation logs from retaining email/name values
- keep checkout/customer behavior unchanged

Related to #1667.

## Verification
- `bunx eslint supabase/functions/_backend/utils/stripe.ts`
- `bun typecheck`
- `git diff --check`

### DIFF
--- a/supabase/functions/_backend/utils/stripe.ts
+++ b/supabase/functions/_backend/utils/stripe.ts
@@ -729,7 +729,16 @@ export interface StripeCustomer {
 }
 
 export async function createCustomer(c: Context, email: string, userId: string, orgId: string, name: string) {
-  cloudlog({ requestId: c.get('requestId'), message: 'createCustomer', userId, orgId, hasEmail: Boolean(email), hasName: Boolean(name) })
+  const stripeConfigured = isStripeConfigured(c)
+  cloudlog({
+    requestId: c.get('requestId'),
+    message: 'createCustomer',
+    hasUserId: Boolean(userId),
+    hasOrgId: Boolean(orgId),
+    hasEmail: Boolean(email),
+    hasName: Boolean(name),
+    stripeConfigured,
+  })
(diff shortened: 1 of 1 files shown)


ocepkejepintu-droid:
Follow-up: CI is green now. Local verification also passed with `bunx eslint supabase/functions/_backend/utils/stripe.ts`, `bun typecheck`, and `git diff --check`.

SpeedyArt (commented):
The `createCustomer` log redaction still leaves stable internal identifiers on the changed log lines.

This patch replaces raw `email`/`name` with booleans, but both `createCustomer` and `createCustomer no stripe key` still log raw `userId`, and the main `createCustomer` log still logs raw `orgId`. For the same retained-log threat model as #1667, those are linkable account/org identifiers even when the email/name values are removed.

I would make these metadata-only as well, e.g. `hasUserId`, `hasOrgId`, `hasEmail`, `hasName`, and possibly `stripeConfigured`. The Stripe customer metadata can still receive the actual IDs for Stripe synchronization, but the cloud logs do not need to retain them.

ocepkejepintu-droid:
Addressed in `484c4b85d`. The `createCustomer` logs now use `hasUserId` / `hasOrgId` presence flags and keep `stripeConfigured`, `hasEmail`, and `hasName` as metadata only. The actual `user_id` / `org_id` metadata still goes to Stripe where needed for synchronization, but cloud logs no longer retain those raw identifiers.

Verification passed again with `bunx eslint supabase/functions/_backend/utils/stripe.ts`, `bun typecheck`, and `git diff --check`.

ocepkejepintu-droid:
Follow-up: remote CI is green now after `484c4b85d`; the full test job and Playwright job both passed. No behavior change beyond keeping the Stripe helper cloud logs metadata-only.

WcaleNieWolny:

PR #2138, opened 2026-05-11
fix: redact queue consumer request logs

jihadMo:
/claim #1667

## Summary
- stop logging the full queue trigger URL and raw queued POST payload before dispatch
- keep debug-safe metadata only: target kind, function name, payload type, payload size, and top-level key count
- add regression coverage proving queued POST log metadata does not include payload keys or sensitive values

## Why
Queued trigger payloads can contain customer identifiers, emails, tokens, or webhook data. The previous pre-dispatch log wrote the resolved endpoint and entire payload into backend logs. This keeps operational visibility while reducing sensitive data exposure.

## Tests
- `git diff --check`
- `npx.cmd --yes vitest run tests/queue-consumer-message-shape.unit.test.ts` *(blocked locally because this checkout has no installed dependencies; Vitest could not resolve `vite`)*

## Summary by CodeRabbit

* **Refactor**
  * Improved queue processing and structured POST logging for clearer diagnostics and more reliable message dispatching.

* **Bug Fixes**
  * Tighter sanitization of error payloads to prevent leaking sensitive values; better handling of per-message failures so one bad message won't fail entire batches.
  * Enforced request timeouts to abort stalled POSTs.

* **Tests**
  * Expanded tests for logging metadata, target classification, error redaction, batch failure handling, and timeout behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2138)

### DIFF
--- a/supabase/functions/_backend/triggers/queue_consumer.ts
+++ b/supabase/functions/_backend/triggers/queue_consumer.ts
@@ -435,7 +435,7 @@ async function extractErrorDetails(response: Response): Promise<{
     return {
       bodyPreview,
       errorCode: typeof errorCode === 'string' ? errorCode : null,
-      errorMessage: typeof errorMessage === 'string' ? errorMessage : null,
+      errorMessage: typeof errorMessage === 'string' ? sanitizeDiscordResponseBody(errorMessage) : null,
     }
   }
   return {
--- a/supabase/functions/_backend/triggers/queue_consumer.ts
+++ b/supabase/functions/_backend/triggers/queue_consumer.ts
@@ -538,7 +538,7 @@ export async function http_post_helper(
       method: 'POST',
       headers,
       body: JSON.stringify(body),
-      // signal: controller.signal,
+      signal: controller.signal,
     })
     return response
   }
--- a/supabase/functions/_backend/triggers/queue_consumer.ts
+++ b/supabase/functions/_backend/triggers/queue_consumer.ts
@@ -54,6 +54,8 @@ interface FailureDetail {
   cf_id: string
 }
 
+type QueuePostTargetKind = 'cloudflare' | 'cloudflare_legacy' | 'cloudflare_pp' | 'supabase'
+
 function extractMessageBody(message: Message): Record<string, unknown> {
   if (message.message?.payload !== undefined)
     return (message.message.payload ?? {}) as Record<string, unknown>
@@ -161,6 +163,54 @@ function sanitizeDiscordResponseBody(value: string): string {
 // Helper function to generate UUID v4
 function generateUUID(): string {
   return crypto.randomUUID()
+}
+
+function getQueuePostTargetKind(functionType: string | null | undefined, cfPpUrl: string, cfUrl: string): QueuePostTargetKind {
+  const normalizedType = (functionType ?? '').trim()
+
+  if (normalizedType === 'cloudflare_pp' && cfPpUrl)
+    return 'cloudflare_pp'
+  if (normalizedType === 'cloudflare' && cfUrl)
+    return 'cloudflare'
+  if (normalizedType === '' && cfUrl)
+    return 'cloudflare_legacy'
+  return 'supabase'
+}
+
+function getQueuePayloadLogMetadata(body: unknown): {
+  payloadKeys: number
+  payloadSize: number
+  payloadType: string
+} {
+  let payloadSize = 0
+  try {
+    payloadSize = JSON.stringify(body)?.length ?? 0
+  }
+  catch {
+    payloadSize = 0
+  }
+
+  return {
+    payloadKeys: body && typeof body === 'object' && !Array.isArray(body) ? Object.keys(body).length : 0,
+    payloadSize,
+    payloadType: Array.isArray(body) ? 'array' : typeof body,
+  }
+}
+
+function getQueuePostLogMetadata(functionName: string, targetKind: QueuePostTargetKind, body: unknown): {
+  functionName: string
+  message: string
+  payloadKeys: number
+  payloadSize: number
+  payloadType: string
+  targetKind: QueuePostTargetKind
+} {
+  return {
+    functionName,
+    message: `[${functionName}] Making queued HTTP POST request.`,
+    targetKind,
+    ...getQueuePayloadLogMetadata(body),
+  }
 }
 
 async function processQueue(c: Context, db: ReturnType<typeof getPgClient>, queueName: string, batchSize: number = DEFAULT_BATCH_SIZE) {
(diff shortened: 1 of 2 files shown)


SpeedyArt (commented):
One residual alert path still keeps raw response text even though the body preview is sanitized.

`extractErrorDetails()` pulls `payload.message ?? payload.errorMessage` into `errorMessage`, and the Discord "Detailed Failures" field later appends that value directly via `detail.error_message.replace(...).trim()`. The separate `response_body` field does go through `sanitizeDiscordResponseBody()`, but the extracted message from the same response bypasses that sanitizer.

If a downstream queued function returns a JSON error like `{ "message": "invalid token <value> for user@example.com" }`, the sanitized body section would hide it, while the detailed failure line would still publish the raw token/email to Discord. I would sanitize `errorMessage` at extraction time, or apply `sanitizeDiscordResponseBody()` before composing `messageInfo`, and add a regression where `message` contains an email/token while `response_body` is already clean.

jihadMo:
Thanks, good catch. Fixed in e874683 by sanitizing extracted JSON error messages before they are reused in the detailed failure field, and added a regression covering an email/token in the JSON message.

kendalework (commented):
Reviewed this against the security hardening goal in #1667.

I did not find remaining raw queued POST payload or resolved trigger URL logging in `http_post_helper`; the new log shape keeps operational metadata (`targetKind`, function name, payload type/size/key count) without serializing field names or values. I also checked the JSON error response path: `errorMessage` now goes through the existing sanitizer, which closes the remaining Discord failure-report leak path from structured error bodies.

Local verification on `e874683a2efd35d0e5841605314f411d6e0870f0`:
- `git diff --check origin/main...HEAD`
- `bunx vitest run tests/queue-consumer-message-shape.unit.test.ts` -> 11 passed
- `bunx eslint supabase/functions/_backend/triggers/queue_consumer.ts tests/queue-consumer-message-shape.unit.test.ts`

No blocking issues found from my review.

mingisrookie (commented):
One reliability regression in the new timeout path: `http_post_helper()` creates an `AbortController` and a 15s timer, but the fetch still has `signal: controller.signal` commented out.

So in practice the timeout never aborts a hung queue target. That means a slow or stuck downstream function can block the queue worker past the visibility window, which is exactly the case this helper is trying to protect.

Please wire the signal into `fetch()` and add a test for a delayed or never-resolving request so this does not silently regress again.

jihadMo:
Thanks, good catch. I pushed 6e03b1c33 to wire the existing AbortController into the queued POST fetch call.

I also added a regression test with fake timers and a never-resolving fetch to verify the helper passes the abort signal and rejects once the 15s timeout elapses.

Verification:
- `vitest run tests/queue-consumer-message-shape.unit.test.ts` -> 12 passed
- `eslint supabase/functions/_backend/triggers/queue_consumer.ts tests/queue-consumer-message-shape.unit.test.ts`
- `git diff --check`

mingisrookie (approved):
Rechecked head `6e03b1c33` after your follow-up. The timeout regression I flagged is resolved: `http_post_helper()` now passes `controller.signal` into `fetch()`, and the new fake-timer regression test proves the signal aborts after 15s. I also verified the queued POST log metadata stays URL/body-value free and the JSON error-message path is sanitized before Discord details reuse it.

Local verification passed:

```text
bun x eslint supabase/functions/_backend/triggers/queue_consumer.ts tests/queue-consumer-message-shape.unit.test.ts
bun x vitest run tests/queue-consumer-message-shape.unit.test.ts  # 12 tests passed
git diff --check 3d184f8..HEAD
```

supabase/functions/_backend/triggers/queue_consumer.ts
```diff
@@ -480,12 +530,15 @@ export async function http_post_helper(
...
+    cloudlog({
+      requestId: c.get('requestId'),
+      ...getQueuePostLogMetadata(function_name, targetKind, body),
+    })
     const response = await fetch(url, {
       method: 'POST',
       headers,
       body: JSON.stringify(body),
-      // signal: controller.signal,
+      signal: controller.signal,
```
darshan-Jahagirdar, line 617: This activates the timeout path, but `processQueue` still awaits `http_post_helper` inside a `Promise.all` without per-message error handling. If one queued target hangs for 15s, the abort makes `http_post_helper` throw and the whole `processQueue` rejects before building `failureDetails`, updating CF IDs, deleting successes, or sending the exhausted-retry Discord alert. Other messages in the same batch can finish but then never get accounted for by the queue bookkeeping in this run.

Could we convert helper exceptions into a synthetic failure response/detail for that message, or switch the batch processing to per-message `try/catch`/`Promise.allSettled`, so the newly enforced timeout is handled like the existing non-2xx failures instead of aborting the whole batch?

jihadMo:
Thanks, good catch. I pushed 67183518e to keep queue batch accounting intact when a queued POST throws.

What changed:
- Moved per-message dispatch into `processQueueMessage()`.
- `http_post_helper` exceptions now become a synthetic failed result for that message instead of rejecting the whole `Promise.all` batch.
- Timeout errors are represented as `408 Request Timeout` with `request_timeout`, so CF ID updates, success deletion, failure details, and exhausted-retry Discord alert logic can still run for the batch.
- Added a regression that processes one throwing message and one successful message together and verifies the batch resolves with both results accounted for.

Verification:
- `vitest run tests/queue-consumer-message-shape.unit.test.ts` -> 13 passed
- `eslint supabase/functions/_backend/triggers/queue_consumer.ts tests/queue-consumer-message-shape.unit.test.ts`
- `git diff --check`
- `vue-tsc --noEmit --pretty false` still stops only on the known unrelated CLI test/module issues (`tests/cli-channel.test.ts`, `tests/cli-hashed-apikey.test.ts`, `tests/cli-sdk-utils.ts`); no queue-consumer errors remain.

jihadMo:
Thanks, addressed in 2b98f27eb:
- switched the two new tests to `it.concurrent(...)`
- replaced the rejection `.then(...)` chain with an attached `expect(request).rejects...` assertion

Verification:
- `vitest run tests/queue-consumer-message-shape.unit.test.ts` -> 13 passed
- `eslint tests/queue-consumer-message-shape.unit.test.ts`
- `git diff --check`

mingisrookie (approved):
Rechecked current head `2b98f27e` after the follow-up commits. The new per-message queue POST failure path keeps timeouts represented as failed results, so one slow target no longer rejects the whole batch before CF ID updates and failure handling can run. The test-style follow-up is addressed, and current head checks are green, including Run tests and SonarCloud.

WcaleNieWolny:

PR #2139, opened 2026-05-11
fix: redact private devices request logs

janishahn:
/claim #1667

Summary:
- Replace the /private/devices raw request-body cloudlog with shape-only metadata.
- Keep device listing/count behavior unchanged while avoiding routine retention of app IDs, device IDs, version filters, search text, order values, and cursors.
- Add focused unit coverage for the log metadata helper.

Test plan:
- git diff --check
- Not run locally: bun and node_modules are unavailable in this checkout.

### DIFF
--- a/supabase/functions/_backend/private/devices.ts
+++ b/supabase/functions/_backend/private/devices.ts
@@ -10,7 +10,7 @@ import { appIdSchema, cursorSchema, deviceIdSchema, hasInvalidQueryLimitInput, h
 import { checkPermission } from '../utils/rbac.ts'
 import { countDevices, readDevices } from '../utils/stats.ts'
 
-interface DataDevice {
+export interface DataDevice {
   appId: string
   count?: boolean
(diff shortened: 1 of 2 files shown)


e4792200-stack (approved):
Reviewed the private devices logging change. The previous `cloudlog` call recorded the parsed request body before the permission check, which could retain app IDs, device IDs, version/search filters, ordering, and cursors. This patch keeps the read/count behavior and downstream `checkPermission`/`readDevices` calls unchanged, but logs only booleans and counts via `getPrivateDevicesRequestLogMetadata`.

I also checked the added unit test: it covers the sensitive request fields and asserts that the generated log metadata does not retain the original app ID, version, device ID, search term, or cursor values. The full GitHub checks are now green, including `Run tests`, Playwright, SonarCloud, Socket, and dead-code check. No blockers from me.

WcaleNieWolny:

PR #2140, opened 2026-05-11
Reduce update request log exposure

ocepkejepintu-droid:
## Summary
- replace the full plugin update request-body log with a compact summary
- keep updater behavior unchanged while avoiding routine retention of device/custom identifiers
- leave enough app/platform/version-shape context for diagnostics

Related to #1667.

## Verification
- `bunx eslint supabase/functions/_backend/utils/update.ts`
- `bun typecheck`
- `git diff --check`

### DIFF
--- a/supabase/functions/_backend/utils/update.ts
+++ b/supabase/functions/_backend/utils/update.ts
@@ -48,6 +48,23 @@ const UPDATE_BLOCKED_CODES = new Set([
   'key_id_mismatch',
 ])
 
+function summarizeUpdateRequest(body: AppInfos) {
+  return {
+    app_id: body.app_id,
+    platform: body.platform,
+    plugin_version: body.plugin_version,
+    version_name: body.version_name,
+    version_os: body.version_os,
+    hasVersionBuild: Boolean(body.version_build),
+    hasDeviceId: Boolean(body.device_id),
+    hasCustomId: Boolean(body.custom_id),
+    hasDefaultChannel: Boolean(body.defaultChannel),
+    hasKeyId: Boolean(body.key_id),
+    is_prod: body.is_prod,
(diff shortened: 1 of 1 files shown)


ocepkejepintu-droid:
Follow-up: CI is green now. Local verification also passed with `bunx eslint supabase/functions/_backend/utils/update.ts`, `bun typecheck`, and `git diff --check`.

SpeedyArt (commented):
There is still a raw update-request log later in the same handler that seems to undo part of this redaction.

The new `summarizeUpdateRequest()` correctly changes `device_id` and `custom_id` into presence booleans for the initial `update request` log, but after `const device = makeDevice(...)`, the handler still does `cloudlog({ requestId: c.get('requestId'), message: 'vals', platform, device })`. `makeDevice()` carries the raw `device_id`, `custom_id`, `app_id`, `version_build`, `version_name`, plugin version, and OS version from the request, so successful update checks will still persist stable device identifiers and custom IDs in logs.

I would replace that branch with the same metadata-only approach, for example `hasDeviceId`, `hasCustomId`, platform/plugin/version presence flags, or remove the log if it is only leftover debugging. Otherwise the request-summary redaction only protects the first log line, not the later normal path.

ocepkejepintu-droid:
Addressed in `a8f7890d2`. The later `vals` log now wraps `makeDevice()` with a metadata-only summary, so it keeps platform/build/plugin/OS presence flags and booleans for `device_id` / `custom_id` instead of persisting the raw device object.

Verification passed again with `bunx eslint supabase/functions/_backend/utils/update.ts`, `bun typecheck`, and `git diff --check`.

WcaleNieWolny:

PR #2141, opened 2026-05-11
fix(frontend): keep invitation tokens out of URLs

hieu-lee:
## Summary

- Set the Supabase session directly after accepting an invitation
- Redirect accepted invitees to the dashboard with `router.replace` instead of forwarding access/refresh tokens through `/login?...`
- Preserve existing captcha reset and error handling when the invitation or session setup fails

/claim #1667

## Motivation

The invitation flow currently receives Supabase access and refresh tokens from `private/accept_invitation`, then places both tokens in the browser URL query string while navigating to `/login`. Even though the login page later scrubs the URL, those tokens can be exposed to browser history, logs, screenshots, extensions, or client-side telemetry during the redirect. The invitation page already has the Supabase client, so it can establish the session without placing credentials in navigation state.

## Test Plan

- [x] `git diff --check`
- [x] `bunx eslint src/pages/invitation.vue`

### DIFF
--- a/src/pages/invitation.vue
+++ b/src/pages/invitation.vue
@@ -137,13 +137,18 @@ async function submitForm() {
       throw new Error(error.message || 'Failed to accept invitation')
     }
 
-    // Store tokens in local storage or cookies
     if (data?.access_token && data?.refresh_token) {
-      // Login successful, redirect to dashboard
-      // window.location.href = '/dashboard';
-      router.push(`/login?access_token=${data.access_token}&refresh_token=${data.refresh_token}`)
+      const { error: sessionError } = await supabase.auth.setSession({
+        access_token: data.access_token,
+        refresh_token: data.refresh_token,
+      })
 
-      // MagicCapgo12@#
+      if (sessionError) {
+        captchaComponent.value?.reset()
+        throw new Error(sessionError.message || 'Failed to start session')
+      }
+
+      await router.replace('/dashboard')
     }
     else {
       captchaComponent.value?.reset()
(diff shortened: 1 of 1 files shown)


hieu-lee:
The first `Run tests` failure was an unrelated CI resource failure in `tests/private-error-cases.test.ts` (`Timed out acquiring connection from connection pool`). This PR only touches `src/pages/invitation.vue`. I do not have permission to rerun the failed job directly, so I pushed an empty retry commit (`325d91e`) to re-trigger CI without changing the patch.

SpeedyArt (commented):
One residual URL-secret path still seems present here: `invite_magic_string` is copied into `inviteMagicString.value`, but the query param is never removed from the current URL.

That means the invitation token stays in the address bar and browser history for the whole accept-invitation flow, including the error/loading states. After `inviteRow` loads, the page can also render external org media such as `inviteRow.org_logo`; depending on the browser/referrer policy and the image host, the full invitation URL can still be exposed as a referrer before the user submits the form.

I would scrub the query immediately after caching the token, before the RPC finishes and before rendering org content. For example, once `inviteMagicString.value` is set, call `router.replace({ path: route.path, query: {} })` or `window.history.replaceState` with `invite_magic_string` removed, while continuing to use the cached ref for `accept_invitation`.

hieu-lee:
Updated in `8d443fb`.

The page now copies `invite_magic_string` into the local `inviteMagicString` ref, immediately removes that query parameter from the current URL/history with `replaceState`, and only then runs `get_invite_by_magic_lookup`. The cached ref is still used for `accept_invitation`, so the token no longer remains in the address bar while org logo/content renders.

Local checks: `npx eslint src/pages/invitation.vue` and `git diff --check` pass.

PR #2143, opened 2026-05-11
fix: redact Bento log payload details

strongkeep-debug:
/claim #1667

## Summary
- Redacts Bento email-automation logs so recipient emails, tag names, command payloads, third-party result bodies, and third-party error text are not retained in application logs.
- Replaces raw Bento responses with compact status/count summaries while keeping useful failure signals for debugging.
- Keeps runtime behavior unchanged: Bento requests, return values, and success/failure handling remain the same apart from safer log/error surfaces.

## Scope
- Public-safe log-retention hardening only.
- No live Bento requests were made during verification.
- Tests use mocked Bento responses and mocked log sinks to assert sensitive values are absent from emitted logs.

## Verification
- Focused Vitest coverage for `tests/bento-log-redaction.unit.test.ts`: 5 tests passed locally.
- `git diff --check` passed.
- CI is green after the latest push: `changes`, `Run tests`, `Run Playwright tests`, `Check dead code`, Socket Security, SonarCloud, CodeRabbit, and CodSpeed all pass.

## Notes
This PR is intentionally narrow: it fixes one concrete retention surface and adds regression coverage for the sensitive values that should not appear in routine logs.


cherry-bisht:
C:\Users\vegas\Documents\Codex\2026-05-10\goal-make-me-5-and-do\bounty-work  , thats your directory ? lmao  , though the fix is correct to a certain point , its still ai slop

PR #2144, opened 2026-05-11
fix: redact user trigger record logs

jonathanangel-1:
## Summary

- stop user create/update/delete triggers from writing full webhook records to routine logs
- add a shared `summarizeRecordForLog` helper that keeps selected IDs and presence flags without retaining emails, avatar paths, or profile metadata
- use the same summary in the user-update `no_id` error details

/claim #1667

## Test plan

- `bunx vitest run tests/logging-record-summary.unit.test.ts`
- `bunx eslint supabase/functions/_backend/utils/logging.ts supabase/functions/_backend/triggers/on_user_create.ts supabase/functions/_backend/triggers/on_user_update.ts supabase/functions/_backend/triggers/on_user_delete.ts tests/logging-record-summary.unit.test.ts`
- `git diff --check`

## Screenshots

N/A - backend log hardening only.

## Checklist

- [ ] My code follows the code style of this project and passes
      `bun run lint:backend && bun run lint`.
- [ ] I have [updated the documentation](https://github.com/Cap-go/website)
      accordingly.
- [x] I have tested my code manually, and I have provided steps how to reproduce
      my tests

## Summary by CodeRabbit

* **Refactor**
  * Improved logging for user create/update/delete flows to emit compact, non-sensitive summaries of user data while preserving existing behavior.

* **Tests**
  * Added unit tests ensuring log summaries omit raw sensitive values and correctly report presence flags and field counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2144)

### DIFF
--- a/supabase/functions/_backend/triggers/on_user_create.ts
+++ b/supabase/functions/_backend/triggers/on_user_create.ts
@@ -2,7 +2,7 @@ import type { MiddlewareKeyVariables } from '../utils/hono.ts'
 import type { Database } from '../utils/supabase.types.ts'
 import { Hono } from 'hono/tiny'
 import { BRES, middlewareAPISecret, triggerValidator } from '../utils/hono.ts'
-import { cloudlog } from '../utils/logging.ts'
+import { cloudlog, summarizeRecordForLog } from '../utils/logging.ts'
 import { createApiKey } from '../utils/supabase.ts'
 import { sendEventToTracking } from '../utils/tracking.ts'
 import { syncUserPreferenceTags } from '../utils/user_preferences.ts'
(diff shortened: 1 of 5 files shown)


SpeedyArt (commented):
Thanks for tightening these logs. One residual identifier still looks inconsistent with the redaction goal: `summarizeRecordForLog` copies every configured `idFields` value verbatim, and the user create/update/delete handlers all pass `idFields: ['id']`. That means the retained trigger logs still include the raw `users.id`, and the new unit test explicitly expects `id: 'user-123'`.

For this #1667-style logging cleanup, a stable user id in retained logs is still linkable user data, especially when the other log-redaction PRs are moving identifiers to booleans/counts like `hasUserId`/`hasId`. Could these user-trigger call sites log a presence flag, or another non-reversible diagnostic, instead of persisting `record.id` directly?

jonathanangel-1:
Updated in 74e75ca: user trigger logs now use `has_id` presence flags instead of retaining raw `users.id`, and the helper no longer supports copying configured id fields verbatim. Focused Vitest, ESLint, and diff checks passed locally.

mingisrookie (commented):
The latest summary helper no longer copies `record.id` into the trigger-record log, which addresses the earlier stable-id concern there. One separate user-delete log still keeps the raw user id, though: after cleanup completes, `deleteUser()` logs

`cloudlog({ context: 'user deletion completed', duration_ms: duration, user_id: record.id })`

That is outside the new `summarizeRecordForLog()` path and is not covered by the helper unit test. If the intent is to keep user-trigger operational logs metadata-only, this should probably become `has_user_id: true` (or omit the id and keep the duration only). A regression test around the delete-success path, or a direct assertion that serialized logs do not contain the sample user id, would catch this remaining exposure.

WcaleNieWolny:

PR #2146, opened 2026-05-11
fix(api): redact password compliance logs

HappyReni:
/claim #1667

## Summary
- replace validate_password_compliance request logs with metadata-only summaries
- stop retaining raw email, password, captcha token, organization id, and extra request fields in routine cloud logs
- redact the success log so it records identifier presence without user/org ids

## Motivation
This is a small hardening follow-up for the public security rewards thread. The endpoint already avoided logging the password and captcha token directly, but the request log still retained email/org identifiers and the success log retained user/org ids. Those values are not needed for routine diagnostics.

## Test Plan
- `./node_modules/.bin/vitest run tests/password-compliance-logging.unit.test.ts`
- `./node_modules/.bin/vitest run tests/account-rate-limit.unit.test.ts tests/password-compliance-logging.unit.test.ts`
- `./node_modules/.bin/eslint supabase/functions/_backend/private/validate_password_compliance.ts supabase/functions/_backend/private/validate_password_compliance_logging.ts tests/password-compliance-logging.unit.test.ts`
- `git diff --check`

### DIFF
--- a/supabase/functions/_backend/private/validate_password_compliance.ts
+++ b/supabase/functions/_backend/private/validate_password_compliance.ts
@@ -12,6 +12,7 @@ import { clearFailedAccountAuth, isAccountRateLimited, isIPRateLimited, recordFa
 import { buildRateLimitInfo } from '../utils/rateLimitInfo.ts'
 import { emptySupabase, supabaseClient, supabaseAdmin as useSupabaseAdmin } from '../utils/supabase.ts'
 import { getEnv } from '../utils/utils.ts'
+import { getPasswordComplianceBodyLogMetadata, getPasswordComplianceSuccessLogMetadata } from './validate_password_compliance_logging.ts'
 
 interface ValidatePasswordCompliance {
   email: string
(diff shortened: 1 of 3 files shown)


mingisrookie (commented):
There is still a pre-redaction validation path that can echo raw request values. The new metadata log only runs after `safeParseSchema(bodySchema, rawBody)` succeeds, but on schema failure the route throws:

`simpleError('invalid_body', 'Invalid request body', { errors: validationResult.error.message })`

ArkType's messages include the rejected value for several fields. I checked locally with the same schema: an invalid org id produces `org_id must be a UUID (was "not-a-uuid")`, and an invalid email produces `email must be an email address (was "not-an-email-secret")`. That means malformed password-compliance requests can still put raw email/org identifiers into `moreInfo` before `getPasswordComplianceBodyLogMetadata()` is ever used.

I would replace this error payload with metadata only (for example `getPasswordComplianceBodyLogMetadata(rawBody)` plus issue paths/codes/counts), and add a test for invalid email/org_id that asserts the serialized error cause/logs do not contain the raw submitted values.

WcaleNieWolny:

PR #2150, opened 2026-05-11
fix(db): bind public permission oracle subjects

hieu-lee:
## Summary

- Bind public `check_min_rights(min_right, user_id, ...)` calls to the request JWT subject or to the user resolved from the request `capgkey`
- Keep the temporary anonymous `check_min_rights` grant used by CLI/RLS app-list compatibility, including RBAC-managed API-key flows where `user_id` is intentionally `NULL`
- Remove public execute from internal subject-taking helpers: `check_min_rights_legacy`, `rbac_check_permission_direct`, and `rbac_check_permission_direct_no_password_policy`
- Keep new subject-binding denial logs metadata-only with presence/match booleans instead of raw user/API-key subject UUIDs
- Extend the pgTAP oracle/RPC permission coverage for direct-helper grants, cross-subject denial, self checks, and capgkey-bound checks

/claim #1667

## Motivation

The public SQL surface still exposed subject-taking permission helpers. An anonymous caller could ask whether an arbitrary known user UUID has a right on an org/app/channel through `check_min_rights`, and an authenticated caller had direct execute on RBAC helper functions that accept an arbitrary `p_user_id`. Those helpers should remain available to trusted internal code and self-bound wrappers, but public calls should not act as membership or permission oracles for arbitrary subjects.

## Test Plan

- [x] `git diff --check`
- [x] Static pgTAP plan check: `49_test_apikey_oracle_rpc_permissions.sql` has 42 planned assertions and 42 assertion calls
- [x] `npx eslint supabase/functions/_backend/public/organization/audit.ts`
- [ ] Supabase pgTAP not run locally because this laptop cannot run Docker

## Checklist

- [x] Public subject-taking permission checks are bound to JWT/API-key identity
- [x] Direct internal RBAC helpers are restricted away from public roles
- [x] Mixed app/channel/org scopes are denied before policy enforcement
- [x] New denial diagnostics avoid retaining raw subject UUIDs

### DIFF
--- a/supabase/functions/_backend/public/organization/audit.ts
+++ b/supabase/functions/_backend/public/organization/audit.ts
@@ -3,7 +3,7 @@ import type { AuthInfo } from '../../utils/hono.ts'
 import { type } from 'arktype'
 import { safeParseSchema } from '../../utils/ark_validation.ts'
 import { quickError, simpleError } from '../../utils/hono.ts'
-import { apikeyHasOrgRightWithPolicy, hasOrgRightApikey, supabaseWithAuth } from '../../utils/supabase.ts'
+import { apikeyHasOrgRightWithPolicy, hasOrgRight, hasOrgRightApikey, supabaseAdmin, supabaseWithAuth } from '../../utils/supabase.ts'
 
 const bodySchema = type({
   'orgId': 'string',
@@ -40,14 +40,14 @@ export async function getAuditLogs(c: Context, bodyRaw: any): Promise<Response> 
     throw simpleError('not_authorized', 'Not authorized')
   }
 
-  const supabase = supabaseWithAuth(c, auth)
+  const authSupabase = supabaseWithAuth(c, auth)
   if (auth.authType === 'apikey') {
     if (!auth.apikey) {
       throw simpleError('not_authorized', 'Not authorized')
     }
 
     // Enforce org scoping + API key policy (expiration) before checking user rights.
-    const orgCheck = await apikeyHasOrgRightWithPolicy(c, auth.apikey, body.orgId, supabase)
+    const orgCheck = await apikeyHasOrgRightWithPolicy(c, auth.apikey, body.orgId, authSupabase)
     if (!orgCheck.valid) {
       if (orgCheck.error === 'org_requires_expiring_key') {
         throw quickError(401, 'org_requires_expiring_key', 'This organization requires API keys with an expiration date. Please use a different key or update this key with an expiration date.')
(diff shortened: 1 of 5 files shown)


SpeedyArt (commented):
One residual logging concern on this security fix: the new public subject-binding denials still persist raw stable identifiers in `pg_log` payloads (`user_id`, `auth_uid`, and `apikey_user_id` on the mismatch branches).

Because `check_min_rights` remains callable by public roles, a mismatched request can now cause retained logs to contain both the requested user subject and the caller/API-key subject. For the same #1667-style cleanup that is moving user/org IDs to booleans/counts elsewhere, I would keep these diagnostics metadata-only: for example `hasUserId`, `hasAuthUid`, `hasApikeySubject`, `subjectMatchesAuth`, `subjectMatchesApikey`, plus the deny reason, rather than storing the raw UUIDs.

That should preserve the useful debugging signal for the oracle-binding fix without reintroducing linkable user identifiers into the retained denial logs.

hieu-lee:
Updated in `2d6685c`.

Changes made:
- validate a supplied `capgkey` against `user_id` even when the JWT subject already matches `user_id`, so a request cannot pair its own JWT with another user API key;
- reject mixed channel/app/org scope before the channel can override the effective org/app context;
- explicitly revoke direct RBAC helpers from `anon` as well as `authenticated`;
- added pgTAP coverage for anon helper privileges, mismatched JWT/API-key subjects, and mixed app/channel scope.

Local checks: `git diff --check`, pgTAP assertion count matches `plan(42)`, and `npx eslint supabase/functions/_backend/public/organization/audit.ts` passes. I still cannot run the Docker-backed Supabase test suite locally in this environment, so CI is the runtime validation.

hieu-lee:
Execution-model note for `check_min_rights`:

- Runs inside Postgres as a `SECURITY DEFINER` RLS authorization helper, reached from public table policies and API/RPC permission gates for `anon` and `authenticated` callers; `service_role` keeps access to the direct helpers only.
- Expected frequency is per protected row/policy evaluation, so the fast path must stay as point lookups. The added work is bounded by provided identifiers rather than scans.
- App scope uses `apps.app_id = $app_id`, backed by `apps_pkey` on `apps(app_id)`. Channel scope uses `channels.id = $channel_id`, backed by the channels primary-key identity index. Org 2FA/RBAC flag checks use `orgs.id = $org_id`, backed by `orgs_pkey`.
- API-key subject proof uses `find_apikey_by_value`, backed by `apikeys_key_idx` for plain keys and `idx_apikeys_key_hash` for hashed keys. Password policy compliance uses the unique/indexed `(user_id, org_id)` path on `user_password_compliance`.
- Worst-cost request shape is an RBAC-enabled app+channel invocation with `user_id` set, `capgkey` present, org 2FA/password policy enabled, and channel overrides possible: it exercises app lookup, channel lookup, API-key lookup, org flag lookup, MFA/password-policy checks, then `rbac_check_permission_direct` and its role-binding/override indexed paths.

I cannot attach truthful `EXPLAIN (ANALYZE, BUFFERS)` output from this environment because the repo test database is Docker-backed and Docker is unavailable here, and I do not have a linked project database credential to run the plans remotely. The runtime validation is deferred to CI; local validation completed was `git diff --check`, pgTAP assertion count matching `plan(42)`, and eslint on the touched backend TypeScript file.

PR #2151, opened 2026-05-11
fix(api): redact Discord error alerts

HappyReni:
## Summary
- replace raw request body, headers, URL, client IP/user-agent, error message, stack, and full error object in Discord 500 alerts with presence/count metadata
- stop logging serialized Discord payload content when DISCORD_ALERT is not configured
- add unit coverage that sensitive body/header/URL/error values are absent from alert payloads

/claim #1667

## Test plan
- ./node_modules/.bin/vitest run tests/discord-alert-redaction.unit.test.ts
- ./node_modules/.bin/vitest run tests/on-error-posthog.unit.test.ts tests/discord-alert-redaction.unit.test.ts
- ./node_modules/.bin/eslint supabase/functions/_backend/utils/discord.ts tests/discord-alert-redaction.unit.test.ts
- git diff --check

### DIFF
--- a/supabase/functions/_backend/utils/discord.ts
+++ b/supabase/functions/_backend/utils/discord.ts
@@ -5,69 +5,143 @@ import type { Context } from 'hono'
 import { cloudlog, cloudlogErr } from './logging.ts'
 import { getEnv } from './utils.ts'
 
-// Fields that should be completely removed from logs (never logged)
-const REMOVED_FIELDS = ['password']
-// Fields that should show first 4 and last 4 characters
-const PARTIALLY_REDACTED_FIELDS = ['secret', 'token', 'apikey', 'api_key', 'authorization', 'credential', 'private_key']
-
-// Partially redact a value - show first 4 and last 4 characters
-function partialRedact(value: string): string {
-  if (value.length <= 8) {
-    return '***REDACTED***'
-  }
-  return `${value.slice(0, 4)}...${value.slice(-4)}`
+interface DiscordAlert500PayloadInput {
+  body: string
+  environment: string
+  error: Error
+  functionName: string
+  hasClientIp: boolean
+  hasUserAgent: boolean
+  method: string
+  rawHeaders: Record<string, string>
+  requestId: string
+  timestamp: string
+  url: string
 }
 
-// Remove or redact sensitive fields from a string that might contain JSON
-function sanitizeSensitiveFromString(str: string): string {
-  let result = str
-
-  // Completely remove password fields (including the key)
-  for (const field of REMOVED_FIELDS) {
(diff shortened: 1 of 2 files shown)


SpeedyArt (commented):
One remaining path still looks raw enough to leak the Discord webhook secret if delivery fails.

`sendDiscordAlert()` now keeps the disabled-webhook log metadata-only, which is good, but the `catch` branch still calls `cloudlogErr({ ..., error })` with the original fetch error object. In Deno/Supabase-style fetch failures, the error message or cause can include the request URL; for Discord webhooks that URL contains the webhook token. So an outage, DNS/TLS failure, bad proxy response, or malformed webhook value could move the secret from env into retained logs.

I would summarize this branch the same way as the alert payload: log a bounded error name/message length or `hasError`/`errorName`, plus maybe the webhook host/path-shape if needed, but do not pass the raw `Error` object through `cloudlogErr`. A regression can mock `fetch` throwing an error whose message contains a fake webhook URL/token and assert the serialized log does not retain it.

HappyReni:
Addressed the Discord webhook error-log path in `47df7d854` and tightened the regression in `75e991dad`: `sendDiscordAlert()` now logs metadata-only error details when fetch throws, and the test covers a thrown error containing a fake Discord webhook URL/token to ensure it is not retained.

supabase/functions/_backend/utils/discord.ts
```diff
@@ -5,69 +5,152 @@ import type { Context } from 'hono'
...
 
-  // Partially redact other sensitive fields (show first 4 and last 4 chars)
-  for (const field of PARTIALLY_REDACTED_FIELDS) {
-    const jsonRegex = new RegExp(`("${field}"\\s*:\\s*)"([^"]*)"`, 'gi')
-    result = result.replace(jsonRegex, (_match, prefix, value) => {
-      return `${prefix}"${partialRedact(value)}"`
-    })
+function getSafeErrorName(error: Error) {
+  const errorName = error?.name ?? 'Error'
+  if (/^[\w .:-]{1,80}$/.test(errorName))
```
mingisrookie, line 66: This still leaves a raw error string path through `error.name`. The regex treats word/space punctuation as safe, so values like `capg_super_secret_key`, `token abc123`, or `Error: token` pass and are emitted in the Discord payload description (and in `getErrorLogMetadata`).

Because this PR is trying to avoid retaining raw request/error details, I would either reduce this to a generic error type/boolean metadata or only allow known non-sensitive class-name values, and add a regression where a sensitive `error.name` is not serialized.

mingisrookie (commented):
One remaining redaction gap: the error message/stack are summarized, but `error.name` can still carry raw sensitive text.

WcaleNieWolny:

PR #2152, opened 2026-05-11
fix(api): redact ok request body logs

Petoverflow:
## Summary (AI generated)

- Replace `/ok` POST request-body cloudlog output with a metadata-only summary.
- Keep the endpoint response/version behavior unchanged.
- Add unit coverage that verifies raw request values and field names are not retained in the log summary.

## Motivation (AI generated)

The public `/ok` endpoint accepts arbitrary POST bodies, so routine logs should not store raw payload data that may contain credentials, tokens, emails, or other accidental sensitive values.

This reduces unnecessary data retention in public endpoint logs while preserving existing health-check/API-version behavior.

## Test Plan (AI generated)

- [x] `npm exec --yes --package=bun -- bunx vitest run tests/ok-request-logging.unit.test.ts`
- [x] `npm exec --yes --package=bun -- bun lint:backend`
- [x] `npm exec --yes --package=bun -- bun typecheck`

Related to #1667

Generated with AI

### DIFF
--- a/supabase/functions/_backend/public/ok.ts
+++ b/supabase/functions/_backend/public/ok.ts
@@ -4,9 +4,39 @@ import { cloudlog } from '../utils/logging.ts'
 
 export const app = honoFactory.createApp()
 
+export function summarizeOkRequestBodyForLog(body: unknown) {
+  if (body === null || body === undefined) {
+    return {
+      bodyType: body === null ? 'null' : 'undefined',
+      hasBody: false,
+    }
+  }
+
+  if (Array.isArray(body)) {
+    return {
+      bodyType: 'array',
+      hasBody: true,
+      itemCount: body.length,
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


subhajitlucky:
One remaining concern: this still fully parses the public `/ok` POST body just to produce log metadata. `parseBody()` uses `c.req.raw.clone().json()`, so a large attacker-controlled body is still buffered and JSON-parsed even though the endpoint does not otherwise need the payload.

For this endpoint, consider avoiding body parsing entirely and logging metadata from headers instead, for example `content-type`, parsed `content-length` when present, and a `hasBody` boolean. That keeps the redaction goal while also removing the unnecessary allocation/parse path from a public health/version endpoint. A regression test could pass a large body and assert the log summary is produced without calling `parseBody()` / without retaining values.

WcaleNieWolny:

PR #2155, opened 2026-05-11
Reduce organization store log exposure

ocepkejepintu-droid:
## Summary
- replace raw member, organization, app, and user identifiers in organization-store console diagnostics with metadata booleans
- keep the existing error/role troubleshooting signal without printing full frontend objects
- leave organization lookup and deletion behavior unchanged

## Security impact
The organization store can currently print member fallback keys, organization IDs, app ownership objects, full organization objects, and user IDs to the browser console during signed-image failures, app ownership mismatches, and delete checks. This change keeps those identifiers out of console logs while preserving enough metadata to diagnose whether values were present.

Linked to #1667.

## Verification
- `bunx eslint src/stores/organization.ts`
- `bun typecheck`
- `git diff --check`

### DIFF
--- a/src/stores/organization.ts
+++ b/src/stores/organization.ts
@@ -218,7 +218,7 @@ export const useOrganizationStore = defineStore('organization', () => {
         return signedImageUrl ? [source.key, signedImageUrl] as const : null
       }
       catch (error) {
-        console.warn('Cannot load signed member image', { memberKey: source.key, error })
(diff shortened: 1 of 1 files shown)


janishahn (commented):
Reviewed the one-file diff and I do not see a blocker. The raw member/org/app/user identifiers in the touched diagnostics are replaced with presence booleans, while the remaining role/error context does not reintroduce the IDs this PR is trying to avoid logging. CI is green on the current head.

WcaleNieWolny:

PR #2160, opened 2026-05-11
fix(frontend): require secure scan update URLs

antaloaalonso:
## Summary
- normalize scanned and manually entered update URLs before opening or downloading
- require HTTPS update URLs, with localhost HTTP kept for local development
- reject non-web schemes, protocol-relative targets, credentialed URLs, and backslash-normalized inputs
- add unit coverage for accepted and rejected update URL inputs

## Tests
- `npx --yes bun@latest x vitest run tests/update-urls.unit.test.ts`
- `npx --yes bun@latest x eslint src/services/updateUrls.ts src/pages/scan.vue tests/update-urls.unit.test.ts`
- `npx --yes bun@latest run lint`
- `npx --yes bun@latest run typecheck`
- `git diff --check`

/claim #1667

### DIFF
--- a/src/pages/scan.vue
+++ b/src/pages/scan.vue
@@ -160,7 +160,7 @@ async function submitManualUrl() {
 
   if (!isNativePlatform) {
     toast.success(`Opening ${new URL(updateUrl).host}`)
-    window.location.assign(updateUrl)
+    globalThis.location.assign(updateUrl)
     return
   }
 
--- a/src/pages/scan.vue
+++ b/src/pages/scan.vue
@@ -12,6 +12,7 @@ import IconArrowPath from '~icons/heroicons/arrow-path-20-solid'
 import IconLink from '~icons/heroicons/link-20-solid'
 import IconQrCode from '~icons/heroicons/qr-code-20-solid'
 import IconShieldCheck from '~icons/heroicons/shield-check-20-solid'
+import { normalizeUpdateUrl } from '~/services/updateUrls'
 import { useDisplayStore } from '~/stores/display'
 
 const router = useRouter()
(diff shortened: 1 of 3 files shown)


janishahn (commented):
Small bug in the localhost exception: `URL.hostname` includes brackets for IPv6 literals, so `new URL("http://[::1]:5173/bundle.zip").hostname` is `[::1]`, but `localHttpHosts` contains `::1`. That means `http://[::1]:...` is rejected even though the code appears to intend to allow IPv6 loopback HTTP for local development. Adding `[::1]` to the set or normalizing the brackets, plus a unit test, would make the exception match the intended behavior.

antaloaalonso:
Thanks, good catch. Addressed in 344e9dc2c by normalizing bracketed IPv6 hostnames before the localhost allowlist check and adding coverage for http://[::1]:5173/bundle.zip. Re-ran targeted Vitest, targeted eslint, full lint, typecheck, and git diff --check locally.

janishahn (commented):
Confirmed the latest diff addresses my earlier review: `isLocalHttpHost` now strips IPv6 brackets before checking the local-host allowlist, and the test covers `http://[::1]:5173/bundle.zip`. I do not see another blocker in this URL-normalization change.

Mykidzou (commented):
I found one validation regression worth covering before merge: `normalizeUpdateUrl` treats any leading token followed by `:` as an explicit scheme. That means valid schemeless host+port inputs are rejected instead of being normalized to HTTPS.

Concrete case:

```ts
normalizeUpdateUrl('updates.example.com:8443/bundle.zip') // returns '' today
```

That happens because `explicitSchemePattern` matches `updates.example.com:` and `new URL(...)` parses it as the custom protocol `updates.example.com:`. This also affects `localhost:5173/bundle.zip`, even though the helper otherwise allows local HTTP development URLs when the scheme is explicit.

Suggested fix: only treat `http:`/`https:` URLs as explicitly schemed for this helper, or require `://` before skipping the `https://` prefix, then add coverage for `updates.example.com:8443/...` and `localhost:5173/...`. The rest of the tightening looks directionally good.

antaloaalonso:
Thanks, good catch. Addressed in ea634abcf by treating only http:// and https:// as explicit web URL schemes before parsing; schemeless host:port values now get the HTTPS default instead of being misread as custom protocols. Added coverage for updates.example.com:8443/bundle.zip and localhost:5173/bundle.zip, while keeping explicit local HTTP allowed.

Re-ran locally:
- npx --yes bun@latest x vitest run tests/update-urls.unit.test.ts
- npx --yes bun@latest x eslint src/services/updateUrls.ts tests/update-urls.unit.test.ts
- npx --yes bun@latest run lint
- npx --yes bun@latest run typecheck
- git diff --check

jooj211 (approved):
Approved. I verified the latest URL normalization keeps HTTPS as the default, allows explicit local HTTP including bracketed IPv6 localhost, rejects non-web schemes/protocol-relative/credentialed/backslash-normalized inputs, and preserves valid schemeless host:port update URLs.

Checks run locally:
- bun x vitest run tests/update-urls.unit.test.ts
- bun x eslint src/services/updateUrls.ts src/pages/scan.vue tests/update-urls.unit.test.ts
- bun -e edge-case smoke test for host:port, IPv6 localhost, non-web schemes, credentials, protocol-relative, and backslash inputs
- git diff --check HEAD~4..HEAD

WcaleNieWolny:

PR #2169, opened 2026-05-11
fix(api): redact failed version delete logs

nickita-khylkouski:
## Summary
- stop logging the raw delete_failed_version request body
- return metadata-only validation details for missing request fields
- add focused coverage for the body metadata helper

## Tests
- bunx vitest run tests/delete-failed-version-redaction.unit.test.ts
- bunx eslint supabase/functions/_backend/private/delete_failed_version.ts tests/delete-failed-version-redaction.unit.test.ts
- git diff --check

/claim #1667

### DIFF
--- a/supabase/functions/_backend/private/delete_failed_version.ts
+++ b/supabase/functions/_backend/private/delete_failed_version.ts
@@ -13,11 +13,19 @@ interface DataUpload {
   name: string
 }
 
+export function getDeleteFailedVersionBodyMetadata(body: Partial<DataUpload>) {
+  return {
+    hasAppId: typeof body.app_id === 'string' && body.app_id.length > 0,
+    hasName: typeof body.name === 'string' && body.name.length > 0,
+  }
+}
+
 export const app = new Hono<MiddlewareKeyVariables>()
 
 app.delete('/', middlewareKey(['all', 'write', 'upload']), async (c) => {
   const body = await parseBody<DataUpload>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'delete failed version body', body })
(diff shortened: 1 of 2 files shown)


jooj211 (approved):
Reviewed the diff and checked the runtime surface for this endpoint. The change replaces the raw delete-failed-version body log and validation error details with presence-only metadata, while still passing the original app_id/name values through to the existing Supabase delete path after validation. The helper keeps the log payload free of the app id and bundle/version name, and the error responses no longer echo the submitted body.

Local verification on this branch:
- `bun x vitest run tests/delete-failed-version-redaction.unit.test.ts` -> 1 file / 2 tests passed
- `bun x eslint supabase/functions/_backend/private/delete_failed_version.ts tests/delete-failed-version-redaction.unit.test.ts` -> no findings
- `git diff --check origin/main...HEAD` -> clean

No blocker from me.

mingisrookie (commented):
One edge case in the new missing-field redaction path: the `!body.app_id` / `!body.name` checks still run after `checkPermission(c, 'bundle.delete', { appId: body.app_id })` and after the `apps.eq('app_id', body.app_id).single()` lookup.

So a request that is actually missing `app_id` may never reach the new `error_app_id_missing` response with `bodyMetadata`; it can be rejected earlier by the permission check or app lookup instead. Since this PR specifically changes the missing-field error payloads, it would be safer to move the presence validation immediately after parsing/logging the metadata, before using `body.app_id` for permission or database filters. A route-level test for `{ name: '...' }` would catch this.

WcaleNieWolny:

PR #2173, opened 2026-05-11
fix(api): redact create app validation body

nickita-khylkouski:
## Summary
- replace raw create-app request bodies in missing-field errors with field-presence metadata
- keep existing validation error codes and statuses unchanged
- add focused unit coverage for helper and thrown error metadata redaction

/claim #1667

## Tests
- `bunx vitest run tests/create-app-redaction.unit.test.ts`
- `bunx eslint supabase/functions/_backend/public/app/post.ts tests/create-app-redaction.unit.test.ts`
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/supabase/functions/_backend/public/app/post.ts
+++ b/supabase/functions/_backend/public/app/post.ts
@@ -17,18 +17,45 @@ export interface CreateApp {
   android_store_url?: string
 }
 
+const createAppMetadataFields = [
+  'android_store_url',
+  'app_id',
+  'existing_app',
+  'icon',
+  'ios_store_url',
+  'name',
+  'need_onboarding',
+  'owner_org',
+] as const
+
+export function getCreateAppBodyMetadata(body: unknown) {
+  if (!body || typeof body !== 'object') {
+    return { hasBody: false }
+  }
+
+  const bodyRecord = body as Record<string, unknown>
+  const presentFields = createAppMetadataFields.filter(field => field in bodyRecord)
+
+  return {
+    fieldCount: Object.keys(bodyRecord).length,
+    hasBody: true,
+    presentFields,
+    unknownFieldCount: Math.max(0, Object.keys(bodyRecord).length - presentFields.length),
+  }
+}
+
 export async function post(c: Context<MiddlewareKeyVariables>, body: CreateApp): Promise<Response> {
   if (!body.app_id) {
-    throw simpleError('missing_app_id', 'Missing app_id', { body })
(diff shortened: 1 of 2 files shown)


joelhagvall:
Small review note: the missing-field cases are redacted now, but `invalid_app_id` still returns
  the raw submitted value:

  ```ts
  throw simpleError('invalid_app_id', 'App ID must be a reverse domain string', { app_id:
  body.app_id })
  ```

  That seems to leave the same class of leak open for malformed `app_id` inputs. For example, if a
  caller accidentally sends a token, email, internal slug, or full URL in `app_id`, the value fails
  validation and is still copied into `moreInfo`.

  Since this PR's goal is to avoid exposing create-app request identifiers/URLs in validation
  errors, I think this branch should use the same metadata-only pattern, or at least omit the raw
  `app_id`.

  Suggested test case:

  ```ts
  const cause = await getRejectedCause(() => post({} as any, {
    app_id: 'https://internal.example/secret-token',
    name: 'Secret App',
    owner_org: 'secret-org-id',
  } as any))

  expect(cause.error).toBe('invalid_app_id')
  expect(JSON.stringify(cause.moreInfo)).not.toContain('secret-token')
  expect(JSON.stringify(cause.moreInfo)).not.toContain('internal.example')

```

  I ran the focused PR test locally after installing deps: `bunx vitest run tests/create-app-
  redaction.unit.test.ts` passed.

jooj211 (approved):
Approved. I verified the create-app missing-field errors now include only body metadata instead of raw request body values, while keeping the validation error codes and statuses unchanged.

Checks run locally:
- bun x vitest run tests/create-app-redaction.unit.test.ts
- bun x eslint supabase/functions/_backend/public/app/post.ts tests/create-app-redaction.unit.test.ts
- git diff --check HEAD^..HEAD

mingisrookie (approved):
Reviewed current head `6c27176` after the invalid-app-id follow-up. The validation branches now route missing/invalid create-app input through `getCreateAppBodyMetadata()` instead of echoing raw `app_id`, `owner_org`, store URLs, or unknown request values. The added regression specifically covers an invalid `app_id` containing `https://internal.example/secret-token` and asserts neither the host nor token leaks in `moreInfo`.

- `bun x eslint supabase/functions/_backend/public/app/post.ts tests/create-app-redaction.unit.test.ts`
- `bun x vitest run tests/create-app-redaction.unit.test.ts` → 4 passed

No remaining touched-file blocker from me.

WcaleNieWolny:

PR #2180, opened 2026-05-11
fix(api): redact admin stats unauthorized logs

jooj211:
## Summary
- Replace the unauthorized admin_stats body log with a whitelisted request context
- Keep caller-supplied extra JSON fields out of logs while preserving metric/date/filter context
- Add a unit test covering extra-field redaction after schema validation

## Verification
- bun x vitest run tests/admin-stats.unit.test.ts
- bun x eslint supabase/functions/_backend/private/admin_stats.ts tests/admin-stats.unit.test.ts

Related to #1667

### DIFF
--- a/supabase/functions/_backend/private/admin_stats.ts
+++ b/supabase/functions/_backend/private/admin_stats.ts
@@ -91,6 +91,25 @@ interface AdminStatsBody {
   offset?: number
 }
 
+export function buildAdminStatsLogContext(body: Partial<AdminStatsBody>) {
+  const context: Partial<AdminStatsBody> = {
+    metric_category: body.metric_category,
+    start_date: body.start_date,
+    end_date: body.end_date,
+  }
+
+  if (body.app_id)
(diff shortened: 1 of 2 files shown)


cwccty:

What I checked:

1. The unauthorized `not_admin` log no longer receives the raw request body, and now uses `buildAdminStatsLogContext(parsedBodyResult.data)`.
2. That context is built from the validated schema result, so caller supplied extra JSON fields are not forwarded into the log context on the valid body unauthorized path.
3. The added regression test covers this by passing `token` and `authorization` as extra fields and asserting they are absent from the log context.

One small optional follow-up: depending on the logging policy for admin-only endpoints, you may also want to decide whether `app_id` and `org_id` should stay as raw values in unauthorized logs or be reduced to presence flags. I do not see that as a blocker for this PR.

I could not run the test commands locally because this environment does not have `bun` installed, so this review is based on static diff inspection.

WcaleNieWolny:

PR #2181, opened 2026-05-11
fix(frontend): reject unsafe dialog links

antaloaalonso:
/claim #1667

## Summary
- add a dialog-link URL safety check that allows HTTPS, relative app paths, mailto/tel, and localhost HTTP only
- prevent unsafe dialog hrefs from being rendered or followed by the programmatic dialog close path
- cover unsafe javascript/data/protocol-relative URLs and safe Stripe-style blank-target navigation

## Verification
- npx --yes bun@latest x vitest run tests/dialogv2-navigation.test.ts
- npx --yes bun@latest x eslint src/stores/dialogv2.ts src/components/DialogV2.vue tests/dialogv2-navigation.test.ts --ext .ts,.vue
- git diff --check
- npx --yes bun@latest run lint
- npx --yes bun@latest run typecheck
- npx --yes bun@latest run test:all (blocked before tests: local Docker daemon is not running for the Supabase harness)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced dialog link safety with strict href validation, preventing navigation to unsafe or malicious URLs while permitting internal links, whitelisted payment services, and special protocols like mailto and tel.

* **Chores**
  * Improved developer environment compatibility for Node versions below 22 in startup scripts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2181)

### DIFF
--- a/src/stores/dialogv2.ts
+++ b/src/stores/dialogv2.ts
@@ -1,7 +1,11 @@
 import { acceptHMRUpdate, defineStore } from 'pinia'
 import { ref, watch } from 'vue'
 
-const safeDialogHrefProtocols = new Set(['https:', 'mailto:', 'tel:'])
+const safeDialogHrefProtocols = new Set(['mailto:', 'tel:'])
+const safeExternalDialogHostnames = new Set([
+  'billing.stripe.com',
+  'checkout.stripe.com',
+])
 const localHttpHostnames = new Set(['localhost', '127.0.0.1', '::1', '[::1]'])
 const fallbackDialogHrefBase = 'https://app.capgo.app'
 
@@ -15,7 +19,16 @@ export function isSafeDialogHref(href?: string): href is string {
   try {
     const base = globalThis.location?.origin || fallbackDialogHrefBase
     const url = new URL(trimmedHref, base)
-    return safeDialogHrefProtocols.has(url.protocol) || (url.protocol === 'http:' && localHttpHostnames.has(url.hostname))
+    const baseOrigin = new URL(base).origin
+
+    if (safeDialogHrefProtocols.has(url.protocol))
+      return true
+    if (url.protocol === 'http:')
+      return localHttpHostnames.has(url.hostname)
+    if (url.protocol !== 'https:')
+      return false
+
+    return url.origin === baseOrigin || safeExternalDialogHostnames.has(url.hostname)
   }
   catch {
     return false
--- a/src/stores/dialogv2.ts
+++ b/src/stores/dialogv2.ts
@@ -9,6 +9,10 @@ const safeExternalDialogHostnames = new Set([
 const localHttpHostnames = new Set(['localhost', '127.0.0.1', '::1', '[::1]'])
 const fallbackDialogHrefBase = 'https://app.capgo.app'
 
+function isLocalHttpUrl(url: URL) {
+  return url.protocol === 'http:' && localHttpHostnames.has(url.hostname)
+}
+
 export function isSafeDialogHref(href?: string): href is string {
   const trimmedHref = href?.trim()
   if (!trimmedHref)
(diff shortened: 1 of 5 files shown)


antaloaalonso:
Updated this branch to address the CodeRabbit review note: unsafe dialog href clicks now prevent navigation but still close the dialog with the sanitized button payload.

Local verification after the update:
- `npx --yes bun@latest x vitest run tests/dialogv2-navigation.test.ts`
- `npx --yes bun@latest x eslint src/components/DialogV2.vue tests/dialogv2-navigation.test.ts --ext .vue,.ts`
- `git diff --check`
- `npx --yes bun@latest run lint`
- `npx --yes bun@latest run typecheck`
- `npx --yes bun@latest run build`

The `~/stores/dialogv2` dynamic import note is already reflected in the test file.

mingisrookie (commented):
One policy question before treating this as complete hardening: `isSafeDialogHref()` currently allows any `https:` URL, not just same-origin or known external destinations.

That blocks `javascript:`/`data:`/protocol-relative links, which is good, but it still means any dialog button whose `href` is influenced by server data or user-controlled state can navigate to `https://evil.example/...` and the helper will consider it safe. That is a different policy from the redirect helper, which rejects external targets by default.

If dialog buttons are guaranteed to be static developer-authored URLs, this may be fine. If not, I would either restrict to same-origin relative URLs plus a small allowlist (for example Stripe checkout/billing, mailto, tel, local dev), or rename/document the helper as "safe protocol" rather than "safe dialog href" and add a test making the arbitrary-HTTPS decision explicit.

antaloaalonso:
Addressed the dialog href policy question in 67605b76d: arbitrary HTTPS destinations are no longer accepted. Dialog hrefs now allow same-origin HTTPS, local HTTP for dev, mailto/tel, and the specific Stripe checkout/billing hosts used by the current dialog href flow. Added regression coverage for billing.stripe.com and for rejecting https://evil.test.\n\nLocal verification:\n- npx --yes bun@latest x vitest run tests/dialogv2-navigation.test.ts\n- npx --yes bun@latest x eslint src/stores/dialogv2.ts tests/dialogv2-navigation.test.ts --ext .ts\n- npx --yes bun@latest run typecheck\n- git diff --check

src/stores/dialogv2.ts
```diff
@@ -1,6 +1,40 @@
...
+
+  try {
+    const base = globalThis.location?.origin || fallbackDialogHrefBase
+    const url = new URL(trimmedHref, base)
+    const baseOrigin = new URL(base).origin
+
+    if (safeDialogHrefProtocols.has(url.protocol))
+      return true
+    if (url.protocol === 'http:')
+      return localHttpHostnames.has(url.hostname)
```
mingisrookie: One remaining policy mismatch: the PR comment describes local HTTP as dev-only, but the current check allows localhost HTTP regardless of the app origin. With `globalThis.location.origin = 'https://app.capgo.app'`, `isSafeDialogHref('http://localhost:5173/settings')` reaches this branch and returns `true` because only the target hostname is checked.

That means a production dialog href can still navigate a user to a local service. Consider gating the localhost exception on a dev/local app origin (or `import.meta.env.DEV`) and adding a regression test where a production origin rejects `http://localhost:...`.

mingisrookie (commented):
Follow-up after the allowlist update: arbitrary HTTPS is fixed, but the localhost HTTP exception still appears to be enabled in production contexts.

antaloaalonso:
Addressed the localhost production-context concern in ca0b472cf: local HTTP dialog hrefs now require Vite dev mode in addition to a local app origin, so production builds reject them.

Local verification after the update:
- `npx --yes bun@latest x vitest run tests/dialogv2-navigation.test.ts`
- `npx --yes bun@latest x eslint src/stores/dialogv2.ts tests/dialogv2-navigation.test.ts --ext .ts`
- `git diff --check`
- `npx --yes bun@latest run typecheck`
- `npx --yes bun@latest run lint`
- `npx --yes bun@latest run build`

antaloaalonso:
Addressed the latest Wrangler fallback review in df2478282: the Node <22 path now verifies the local Wrangler binary before using it, falls back to a global `wrangler` when available, and otherwise exits with an explicit setup error instead of failing later during worker startup.

Focused verification:
- `bash -n scripts/start-cloudflare-workers.sh`
- `npx --yes bun@latest x node@24 ./node_modules/wrangler/bin/wrangler.js --version`
- simulated missing-local-wrangler branch exits explicitly
- `npx --yes bun@latest x vitest run tests/dialogv2-navigation.test.ts`
- `git diff --check`

mingisrookie (approved):
Rechecked the current head (`df2478282`): the Node <22 Wrangler path now verifies the local binary first, falls back to a global `wrangler` when available, and otherwise exits with a clear setup error. I also ran `bash -n scripts/start-cloudflare-workers.sh` and it parsed cleanly.

WcaleNieWolny:

PR #2183, opened 2026-05-11
fix(api): redact invitation accept token logs

nickita-khylkouski:
## Summary

- Stop logging `magic_invite_string` in the accept-invitation raw and validated body logs.
- Keep a boolean `has_magic_invite_string` marker so the logs retain request-shape context without storing the bearer token.
- Add a focused unit test for the lookup-failure and validated-body paths.

## Root cause

The accept-invitation handler removed `password` and `captchaToken` before logging request bodies, but left `magic_invite_string`. That invite string is the bearer credential from the invite link and may still be valid when validation or account creation fails.

## Validation

- `bunx vitest run tests/accept-invitation-redaction.unit.test.ts`
- `bun x eslint supabase/functions/_backend/private/accept_invitation.ts tests/accept-invitation-redaction.unit.test.ts`
- `git diff --check`

### DIFF
--- a/supabase/functions/_backend/private/accept_invitation.ts
+++ b/supabase/functions/_backend/private/accept_invitation.ts
@@ -272,8 +272,8 @@ app.post('/', async (c) => {
   }
 
   const baseBody = baseValidationResult.data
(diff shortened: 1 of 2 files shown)


WcaleNieWolny:

PR #2184, opened 2026-05-11
fix(api): hide org member delete enumeration

nickita-khylkouski:
## Summary
- Return the same sanitized 404 when deleting a nonexistent user email and an existing nonmember
- Stop echoing target email or raw lookup errors in the not-member response
- Add unit coverage for the collapsed response shape and update the organization API expectation

## Testing
- bunx vitest run tests/organization-member-delete-enumeration.unit.test.ts
- bun x eslint supabase/functions/_backend/public/organization/members/delete.ts tests/organization-member-delete-enumeration.unit.test.ts tests/organization-api.test.ts
- git diff --check

### DIFF
--- a/supabase/functions/_backend/public/organization/members/delete.ts
+++ b/supabase/functions/_backend/public/organization/members/delete.ts
@@ -13,6 +13,10 @@ const deleteBodySchema = type({
   email: 'string.email',
 })
 
+function memberNotFoundError(orgId: string): never {
+  throw quickError(404, 'organization_member_not_found', 'User is not a member of this organization', { orgId })
+}
+
 export async function deleteMember(c: Context<MiddlewareKeyVariables>, bodyRaw: any, _apikey: Database['public']['Tables']['apikeys']['Row']) {
   const bodyParsed = safeParseSchema(deleteBodySchema, bodyRaw)
   if (!bodyParsed.success) {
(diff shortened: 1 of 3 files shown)


cwccty:

What I checked:

- The missing-user branch and the existing-user-but-not-member branch now collapse to the same `organization_member_not_found` response, which removes the email enumeration signal from the public API.
- The previous `user_not_found` branch no longer returns the raw Supabase lookup error in `moreInfo`.
- The nonmember branch no longer includes the submitted target email in the error payload.
- The temporary `cloudlog` calls for `userData.id` and `body.orgId` were removed, so the delete path no longer adds those identifiers to routine logs before the membership deletion.
- The new unit test covers both relevant branches and asserts that the serialized error data does not contain the submitted emails, the Supabase `PGRST116` detail, the raw "No rows found" message, or the looked-up user id.

One small optional follow-up: if you want integration coverage for the second branch as well, `tests/organization-api.test.ts` currently covers the nonexistent email case after this PR, while the existing-user nonmember case is covered by the new mocked unit test. I do not think this should block the PR because the unit test exercises the branch directly.

I could not run the test commands locally because this environment does not have Bun installed, so this review is based on static diff inspection.

WcaleNieWolny:

PR #2186, opened 2026-05-11
fix: reduce sensitive backend request logging

JB-Bryant:
## Summary

- replace full request-body logging in billing/download/delete-failed-version endpoints with metadata-only summaries
- avoid logging raw Stripe customer IDs from loaded org rows
- preserve request flow and permission checks; this only reduces routine log retention

## Security note

This is a public-safe log hardening change for the security retribution thread. It avoids retaining callback/success/cancel URLs, client reference IDs, attribution IDs, bundle names, and raw customer IDs in routine cloud logs where presence flags or stable IDs are enough for debugging.

Related: #1667
/claim #1667

## Testing

- `git diff --check`
- `bun run lint:backend` could not be run locally because `bun` is not installed in this environment

## Summary by CodeRabbit

* **Chores**
  * Replaced raw request-body logging with structured, presence-flag style logs across several private backend endpoints to reduce exposed data in logs and improve observability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2186)

### DIFF
--- a/supabase/functions/_backend/private/stripe_checkout.ts
+++ b/supabase/functions/_backend/private/stripe_checkout.ts
@@ -23,7 +23,7 @@ app.use('/', useCors)
 
 app.post('/', middlewareAuth, async (c) => {
   const body = await parseBody<CheckoutData>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout request', orgId: body.orgId, recurrence: body.recurrence, hasClientReferenceId: !!body.clientReferenceId, hasAttributionId: !!body.attributionId, hasSuccessUrl: !!body.successUrl, hasCancelUrl: !!body.cancelUrl })
+  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout request', hasOrgId: !!body.orgId, recurrence: body.recurrence, hasClientReferenceId: !!body.clientReferenceId, hasAttributionId: !!body.attributionId, hasSuccessUrl: !!body.successUrl, hasCancelUrl: !!body.cancelUrl })
 
   if (!body.orgId)
     throw simpleError('no_org_id_provided', 'No org_id provided')
--- a/supabase/functions/_backend/private/stripe_checkout.ts
+++ b/supabase/functions/_backend/private/stripe_checkout.ts
@@ -23,7 +23,7 @@ app.use('/', useCors)
 
 app.post('/', middlewareAuth, async (c) => {
   const body = await parseBody<CheckoutData>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout body', body })
+  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout request', orgId: body.orgId, recurrence: body.recurrence, hasClientReferenceId: !!body.clientReferenceId, hasAttributionId: !!body.attributionId, hasSuccessUrl: !!body.successUrl, hasCancelUrl: !!body.cancelUrl })
 
   if (!body.orgId)
     throw simpleError('no_org_id_provided', 'No org_id provided')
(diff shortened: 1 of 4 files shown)


ThiagoCAltoe:
There is still one sensitive logging path left in the two billing endpoints touched here: both `stripe_checkout.ts` and `stripe_portal.ts` keep logging `authContext.userId` with `message: 'auth'` after the request-body/customer logs are redacted. That user id is still an authenticated user identifier tied to billing activity, so the PR reduces some leakage but does not fully remove the sensitive backend request logging for these flows.

Can you drop that `cloudlog({ ..., message: 'auth', auth: authContext.userId })` line or replace it with a non-identifying boolean such as `hasAuthContext: true`?

JB-Bryant:
Thanks, agreed. I removed the two `authContext.userId` cloudlog calls from the billing endpoints and force-pushed the branch.

I also compacted the added metadata logs to avoid the Sonar duplication failure from the first revision. Waiting for checks to rerun now.

ThiagoCAltoe:
Thanks for the quick update. I noticed one similar metadata path still remains in `delete_failed_version.ts`: the `apikey context` log still includes `apikeyId` and `userId`.

Since this PR is removing authenticated-user identifiers from backend request logs, could this log also be reduced to non-identifying metadata, e.g. `hasApikey`, `hasUser`, and `mode`, or dropped if it is no longer needed?

JB-Bryant:
Thanks, good catch. I updated that remaining apikey context log to keep only non-identifying metadata: `hasApikey`, `hasUser`, and `mode`.

mingisrookie (commented):
This still leaves raw request identifiers in the new "redacted" request logs, and several of them happen before the authorization check for that resource.

Examples from the patch:
- `delete_failed_version.ts` logs `app_id: body.app_id` before `checkPermission(..., { appId: body.app_id })`.
- `download_link.ts` logs `app_id: body.app_id` / `bundleId: body.id` before `checkPermission(..., { appId: body.app_id })`.
- `stripe_checkout.ts` and `stripe_portal.ts` log raw `orgId: body.orgId` before `checkPermission(..., { orgId: body.orgId })`.

So the PR removes full-body logging, but an unauthorized caller can still cause arbitrary app/org identifiers to be written to operational logs. If the goal is metadata-only request logging for #1667, these should probably be `hasAppId` / `hasBundleId` / `hasOrgId` (and maybe type/field-count metadata) rather than the raw IDs. A small regression test around unauthorized requests would catch this difference, because the current functional path still uses the identifiers while the logs do not need to retain them.

mingisrookie (approved):
Rechecked the latest commit (`36aab683`): the request-entry logs in the affected endpoints now use presence booleans (`hasAppId`, `hasBundleId`, `hasOrgId`) instead of raw app/org/bundle identifiers. I also ran ESLint on the touched files:

`bun x eslint supabase/functions/_backend/private/delete_failed_version.ts supabase/functions/_backend/private/download_link.ts supabase/functions/_backend/private/stripe_checkout.ts supabase/functions/_backend/private/stripe_portal.ts`

Result: passed with no reported issues.

ThiagoCAltoe (approved):

supabase/functions/_backend/private/stripe_checkout.ts
```diff
@@ -54,7 +53,7 @@ app.post('/', middlewareAuth, async (c) => {
   if (!await checkPermission(c, 'org.update_billing', { orgId: body.orgId }))
     throw simpleError('not_authorize', 'Not authorize')
 
-  cloudlog({ requestId: c.get('requestId'), message: 'user', org })
+  cloudlog({ requestId: c.get('requestId'), message: 'stripe checkout org loaded', hasCustomer: !!org.customer_id })
```
darshan-Jahagirdar, line 56: The checkout/portal body logs are now metadata-only, but the invalid redirect path still includes the raw URL. `createCheckout` passes `body.successUrl`/`body.cancelUrl` into `getAllowedRedirectUrl`; that helper throws `simpleError('invalid_redirect_url', ..., { field, value, expectedOrigin })`, where `value` is the user-provided URL. A rejected URL containing a token in the query string can therefore still land in error metadata. `stripe_portal` has the same issue for `callbackUrl`.

Can you redact that helper too, e.g. log only `field`, `hasValue`, and whether the origin matched, rather than the raw URL?

WcaleNieWolny:

PR #2187, opened 2026-05-11
fix: route app settings updates through API

JB-Bryant:
## Summary

- route app settings edits through `PUT /app/:id` instead of direct browser-side PostgREST updates
- allow JWT auth on the app update endpoint via `middlewareV2`, while preserving API-key/subkey support
- add backend handling for `build_timeout_seconds` and `allow_preview`, including server-side build-timeout validation

## Security note

This keeps app-setting validation and side effects on the backend path. Previously the settings page updated `public.apps` directly from the browser for fields such as retention, build timeout, preview, metadata exposure, and custom device IDs. That made the browser path depend entirely on table RLS and bypassed the existing API validation/side-effect path.

Related: #1667
/claim #1667

## Testing

- `git diff --check`
- Could not run `bun run lint` locally because `bun` is not installed in this environment

### DIFF
--- a/src/components/dashboard/AppSetting.vue
+++ b/src/components/dashboard/AppSetting.vue
@@ -15,7 +15,7 @@ import iconName from '~icons/ph/user?raw'
 import Toggle from '~/components/Toggle.vue'
 import { checkPermissions } from '~/services/permissions'
 import { createSignedImageUrl, getImmediateImageUrl } from '~/services/storage'
-import { useSupabase } from '~/services/supabase'
+import { defaultApiHost, useSupabase } from '~/services/supabase'
 import { useDialogV2Store } from '~/stores/dialogv2'
 
 const props = defineProps<{ appId: string }>()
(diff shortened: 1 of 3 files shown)


mingisrookie (commented):
This is a useful boundary change, but I would add a route-level regression before merging because the important behavior moved from client-side Supabase RLS writes to the `/app/:id` API path.

The patch changes the route from `middlewareKey(['all', 'write'])` to `middlewareV2(['all', 'write'])`, switches the update client from `supabaseApikey(c, apikey.key)` to `supabaseWithAuth(c, auth)`, and adds new writable fields (`build_timeout_seconds`, `allow_preview`). The UI path exercises JWT auth, but API-key/subkey callers are still accepted by `middlewareV2` and now depend on `supabaseWithAuth` preserving the old update semantics.

A small backend test for `PUT /app/:id` with a mocked JWT auth context and one with an API key/subkey auth context would catch regressions where `checkPermission` passes but the Supabase update fails under the new auth client, or where the new fields are silently dropped. Right now the PR has no direct coverage for the route contract it is moving the UI onto.

nagiexplorer88:
One app-settings path still bypasses the new API route: `updateStoreUrls()` continues to write `ios_store_url` and `android_store_url` directly through the browser Supabase client.

Since this PR moves settings updates to `PUT /app/:id` so the backend permission/validation path is used, could the store URL update use `updateAppSettings({ ios_store_url, android_store_url })` as well? The backend `UpdateApp` body already accepts both fields, so leaving this one direct PostgREST update means the settings form still has a parallel browser-side write path after this change.

WcaleNieWolny:

PR #2188, opened 2026-05-11
fix: redact queue request body logs

TATANKA97:
## Summary
- Redact sensitive fields before queue request bodies are written to worker logs.
- Reuse the existing response-body sanitization rules for string values and add recursive request-body sanitization for nested objects/arrays.
- Avoid logging raw queue request URLs; log URL shape metadata instead so query-string secrets are not persisted.
- Add regression coverage for bearer tokens, emails, session keys, refresh tokens, inline token-like metadata, and URL query secrets.

## Why
`queue_consumer` currently logs request details before dispatching internal trigger calls. Queue payloads and destination URLs can contain operational identifiers or secrets such as authorization headers, tokens, session keys, captcha tokens, user emails, callback tokens, or signed query parameters. Those values should not be persisted in logs when the only goal is operational tracing.

This keeps the log useful for routing/debugging while avoiding raw credential or PII exposure.

## Verification
- `git diff --check`
- `bunx vitest run tests/queue-consumer-message-shape.unit.test.ts`
- `bun run lint:backend`
- GitHub Actions: Run tests passed
- GitHub Actions: CodSpeed passed
- SonarQube Cloud: Quality Gate passed

/claim #1667

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data protection: request bodies and alert text are now centrally sanitized before logging, and sensitive query values are omitted from logs.

* **Tests**
  * Added tests to verify sensitive fields (tokens, emails, credentials) are redacted and original inputs are not mutated during queue processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2188)

### DIFF
--- a/supabase/functions/_backend/triggers/queue_consumer.ts
+++ b/supabase/functions/_backend/triggers/queue_consumer.ts
@@ -191,6 +191,24 @@ function sanitizeQueueLogValue(value: unknown): unknown {
   ]))
 }
 
+function getQueueLogUrlMetadata(value: string): Record<string, unknown> {
+  try {
+    const parsed = new URL(value)
+    return {
+      host: parsed.host,
+      pathSegments: parsed.pathname.split('/').filter(Boolean).length,
+      hasQuery: parsed.search.length > 0,
+    }
+  }
+  catch {
+    return {
+      host: null,
+      pathSegments: 0,
+      hasQuery: value.includes('?'),
+    }
+  }
+}
+
 // Helper function to generate UUID v4
 function generateUUID(): string {
   return crypto.randomUUID()
@@ -515,7 +533,8 @@ export async function http_post_helper(
   try {
     cloudlog({
       requestId: c.get('requestId'),
-      message: `[${function_name}] Making HTTP POST request to "${url}" with body:`,
+      message: `[${function_name}] Making HTTP POST request with body:`,
+      url: getQueueLogUrlMetadata(url),
       body: sanitizeQueueLogValue(body),
     })
     const response = await fetch(url, {
@@ -665,5 +684,6 @@ export const __queueConsumerTestUtils__ = {
   extractMessageBody,
   getActionableQueueFailures,
   sanitizeDiscordResponseBody,
+  getQueueLogUrlMetadata,
   sanitizeQueueLogValue,
 }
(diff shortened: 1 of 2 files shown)


cwccty:

What I checked:

1. The queue HTTP POST log no longer writes the raw request body directly; it now passes the body through `sanitizeQueueLogValue` before calling `cloudlog`.
2. The sanitizer handles nested objects and arrays, while redacting key-based secrets such as `authorization`, `session_key`, and `refreshToken`.
3. The string-level sanitizer is reused for queue log values, so email-like substrings, bearer tokens, long token-like values, and inline `token=` metadata are also covered.
4. The regression test keeps the original body intact for the actual request path and verifies that the logged shape no longer contains the sample email or token value.

One small optional follow-up: if queue payloads can include other organization-specific identifiers that should not be retained in logs, those can be handled as a policy follow-up by adding the relevant keys to `isSensitiveLogKey`. I do not see that as a blocker for this PR.

I could not run the test commands locally because this environment does not have `bun` installed, so this review is based on static diff inspection.

mingisrookie (commented):
One sensitive value is still outside the new body sanitizer: the log message interpolates the raw destination URL before the sanitized body is attached.

`http_post_helper()` now does:

`message: [${function_name}] Making HTTP POST request to "${url}" with body:`

So if a queued HTTP job is ever created with credentials or signed state in the query string (for example a webhook/test endpoint, callback token, or pre-signed URL), the body will be redacted but the log message will still retain the raw URL. This is easy to miss because the new unit test only serializes `sanitizeQueueLogValue(body)` and does not exercise the actual `cloudlog` payload from `http_post_helper()`.

I would log URL metadata instead (host present / path segment count / hasQuery, or a sanitized URL with query stripped) and add a test that calls `http_post_helper()` with `https://example.test/hook?token=secret` and asserts the serialized cloudlog calls do not contain `secret`.

mingisrookie (approved):
Thanks for the quick follow-up. I rechecked the new `fix: avoid logging queue request URL secrets` commit: the raw URL is no longer interpolated into the log message, the log now stores URL shape metadata (`host`, path segment count, `hasQuery`), and the new `http_post_helper()` test covers both the URL query token and body token not appearing in serialized `cloudlog` calls while still sending the original request to `fetch`.

This addresses my earlier concern. No blocker from me.

WcaleNieWolny:

PR #2189, opened 2026-05-11
fix: route role binding writes through API

JB-Bryant:
## Summary
- route group role binding create/update/delete operations through the existing private/role_bindings API
- route access table role deletion through the same backend endpoint
- remove direct browser-side writes to the role_bindings table for these RBAC flows

## Why
The backend private/role_bindings routes already centralize RBAC write validation, including permission checks, role scope validation, principal/org validation, limited API key guards, and anti-escalation checks. Keeping browser-side writes on role_bindings can bypass those server-side guards and rely only on table policies.

Related: #1667
/claim #1667

## Testing
- Not run locally: local dependencies are not installed, so ESLint cannot load the repo config in this workspace.

### DIFF
--- a/src/components/tables/AccessTable.vue
+++ b/src/components/tables/AccessTable.vue
@@ -437,11 +437,9 @@ async function deleteElement(element: Element) {
 
   isLoading.value = true
   try {
-    // Delete directly via RLS
-    const { error } = await supabase
-      .from('role_bindings')
-      .delete()
-      .eq('id', element.id)
+    const { error } = await supabase.functions.invoke(`private/role_bindings/${element.id}`, {
+      method: 'DELETE',
+    })
 
     if (error)
       throw error
--- a/src/pages/settings/organization/Groups.[id].vue
+++ b/src/pages/settings/organization/Groups.[id].vue
@@ -387,10 +387,6 @@ function getAppName(appId: string) {
   return app ? (app.name || app.app_id) : appId
 }
 
-function getRoleIdByName(roleName: string) {
-  return roles.value.find((r: Role) => r.name === roleName)?.id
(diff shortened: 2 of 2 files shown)


Flob365 (commented):
Reviewed the route-binding changes and I do not see a blocker. The moved delete/create/update paths now go through `private/role_bindings`, and the backend endpoint still loads the managed binding, validates the target role, checks the caller priority rank, and handles the last-super-admin guard on updates/deletes.

I also checked the latest PR checks: `Run tests`, `Run Playwright tests`, `Check dead code`, benchmarks, and Sonar are all green. The frontend payloads in `Groups.[id].vue` match the backend schema for org/app bindings (`principal_type`, `principal_id`, `role_name`, `scope_type`, `org_id`, `app_id`, `channel_id`, `reason`).

One non-blocking observation: these frontend helpers do not encode `bindingId` in the route, but the values are UUIDs from existing `role_bindings`, so this should be fine in practice.

mingisrookie (approved):
I reviewed the changed role-binding write paths and did not find a blocker. The group edit page now routes create/update/delete through `private/role_bindings`, so role names are resolved server-side and the existing backend checks apply (`validatePrincipalAccess`, scope validation, app ownership for app bindings, and the caller max-priority anti-escalation checks). The app access table delete path now uses the same `DELETE /private/role_bindings/:id` route as the other RBAC surfaces, instead of deleting directly via RLS.

I also ran ESLint on the touched files:

`bun x eslint src/components/tables/AccessTable.vue src/pages/settings/organization/Groups.[id].vue`

Result: passed with no reported issues.

WcaleNieWolny:

PR #2193, opened 2026-05-11
fix(rbac): scope forced-device policies to channels

selimeneserd:
## Summary
- update `channel_devices` RLS to use request-aware channel-scoped forced-device permissions
- require `channel.read_forced_devices` for direct reads and `channel.manage_forced_devices` for direct writes
- sync `supabase/schemas/prod.sql` and add a pgTAP policy regression test

## Test plan
- `git diff --check`
- GitHub Actions `Run tests` passed, including the Supabase pgTAP suite with `supabase/tests/55_test_channel_devices_forced_device_rbac.sql`
- GitHub Actions `Run Playwright tests` passed
- SonarCloud, dead-code, CodSpeed, benchmark, Socket, and CodeRabbit checks passed

I could not run the Supabase pgTAP suite locally because this environment does not have `bun`, `supabase`, or `psql` installed. The first CI run exposed a pgTAP compatibility issue in the new policy test; I fixed it and the latest CI run is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced access control and security for channel device management operations with improved permission validation across the system.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2193)

### DIFF
--- a/supabase/migrations/20260511090000_channel_devices_forced_device_rbac.sql
+++ b/supabase/migrations/20260511090000_channel_devices_forced_device_rbac.sql
@@ -1,0 +1,93 @@
+-- Ensure direct PostgREST access to forced-device overrides uses the
+-- channel-scoped RBAC permissions instead of broad app read/write rights.
+
+DROP POLICY IF EXISTS "Allow delete for auth, api keys (write+)"
+ON public.channel_devices;
+
+DROP POLICY IF EXISTS "Allow insert for auth (write+)"
+ON public.channel_devices;
+
+DROP POLICY IF EXISTS "Allow read for auth, api keys (read+)"
+ON public.channel_devices;
+
+DROP POLICY IF EXISTS "Allow read for auth (read+)"
+ON public.channel_devices;
+
+DROP POLICY IF EXISTS "Allow update for auth, api keys (write+)"
+ON public.channel_devices;
+
+CREATE POLICY "Allow delete for auth, api keys (write+)"
+ON public.channel_devices
+FOR DELETE
+TO anon, authenticated
+USING (
+  public.rbac_check_permission_request(
+    public.rbac_perm_channel_manage_forced_devices(),
+    owner_org,
+    app_id,
+    channel_id
+  )
+);
+
+CREATE POLICY "Allow insert for auth (write+)"
+ON public.channel_devices
+FOR INSERT
+TO authenticated
+WITH CHECK (
+  public.rbac_check_permission_request(
+    public.rbac_perm_channel_manage_forced_devices(),
+    owner_org,
+    app_id,
+    channel_id
+  )
+);
+
+CREATE POLICY "Allow read for auth (read+)"
+ON public.channel_devices
+FOR SELECT
+TO anon, authenticated
+USING (
+  public.rbac_check_permission_request(
+    public.rbac_perm_channel_read_forced_devices(),
+    owner_org,
+    app_id,
+    channel_id
+  )
+);
+
+CREATE POLICY "Allow update for auth, api keys (write+)"
+ON public.channel_devices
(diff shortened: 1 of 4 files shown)


tngus6007:
There is a policy-name mismatch that looks security-relevant here. The migration drops `"Allow read for auth, api keys (read+)"`, but the schema snapshot shows the existing `channel_devices` SELECT policy is named `"Allow read for auth (read+)"`.

If production has the schema-snapshot name, the old broad SELECT policy will not be dropped. PostgreSQL combines permissive RLS policies with OR semantics, so the new `channel.read_forced_devices` policy can be added while the old `check_min_rights('read', get_identity_org_appid(...), ..., NULL::bigint)` policy still grants app-scoped reads. That would leave the direct-read path broader than intended.

I would drop both legacy names in the migration and add a regression assertion that no `channel_devices` policy still references `check_min_rights` after the migration, e.g. checking `pg_policies` for `tablename = 'channel_devices'` and `qual/with_check NOT LIKE '%check_min_rights%'`.

selimeneserd:
Addressed in `b424d9a`.

- The migration now drops both legacy SELECT policy names: `Allow read for auth, api keys (read+)` and `Allow read for auth (read+)`.
- The recreated SELECT policy now uses the schema-snapshot name `Allow read for auth (read+)`, so migrated databases and fresh schema state align.
- Added a pgTAP regression assertion that no `channel_devices` RLS policy retains `check_min_rights` in `qual` or `with_check`.

`git diff --check` passes locally; waiting for the refreshed CI run.

digzrow-coder:
I think this still needs a regression that the row's `app_id` / `owner_org` must match the submitted `channel_id` before the policy grants access.

The new `channel_devices` policies pass the row values into `rbac_check_permission_request(..., owner_org, app_id, channel_id)`, but `rbac_check_permission_direct` currently resolves `p_channel_id` back to `channels.owner_org/app_id` and uses that as the effective permission scope. Since `channel_devices` only has independent foreign keys to `apps(app_id)` and `channels(id)`, I don't see a constraint proving those row values describe the same channel.

That means a mismatched row can satisfy `channel.manage_forced_devices` for channel A while writing a `channel_devices` record whose `app_id` points at app B. `auto_owner_org_by_app_id` only normalizes `owner_org` from `app_id`; it doesn't prove the channel belongs to that app. I'd add either a composite FK/trigger enforcing `channel_devices.(app_id, owner_org)` matches `channels.(app_id, owner_org)`, or an explicit helper/policy check plus a pgTAP regression for mismatched channel/app rows.

KCDaemon (changes requested):
Rechecked the latest head (586f69f) and I think this still needs one more authorization/data-integrity guard before merge. The new `channel_devices` policies pass the row's `owner_org`, `app_id`, and `channel_id` into `rbac_check_permission_request()`, but `rbac_check_permission_direct()` resolves `p_channel_id` back through `public.channels` and then uses the channel's own `owner_org/app_id` as the effective permission scope.

That means the permission check can succeed for channel A while the `channel_devices` row being inserted/updated still carries an independent `app_id` for app B. The table currently has separate FKs to `apps(app_id)` and `channels(id)`, and the `auto_owner_org_by_app_id` trigger only normalizes `owner_org` from the row's `app_id`; it does not prove that the submitted channel belongs to that same app/org.

The added pgTAP test verifies the policy text uses `channel_id` and that old `check_min_rights` grants are gone, but it does not exercise a mismatched channel/app row. I would keep this blocked until either a composite FK/trigger enforces `channel_devices.(app_id, owner_org)` matches the referenced channel, or the RLS policy explicitly checks that relationship, with a regression covering the mismatch case. `git diff --check origin/main...origin/pr-2193` passes locally.

PR #2194, opened 2026-05-11
fix(webhooks): align RLS API key scope with org webhook policy

teixr12:
## Summary

- Adds a webhook-specific API-key identity helper so direct webhook RLS resolution rejects app-scoped API keys for org-level webhook resources.
- Applies the same org required-expiration API-key policy at the webhook RLS layer that webhook route handlers already enforce.
- Adds regressions covering direct REST webhook visibility for app-scoped keys and legacy keys after an org requires expiring API keys.

## Why

Webhook handlers already enforce these API-key constraints before managing org-level webhook resources. Direct table access through RLS should fail closed with the same constraints while leaving the shared org API-key helper compatible for non-webhook callers.

## Test plan

- `git diff --check origin/main...HEAD`
- `bash scripts/check-supabase-migration-order.sh`
- Regression coverage added in `tests/webhooks.test.ts` and `tests/webhooks-apikey-policy.test.ts`.
- I could not run the targeted Vitest files locally in this checkout because the local environment does not have `bun` or installed project dependencies. CI is expected to exercise the added regression tests.

## Screenshots

N/A. This is a backend SQL/RLS authorization change with regression test coverage and no frontend/CLI UI changes.

## Checklist

- [ ] My code follows the code style of this project and passes `bun run lint:backend && bun run lint`.
      Local lint could not be run in this checkout because `bun` and installed dependencies are unavailable; GitHub CI is expected to run the project checks.
      N/A: this is a backend SQL/RLS authorization hardening change with no user-facing behavior or docs surface.
      N/A: no documentation change is required.
      Backend regression coverage was added in `tests/webhooks.test.ts` and `tests/webhooks-apikey-policy.test.ts`; no frontend E2E path is involved.
      Manual DB-backed testing could not be run locally without `bun`/dependencies; reproducible validation steps are listed in the Test plan.

Refs #1667
/claim #1667

## Summary by CodeRabbit

* **Security**
  * Stricter validation of webhook API key scopes (read, write, upload, all)
  * Enforced organization and app-level restrictions for webhook API keys
  * API key expiration requirements honored when configured
  * Access control policies tightened for webhook and delivery operations

* **Tests**
  * New REST tests verifying legacy non-expiring keys are denied and compliant expiring subkeys can read webhooks and deliveries
  * Added test confirming app-scoped keys cannot read specific webhook rows

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2194)

### DIFF
--- a/supabase/migrations/20260511061700_harden_webhook_rls_apikey_scope.sql
+++ b/supabase/migrations/20260511061700_harden_webhook_rls_apikey_scope.sql
@@ -135,6 +135,7 @@ COMMENT ON FUNCTION "public"."webhook_allowed_org_ids" ("min_right" "public"."us
 'Returns org IDs whose webhook rows are accessible to the current authenticated user or Capgo API key. It evaluates candidate orgs from legacy/RBAC bindings once per statement, applies webhook-specific API-key constraints, then verifies each candidate with check_min_rights() so webhook RLS can filter by indexed org_id instead of invoking authorization helpers per row.';
 
 DROP POLICY IF EXISTS "Allow admin to select webhooks" ON public.webhooks;
+DROP POLICY IF EXISTS "Allow org members to select webhooks" ON public.webhooks;
 DROP POLICY IF EXISTS "Allow admin to insert webhooks" ON public.webhooks;
 DROP POLICY IF EXISTS "Allow admin to update webhooks" ON public.webhooks;
 DROP POLICY IF EXISTS "Allow admin to delete webhooks" ON public.webhooks;
--- a/supabase/migrations/20260511061700_harden_webhook_rls_apikey_scope.sql
+++ b/supabase/migrations/20260511061700_harden_webhook_rls_apikey_scope.sql
@@ -7,166 +7,132 @@
 -- the same constraints without changing non-webhook callers of the shared helper.
 -- =============================================================================
 
-CREATE OR REPLACE FUNCTION "public"."get_identity_org_allowed_apikey_only" (
-  "keymode" "public"."key_mode" [],
-  "org_id" uuid
-) RETURNS uuid
-LANGUAGE "plpgsql"
-SECURITY DEFINER
+CREATE OR REPLACE FUNCTION "public"."webhook_allowed_org_ids" (
+  "min_right" "public"."user_min_right",
+  "keymode" "public"."key_mode" []
+) RETURNS uuid[]
+LANGUAGE "plpgsql" STABLE SECURITY DEFINER
 SET search_path = ''
 AS $$
 DECLARE
-    api_key_text text;
-    api_key record;
+  v_auth_user_id uuid;
+  v_user_id uuid;
+  v_api_key_text text;
+  v_api_key public.apikeys%ROWTYPE;
+  v_allowed uuid[] := '{}'::uuid[];
 BEGIN
-  SELECT "public"."get_apikey_header"() into api_key_text;
+  SELECT auth.uid() INTO v_auth_user_id;
+  SELECT public.get_apikey_header() INTO v_api_key_text;
 
-  -- No api key found in headers, return
-  IF api_key_text IS NULL THEN
-    PERFORM public.pg_log('deny: IDENTITY_ORG_NO_AUTH', jsonb_build_object('org_id', org_id));
-    RETURN NULL;
-  END IF;
+  IF v_api_key_text IS NOT NULL THEN
+    SELECT * FROM public.find_apikey_by_value(v_api_key_text) INTO v_api_key;
 
-  -- Use find_apikey_by_value to support both plain and hashed keys
-  SELECT * FROM public.find_apikey_by_value(api_key_text) INTO api_key;
-
-  -- Check if key was found (api_key.id will be NULL if no match) and mode matches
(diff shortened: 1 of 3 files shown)


teixr12:
Runtime/performance note for reviewers:

- Execution model: the new webhook-specific helper is used only by webhook and webhook delivery RLS policies when a request carries an API key header. It is not on the shared non-webhook helper path.
- Per-row work: one API-key lookup via the existing `find_apikey_by_value()` path, simple array/expiration checks, and one primary-key lookup against `orgs` for the target org policy.
- Cardinality: the target org lookup is one row; API-key resolution is the same bounded lookup pattern already used by the existing RLS/API-key helpers.
- CI evidence: CodSpeed reports that merging this PR does not alter performance.
- Local EXPLAIN: I cannot produce truthful `EXPLAIN (ANALYZE, BUFFERS)` output from this checkout because there is no local `psql`, `supabase`, `bun`, or running test DB available. I do not want to invent timings. A maintainer can run the helper under a local reset DB with a representative API key/org pair if project policy requires DB-backed timings before merge.

mingisrookie (commented):
One consistency point to double-check before merge: the new webhook RLS helper rejects direct table access when an existing API key's `expires_at` is farther out than `orgs.max_apikey_expiration_days`, but the route-level webhook guard still goes through `checkApikeyMeetsOrgPolicy()`, which only selects/checks `require_apikey_expiration`.

A concrete case is: create a 90-day key while the org max is 120 days, then lower the org max to 30 days. The direct PostgREST/RLS path added here denies that key, while the `/webhooks` route-level policy check still treats it as valid because `max_apikey_expiration_days` is never loaded there. The migration header says direct PostgREST should fail closed with the same constraints as route-level handlers, so either the route helper should also enforce `max_apikey_expiration_days`, or the PR should call out that direct REST is intentionally stricter than the webhook endpoints for existing keys.

teixr12:
Addressed the consistency point in `ac03fb5`: the webhook-specific RLS helper now matches the route-level webhook guard by enforcing the required-expiration policy only, without adding a direct-REST-only `max_apikey_expiration_days` denial. I also adjusted the regression setup so it no longer implies max-days route coverage.

mingisrookie (commented):
Rechecked the current head (`a622f0b`): the webhook helper now only gates on `orgs.require_apikey_expiration` + non-null `v_api_key.expires_at`, and there is no `max_apikey_expiration_days` branch in `20260511061700_harden_webhook_rls_apikey_scope.sql`. The regression fixture also sets `max_apikey_expiration_days: null`, so my original route-vs-RLS max-days consistency concern looks resolved.

digzrow-coder:
The new RLS helper appears to drop valid RBAC-only access at the final filter. `webhook_allowed_org_ids()` builds `candidate_orgs` from legacy `org_users`, user RBAC bindings, API-key RBAC bindings, and group RBAC bindings, but then every candidate is filtered with `public.check_min_rights(...)`. That helper only checks `org_users.user_right`; it does not evaluate `rbac_has_permission()` / `rbac_check_permission_direct()`.

So an org that relies on new RBAC for webhook administration can have a valid org-scoped role binding in `candidate_orgs`, then lose the row because the user/key does not also have a legacy `org_users.user_right >= admin/read` row. The tests cover legacy org keys, but not an RBAC-only org/key path through direct REST.

I would make the final filter call the same RBAC-aware permission path used by route handlers (or explicitly map webhook read/admin to RBAC permission keys), and add a regression where `use_new_rbac = true`, the principal has only an org-scoped RBAC role binding, and direct REST can still read/manage the webhook rows.

austin33133-maker:
Quick safety check on the expiration guard here: what does `public.is_apikey_expired()` return when `expires_at IS NULL`?

The PR's stated goal is to enforce the org's expiration policy at the RLS layer, but if `is_apikey_expired(NULL)` returns `false` (i.e. "a key with no expiration is not expired"), legacy non-expiring keys against an org that has `require_apikey_expiration = true` would still pass this guard and reach the table — which would contradict the route-level enforcement added in #2130.

Could you either link the `is_apikey_expired` definition for quick confirmation, or add a test case for `expires_at IS NULL` on an org with `require_apikey_expiration = true` (expected: deny + a `WEBHOOK_ALLOWED_ORGS_EXPIRED` log entry, or a new dedicated code for the null case)?

Related minor: the `limited_to_apps` check uses `COALESCE(array_length(..., 1), 0) > 0`. `array_length` returns `NULL` for both `NULL` columns and empty arrays `'{}'` — the `COALESCE` handles both consistently, but `cardinality()` would be more explicit if the column is non-null in practice.

KCDaemon (approved):
Rechecked latest head (a622f0b). I do not see the two remaining concerns reproducing on the current diff.

For null API-key expiration, the final filter explicitly requires `v_api_key.expires_at IS NOT NULL` whenever the org has `require_apikey_expiration` enabled, before the webhook rows are exposed. For RBAC-only access, `candidate_orgs` includes the RBAC sources and the final `check_min_rights(...)` call preserves the shared authorization semantics instead of relying only on legacy `org_users` membership.

The webhook-specific helper now keeps the org expiration requirement aligned with route-level API-key behavior, GitHub CI is green, and `git apply --check --whitespace=error-all` passes locally. I do not see a blocker.

PR #2196, opened 2026-05-11
fix(api): harden webhook and store metadata fetches

tngus6007:
/claim #1667

## Summary
- Keep local webhook test mode constrained to HTTP(S) schemes so enabling `CAPGO_ALLOW_LOCAL_WEBHOOK_URLS` does not also allow unsupported schemes such as `file:` or `ftp:`.
- Add a bounded icon response reader for store metadata imports so allowlisted app-store icon URLs cannot force unbounded buffering before conversion to a data URL.
- Avoid logging raw blocked webhook URLs; blocked deliveries now log URL shape metadata instead of query strings or credentials.
- Add focused regression coverage for local webhook URL scheme validation, redacted blocked-delivery logging, and oversized store metadata icon responses.

## Security note
This is public-safe hardening. The webhook changes preserve the intended local HTTP testing path while blocking non-HTTP schemes and avoiding secret-bearing URL retention in blocked-delivery logs. The store metadata change preserves existing allowlisted app-store behavior while capping fetched icon payload size.

## Test plan
- `./node_modules/.bin/vitest.exe run tests/webhook-delivery-security.unit.test.ts tests/store-metadata.unit.test.ts --reporter=verbose`
- `git diff --check`

## Screenshots
Not applicable; backend/security hardening only.

## Checklist
- [x] Code style checked
- [x] Focused tests added

### DIFF
--- a/supabase/functions/_backend/private/website_preview.ts
+++ b/supabase/functions/_backend/private/website_preview.ts
@@ -1,5 +1,6 @@
 import type { Context } from 'hono'
 import { createHono, middlewareAuth, parseBody, quickError, useCors } from '../utils/hono.ts'
+import { readResponseBytesWithLimit } from '../utils/response.ts'
 import { version } from '../utils/version.ts'
 import { getWebhookUrlValidationError } from '../utils/webhook.ts'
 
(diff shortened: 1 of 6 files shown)


mingisrookie (commented):
One logging path in the webhook hardening still keeps the raw webhook URL: when `getWebhookUrlValidationError()` blocks delivery, `deliverWebhook()` calls `cloudlogErr({ ..., url, error: urlValidationError, ... })`.

That means a rejected URL such as `https://example.test/webhook?token=secret` or a malformed/local URL with embedded credentials/query state can still be retained verbatim in operational logs, even though the delivery is correctly blocked before `fetch()`. This is especially easy to miss because the new tests assert the validation result but do not inspect the blocked-delivery log payload.

I would log URL shape metadata here instead (scheme category, host present, path segment count, `hasQuery`) or at least strip `search`/credentials before logging. A regression can call `deliverWebhook()` with a URL containing a fake query token that is rejected and assert the serialized `cloudlogErr` calls do not include the token.

tngus6007:
Addressed the logging concern in the latest commit.

Blocked webhook deliveries no longer log the raw URL. The validation log now records URL shape metadata only (`protocol`, hostname presence/length, path segment count, query/credential presence), so query strings and credentials are not retained when validation blocks delivery.

Also added regression coverage for a blocked URL containing a fake query token and the two validation edge cases CodeRabbit suggested.

Validation:
- `./node_modules/.bin/vitest.exe run tests/webhook-delivery-security.unit.test.ts tests/store-metadata.unit.test.ts --reporter=verbose`
- `git diff --check`

mingisrookie (approved):
Rechecked the current head (`1c4231f7`) and this addresses my logging concern: blocked deliveries now log `urlInfo` metadata instead of the raw `url`, and the regression asserts a query token is absent from serialized `cloudlogErr` calls. I also ran the focused tests locally:

`bun x vitest run tests/webhook-delivery-security.unit.test.ts tests/store-metadata.unit.test.ts`

Result: 2 files passed, 10 tests passed.

WcaleNieWolny:

PR #2197, opened 2026-05-11
fix: route API key writes through backend

jihadMo:
/claim #1667

## Summary
- add shared frontend helpers for the existing `PUT /apikey/:id` and `DELETE /apikey/:id` backend routes
- route API-key rename, scope save, regenerate, and delete UI writes through those backend routes instead of direct browser-side `apikeys` table mutations
- align `GET /apikey/:id` and `DELETE /apikey/:id` lookup behavior with `PUT /apikey/:id` so numeric ids and UUID key values are queried separately
- add regression coverage for fetching and deleting a plain API key by UUID key value

## Why
The backend API-key routes already centralize ownership checks, limited-key guards, validation, and policy enforcement. The UI should use that path for API-key mutations instead of writing the table directly.

The GET/DELETE lookup change also avoids mixing UUID key values into the numeric `id` filter, matching the safer lookup pattern already used by the PUT handler.

Related: #1667

## Testing
- `git diff --check`
- `npx.cmd --yes vitest run tests/apikeys.test.ts` *(blocked locally: this checkout has no installed dependencies; Vitest cannot resolve `vite` from `vitest.config.ts`)*
- `npx.cmd --yes eslint src/services/apikeys.ts src/pages/ApiKeys.vue src/components/organization/ApiKeyRbacManager.vue src/pages/settings/organization/ApiKeys.[id].vue supabase/functions/_backend/public/apikey/get.ts supabase/functions/_backend/public/apikey/delete.ts tests/apikeys.test.ts` *(blocked locally: ESLint cannot resolve `@antfu/eslint-config` without installed dependencies)*

## Summary by CodeRabbit

* **New Features**
  * UI now uses centralized API-key service for create/rename/regenerate/delete flows.

* **Bug Fixes**
  * Tighter validation and scoping for API key deletion to prevent unintended access.
  * More reliable lookup logic for identifying keys across legacy and numeric formats.

* **Tests**
  * Added end-to-end tests covering legacy plain (UUID) API key creation, retrieval, and deletion.

* **Chores**
  * Introduced a helper to standardize API-key lookup and service routes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2197)

### DIFF
--- a/src/components/organization/ApiKeyRbacManager.vue
+++ b/src/components/organization/ApiKeyRbacManager.vue
@@ -12,10 +12,12 @@ import DataTable from '~/components/DataTable.vue'
 import {
   confirmApiKeyDeletion,
   confirmApiKeyRegeneration,
+  deleteApiKey,
   formatApiKeyScope,
   isApiKeyExpired,
   showApiKeySecretModal,
   sortApiKeyRows,
+  updateApiKey,
 } from '~/services/apikeys'
 import { formatDate, formatLocalDate } from '~/services/date'
 import { useSupabase } from '~/services/supabase'
@@ -362,7 +364,7 @@ async function deleteKey(apiKey: OrgApiKey) {
 
   isSubmitting.value = true
   try {
-    const { error } = await supabase.from('apikeys').delete().eq('id', apiKey.id)
+    const { error } = await deleteApiKey(supabase, apiKey.id)
     if (error)
       throw error
     toast.success(t('removed-apikey'))
@@ -381,10 +383,7 @@ async function regenerateKey(apiKey: OrgApiKey) {
   if (!await confirmApiKeyRegeneration(dialogStore, t))
     return
 
-  const { data, error } = await supabase.functions.invoke('apikey', {
-    method: 'PUT',
-    body: { id: apiKey.id, regenerate: true },
-  })
+  const { data, error } = await updateApiKey(supabase, apiKey.id, { regenerate: true })
   if (error || !data) {
     toast.error(t('failed-to-regenerate-api-key'))
     return
--- a/src/pages/ApiKeys.vue
+++ b/src/pages/ApiKeys.vue
@@ -16,10 +16,12 @@ import IconTrash from '~icons/heroicons/trash'
 import {
   confirmApiKeyDeletion,
   confirmApiKeyRegeneration,
+  deleteApiKey,
   formatApiKeyScope,
   isApiKeyExpired,
   showApiKeySecretModal,
   sortApiKeyRows,
+  updateApiKey,
 } from '~/services/apikeys'
 import { formatLocalDate } from '~/services/date'
 import { useSupabase } from '~/services/supabase'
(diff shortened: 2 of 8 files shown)


Flob365 (commented):
Thanks for moving the API key mutations behind the backend route. One blocker I found from the latest CI run: `bun typecheck` fails in `supabase/functions/_backend/public/apikey/delete.ts` after the new `getApiKeyLookupFilter()` helper is used.

The failing job reports:

```
supabase/functions/_backend/public/apikey/delete.ts(...): error TS2339: Property 'id' does not exist on type 'GenericStringError'.
```

This looks like a Supabase type inference issue caused by passing the dynamic `lookup.column` union into `.eq(...)`; the query builder can no longer prove that the selected row is an `apikeys` row, so `data` becomes `GenericStringError` and `apikey.id` is rejected.

A small fix would be to keep the runtime behavior but branch explicitly, similar to `put.ts`:

```ts
const baseQuery = supabase
  .from('apikeys')
  .select('id')
  .eq('user_id', auth.userId)

const apikeyQuery = /^\d+$/.test(id)
  ? baseQuery.eq('id', Number(id))
  : baseQuery.eq('key', id)

const { data: apikey, error: apikeyError } = await apikeyQuery.single()
```

That should preserve the cast-error avoidance while restoring the typed row shape. The same pattern may also be worth using in `get.ts` if the typecheck reaches that file after `delete.ts` is fixed.

mingisrookie (commented):
One security tradeoff to call out: the UI changes now pass numeric IDs, which is good, but the backend/tests still deliberately support using a plaintext API key value as the `:id` path segment for GET/DELETE. That preserves compatibility with the old `.or(key.eq...,id.eq...)` behavior, but it also means raw API key material can appear in URLs, reverse-proxy logs, request traces, browser history, and test output.

If that compatibility path is required, I would document it as legacy/deprecated and keep the new UI/helper examples anchored on numeric `id` only. The new regression tests currently exercise `/apikey/${createData.key}` directly; a short comment there (or an additional test asserting the UI path uses `id`) would make it clearer that secret-in-path is only for backward compatibility rather than the preferred contract going forward.

Flob365:
Good point. I agree the numeric id path should be the preferred contract, and the key-as-path-segment behavior should be treated as legacy compatibility only. The UI helpers added here already call the route with numeric ids, so documenting/deprecating the plaintext-key route and making the regression tests explicitly label it as legacy would make the intent clearer.

jihadMo:
Thanks, good point. I pushed c04472b to make the intent explicit: the frontend API-key helpers now accept numeric IDs only, the backend lookup helper labels plaintext-key path usage as legacy compatibility, and the regression tests are renamed/commented as legacy coverage while using getEndpointUrl().

mingisrookie (approved):
Verified the follow-up at c04472b. The new frontend service helpers now take a numeric `ApiKeyId` and build routes from that id, the UI callers pass `apiKey.id`, and the remaining key-as-URL tests/helper path are explicitly marked as legacy compatibility. I also re-ran the focused API-key regression suite successfully:

`bun x vitest run tests/apikeys.test.ts` → 36 passed.

The remaining full `vue-tsc` run is still blocked locally by unrelated CLI test/module errors (`@capgo/cli/sdk` resolution and an implicit-any in `tests/cli-channel.test.ts`), not by the touched API-key route files. From the reviewed scope, the legacy-path intent is now clear.

WcaleNieWolny:

PR #2198, opened 2026-05-11
fix(cli): clean up Android onboarding checks

mySebbe:
## Summary
- fix Android onboarding regex lint violations from `regexp/use-ignore-case`
- stop persisting OAuth refresh tokens and service-account JSON key material in resumable onboarding progress
- remove unused picker/revoke helpers and include CLI test files in the knip export graph

## Validation
- `bun run cli:lint`
- `bun run lint:deadcode`
- worker also ran `cli:typecheck`, `cli:build`, `test/test-android-keystore.mjs`, and `test/test-android-oauth.mjs`

Follow-up for #2064 failing checks.

### DIFF
--- a/cli/src/build/onboarding/android/ui/app.tsx
+++ b/cli/src/build/onboarding/android/ui/app.tsx
@@ -144,7 +144,7 @@ const AndroidOnboardingApp: FC<AppProps> = ({ appId, initialProgress, androidDir
     initialProgress?.completedSteps.googleSignInComplete || null,
   )
   const [accessToken, setAccessToken] = useState<string>('')
-  const [refreshTokenState, setRefreshTokenState] = useState<string>(initialProgress?._oauthRefreshToken || '')
+  const [refreshTokenState, setRefreshTokenState] = useState<string>('')
   const [oauthClientId, setOauthClientId] = useState<string>('')
   const [oauthStatusMessages, setOauthStatusMessages] = useState<string[]>([])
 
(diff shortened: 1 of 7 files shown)


Serve63 (commented):
Reviewed the Android onboarding cleanup diff. No blocker from my side.

The important bit looks good to me: progress no longer resumes from persisted `_oauthRefreshToken` or `_serviceAccountKeyBase64`, while the active onboarding run still keeps those values in memory long enough to finish the current flow. That is the safer trade-off for a resumable progress file.

Only caveat I would keep explicit for merge: after a crash between Google/service-account provisioning and the final credential write, the user may need to rerun that step because the sensitive material is no longer saved in progress. Given the stated validation (`cli:lint`, `lint:deadcode`, typecheck/build and Android OAuth/keystore tests), I do not see a PR-specific blocker here.

cli/src/build/onboarding/android/ui/app.tsx
```diff
@@ -144,7 +144,7 @@ const AndroidOnboardingApp: FC<AppProps> = ({ appId, initialProgress, androidDir
     initialProgress?.completedSteps.googleSignInComplete || null,
   )
   const [accessToken, setAccessToken] = useState<string>('')
-  const [refreshTokenState, setRefreshTokenState] = useState<string>(initialProgress?._oauthRefreshToken || '')
+  const [refreshTokenState, setRefreshTokenState] = useState<string>('')
```
darshan-Jahagirdar, line 147: This breaks resume after Google sign-in. `getAndroidResumeStep()` still skips `google-sign-in` whenever `completedSteps.googleSignInComplete` exists, but after this change a restarted CLI initializes `refreshTokenState` to `''`. If the user exits/crashes after sign-in and then resumes into `gcp-projects-loading` or `gcp-setup-running`, `ensureAccessToken()` throws `Not signed in - re-run onboarding to re-authenticate`, and retrying cannot recover because the progress file keeps them past the sign-in step.

Same pattern also affects final save after `serviceAccountProvisioned`: the progress can resume directly into `saving-credentials`, but `serviceAccountKeyBase64` is empty, so `doSaveCredentials()` throws `service-account key not provisioned`. If these secrets should not be persisted, the resume logic needs to route back to a step that can recreate them instead of treating the completed metadata as enough to continue.

PR #2203, opened 2026-05-11
fix(api): cap translation request bodies

tngus6007:
/claim #1667

## Summary
- Replace direct unbounded JSON body parsing on both translation message entrypoints with bounded stream reading.
- Reject translation request bodies above 1 KiB with `413 request_body_too_large` before cache, D1 storage, queue, or AI work.
- Add regression coverage for streamed oversized bodies and `Content-Length` fast-fail handling in both translation implementations.

## Security note
This is public-safe DoS hardening for the translation message endpoint. The accepted payload only needs `targetLanguage`, so the cap preserves normal use while avoiding unnecessary memory work on oversized request bodies.

## Test plan
- `git diff --check`
- `npx --yes bun@latest x vitest run tests/translation-public-body-limit.unit.test.ts tests/translation-queue.unit.test.ts` returned exit code 0 in my local shell, but produced no stdout.

## Screenshots
Not applicable; backend worker/API hardening only.

## Checklist

### DIFF
--- a/tests/translation-public-body-limit.unit.test.ts
+++ b/tests/translation-public-body-limit.unit.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from 'vitest'
+import { app } from '../supabase/functions/_backend/public/translation.ts'
+
+describe('public translation body limit', () => {
+  it('rejects oversized content-length before reading the body', async () => {
+    const response = await app.request(new Request('http://local/messages', {
+      body: JSON.stringify({ targetLanguage: 'fr' }),
+      duplex: 'half',
+      headers: {
+        'Content-Length': '2048',
+        'Content-Type': 'application/json',
+      },
+      method: 'POST',
+    } as RequestInit))
+
+    expect(response.status).toBe(413)
+    expect(await response.text()).toContain('Request body is too large')
+  })
+
+  it('rejects streamed bodies that exceed the translation request limit', async () => {
+    const response = await app.request('http://local/messages', {
+      body: JSON.stringify({
+        targetLanguage: 'fr',
+        padding: 'x'.repeat(2048),
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      method: 'POST',
+    })
+
+    expect(response.status).toBe(413)
+    expect(await response.text()).toContain('Request body is too large')
+  })
+})
--- a/tests/translation-queue.unit.test.ts
+++ b/tests/translation-queue.unit.test.ts
@@ -195,4 +195,55 @@ describe('translation queue helpers', () => {
     expect(payload.status).toBe('pending')
     expect(queue.send).toHaveBeenCalledTimes(1)
   })
+
+  it('rejects oversized translation request bodies before touching storage', async () => {
+    stubWorkerCache()
+    const db = createTranslationStoreMock(null)
+    const queue = {
+      send: vi.fn(),
+    }
+    const response = await translationWorker.fetch(new Request('https://api.capgo.app/translation/messages', {
+      body: JSON.stringify({
+        targetLanguage: 'fr',
+        padding: 'x'.repeat(2048),
+      }),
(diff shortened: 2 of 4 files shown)


mingisrookie (commented):
This looks like a good focused hardening change. One small regression gap I'd consider before merging: the new helper has two separate enforcement paths, but the test only appears to exercise the streaming/read path where the body grows past the limit. The early `Content-Length > MAX_TRANSLATION_REQUEST_BODY_BYTES` fast-fail path is also security-relevant because it is the cheapest DoS rejection path and it happens before any body read.

If this branch can be reached in the Worker test harness, adding a case with an oversized `Content-Length` header and asserting the same `413 request_body_too_large` / no `db.prepare` / no `queue.send` behavior would prevent a future refactor from accidentally dropping the pre-read rejection while leaving the streaming test green.

tngus6007:
Added coverage for the `Content-Length` fast-fail path as requested. The new test asserts oversized declared length returns `413 request_body_too_large` before the request stream is pulled, and before D1 storage or queue access.

Re-ran:
- `npx --yes bun@latest x vitest run tests/translation-queue.unit.test.ts`
- `git diff --check`

tngus6007:
Also covered the existing `/translation/messages` API route imported by `cloudflare_workers/api/index.ts`, so both translation entrypoints now enforce the same 1 KiB request body cap. Added focused tests for the public API route's streamed overflow path and `Content-Length` fast-fail path.

Re-ran:
- `npx --yes bun@latest x vitest run tests/translation-public-body-limit.unit.test.ts tests/translation-queue.unit.test.ts`
- `git diff --check`

tests/translation-public-body-limit.unit.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      headers: {
+        'Content-Length': '2048',
+        'Content-Type': 'application/json',
+      },
+      method: 'POST',
+    } as RequestInit))
+
+    expect(response.status).toBe(413)
+    expect(await response.text()).toContain('Request body is too large')
+    expect(pullCalled).toBe(false)
```
mingisrookie: I re-ran the focused tests locally with the checked-out head (`bun x vitest run tests/translation-public-body-limit.unit.test.ts tests/translation-queue.unit.test.ts`, Bun 1.3.11). Both new Content-Length tests fail on this assertion: the response is still `413`, but `pullCalled` is already `true`.

That makes this regression test runtime-sensitive: `ReadableStream` can be pulled to fill its queue even if the route later fast-fails from `Content-Length`, so `pullCalled === false` is not a portable proof that the handler skipped body reading. I would either remove this assertion and keep the stable `413` / no DB / no queue checks, or spy on a lower-level `getReader()`/limited-reader boundary if you need to prove the handler does not consume the stream.

tests/translation-queue.unit.test.ts
```diff
@@ -195,4 +195,64 @@ describe('translation queue helpers', () => {
...
+      method: 'POST',
+    } as RequestInit), {
+      DB_TRANSLATIONS: db,
+      TRANSLATION_MESSAGES_QUEUE: queue,
+    } as any)
+    const payload = await response.json() as { error: string }
+
+    expect(response.status).toBe(413)
+    expect(payload.error).toBe('request_body_too_large')
+    expect(pullCalled).toBe(false)
```
mingisrookie: Same failing assertion as the public-route test when run locally with `bun x vitest run tests/translation-public-body-limit.unit.test.ts tests/translation-queue.unit.test.ts`: status/error and no storage/queue interaction pass, but `pullCalled` is `true`. The fast-fail behavior is worth keeping, but this specific stream-pull assertion looks brittle across runtimes.

mingisrookie (commented):
Follow-up after re-running the requested coverage: the Content-Length behavior returns 413, but the new tests are currently failing locally because the stream pull spy is not stable evidence of route-level body consumption.

WcaleNieWolny:

PR #2211, opened 2026-05-11
feat(cli/build init): import existing iOS signing credentials from this Mac

WcaleNieWolny:
## Summary

`build init` now offers an **"Import existing from this Mac"** branch as a peer of the existing **"Create new via App Store Connect API"** path. iOS developers who already have a distribution cert in their login Keychain and a matching Xcode provisioning profile can now finish onboarding without revoking certs or touching the Apple Developer Portal.

Closes the user-reported gap: *"the onboarding will never allow the user to export the key from their machine."*

## What changes for the user

When a macOS user runs `npx @capgo/cli build init` and selects iOS, they now see:

```
How do you want to set up iOS credentials?
  📥  Import existing from this Mac (Keychain + Xcode profiles)
  🆕  Create new via App Store Connect API
```

The import branch:
- silently inventories `security find-identity` + scans both legacy (`~/Library/MobileDevice/Provisioning Profiles`) and Xcode 16+ (`~/Library/Developer/Xcode/UserData/Provisioning Profiles`) directories — **no Keychain prompt during inventory**
- matches identities to profiles by SHA1 of embedded developer certs
- prompts the user to pick an identity + a matching profile + distribution mode
- shows a clear heads-up: *"macOS will pop a Keychain dialog. Click **Always Allow**."*
- exports the chosen identity as a single-cert PKCS#12 via `/usr/bin/security export` + node-forge filtering — **exactly one Keychain GUI dialog, regardless of how many identities you have**
- routes `ad_hoc` straight to saving credentials; `app_store` reuses the existing `.p8` / Key ID / Issuer ID screens for TestFlight upload

## Why this is needed

The current iOS flow is single-track and Apple-API-driven: it always creates a brand-new distribution cert + provisioning profile via the App Store Connect API. Two failure modes:

1. **Apple's 3-cert limit.** Users at the limit are forced to revoke an existing cert (which may be in active use by CI or other teammates). The import path sidesteps this entirely.
2. **No ASC API key access.** Some developers (contractors, smaller teams) don't have Admin access to App Store Connect and can't generate the `.p8`. For `ad_hoc` distribution, importing makes the `.p8` requirement go away entirely.

## Implementation

- **New module: `cli/src/build/onboarding/macos-signing.ts`** — wraps `/usr/bin/security` subprocess calls, scans provisioning-profile directories, matches identities to profiles, and exports a filtered single-identity P12. Pure functions where possible; subprocess runner is dependency-injected for testing.
- **Extended `cli/src/build/mobileprovision-parser.ts`** — added `parseMobileprovisionDetailed()` that exposes team ID, expiration date, profile type (`app_store` / `ad_hoc` / `development` / `enterprise`), and SHA1 of every embedded developer cert. Existing `parseMobileprovision` and `parseMobileprovisionFromBase64` are unchanged.
- **Extended `cli/src/build/onboarding/types.ts`** — 7 new step values + corresponding `STEP_PROGRESS` entries + `getPhaseLabel` cases. All additive; existing step values unchanged.
- **Extended `cli/src/build/onboarding/ui/app.tsx`** — adds the fork at `platform-select` (gated by `isMacOS()`), the import sub-flow renders, and corresponding `useEffect` handlers. `doSaveCredentials` is updated to handle the import-mode case.

## v1 scope

| In | Out |
|---|---|
| Single-bundle apps | Multi-bundle / extensions / widgets (use existing `build credentials update --ios-provisioning-profile bundleId=path`) |
| `login.keychain-db` (default Keychain) | Custom keychains |
| Both `app_store` and `ad_hoc` distribution | Enterprise distribution |
| macOS only | Linux/Windows (fork hidden, existing path unchanged) |

## Security considerations

- Generates a cryptographically random 32-byte hex passphrase per export — no static passphrase
- Stored in `~/.capgo-credentials/credentials.json` at `0o600` alongside existing static `'capgo'` passphrase from create-new path → no regression
- The `security export` call writes a temp `.p12` into a `mkdtemp` directory and `rm -rf`s it in `finally` whether export succeeds or fails
- Subprocess uses `spawn(SECURITY_BIN, [...args])` with absolute path and array argv — no shell interpolation

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun run lint` clean
- [x] `bun run test:macos-signing` — 11/11 pass
- [x] `bun test/test-mobileprovision-parser.mjs` — 9/9 pass (4 existing + 5 new)
- [x] Existing regression tests: `test:credentials` (15/15), `test:credentials-validation` (13/13), `test:provisioning-map-validation` (4/4), `test:platform-paths` (4/4)
- [ ] Manual smoke test on a Mac with iOS distribution cert in Keychain — before marking ready
- [ ] Verify exported P12 imports on build server (`security import` on Linux) — confirm on first end-to-end build

## Pre-existing failure (not introduced by this PR)

`test:onboarding-recovery` has 1 pre-existing failure in `updater install state requires node_modules install` on `origin/main`. None of the files touched here relate to `readUpdaterState` or `updater.ts`.

## Rollback

Remove the `'import'` option from the `Select` widget in the `setup-method-select` render branch. The `macos-signing.ts` module + new step values become dead code (kept until next cleanup) but harmless.

## Summary by CodeRabbit

* **New Features**
  * macOS import flow to discover Keychain signing identities, match local provisioning profiles, select identity/profile, and export PKCS#12 for onboarding.
  * Swift helper tool to export a Keychain identity as PKCS#12.

* **Enhancements**
  * Mobileprovision parser now returns detailed metadata (team, expiration, profile type, certificate SHA1s).
  * Onboarding UI and resume flow updated for macOS import-existing scenarios.

* **Tests**
  * Added unit tests for SHA1 utilities, mobileprovision parsing, and macOS signing helpers.

* **Chores**
  * Build/test scripts updated and Swift helper included in CLI builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2211)


Fleehopper (commented):
Left two things I would fix before marking this ready:

1. `cli/src/build/onboarding/macos-signing.ts:376` uses `require('node-forge')` inside the new export path. The CLI package is `type: module` and the TS config emits ESNext modules; this function is not covered by the new tests because they inject `forgeFilter` / avoid the real P12 filter path. On the actual import flow, `filterP12ToSingleIdentity()` can be the first time this branch runs, so I would switch this to an ESM-safe lazy import path or a top-level `import * as forge from 'node-forge'` and add a small test that exercises the default filter path enough to catch module loading.

2. `cli/src/build/onboarding/ui/app.tsx:1050-1083` lets the user select any profile that matches the certificate SHA1, regardless of bundle ID or profile type. That can include an old profile for a different app, wildcard/profile edge cases, or an enterprise profile even though enterprise is explicitly out of v1 scope. `doSaveCredentials()` then persists the map under `chosenProfile.bundleId`, so onboarding can finish successfully while the next build has no profile for the actual Capacitor app target. I would filter/warn at selection time to `appId`-compatible profiles and reject `enterprise`/`development` profiles unless that support is intentionally added.

I couldn't run the Bun test suite locally because this environment does not have `bun` on PATH, so this is a source review rather than a verified run.

cli/src/build/onboarding/ui/app.tsx
```diff
@@ -833,6 +1088,315 @@ const OnboardingApp: FC<AppProps> = ({ appId, initialProgress, iosDir, apikey })
...
+              {matchedProfiles.length}
+              {' '}
+              available for this identity):
+            </Text>
+            <Newline />
+            <Select
+              options={[
+                ...matchedProfiles.map(p => ({
+                  label: `📜  ${p.name} · bundle ${p.bundleId} · ${p.profileType} · expires ${p.expirationDate.split('T')[0]}`,
+                  value: p.path,
```
digzrow-coder: This breaks the Apple-side profile recovery path when the certificate has more than one profile. `import-fetching-profile` synthesizes those profiles with `path: ''`, but this picker uses `p.path` as the option value and then resolves the selection with `find(p => p.path === value)`. Every fetched profile therefore has the same empty value, so selecting the second/third profile either resolves to the first one or does not change selection at all.

That can save the wrong provisioning profile for a different bundle/distribution mode. Use a stable unique value here, for example `p.path || p.uuid`, and do the lookup against the same key.

cli/src/build/onboarding/apple-api.ts
digzrow-coder, line 405: This ad-hoc recovery branch still creates an App Store profile.

The import flow lets the user choose `ad_hoc`, and the no-match recovery menu then offers to create a new profile for the existing certificate. That path calls `createProfile()`, but `createProfile()` hardcodes `profileType: 'IOS_APP_STORE'`. So an ad-hoc user can end up saving an App Store provisioning profile while the saved credentials say `CAPGO_IOS_DISTRIBUTION: 'ad_hoc'`, which will not work for QR/direct device installs.

The profile type should be derived from `importDistribution` here, e.g. pass `IOS_APP_ADHOC` for ad-hoc recovery and keep `IOS_APP_STORE` for TestFlight.

digzrow-coder (commented):
The import flow can save a provisioning profile that does not match the distribution mode selected earlier.

`import-distribution-mode` now asks the user to choose `app_store` or `ad_hoc` before identity/profile selection, and `doSaveCredentials()` later persists that original `importDistribution`. But `import-pick-profile` builds its options from all `matchedProfiles` for the identity without filtering or validating `p.profileType` against `importDistribution`; the Apple recovery path also appends every profile returned by `listProfilesForCert()`.

So a user can choose `App Store / TestFlight`, then pick a locally matching ad-hoc profile. The saved credentials will have `distribution: "app_store"` and an ad-hoc provisioning profile, so onboarding appears successful but the first TestFlight/build signing path fails later with a hard-to-diagnose profile mismatch. The inverse is also possible for ad-hoc mode with an app-store profile.

Please filter the profile picker to the selected distribution mode, or validate before `import-export-warning`/`saving-credentials` and route the user back to the distribution/profile choice with a clear error. A regression test can cover a match set containing one `app_store` and one `ad_hoc` profile and assert only the selected mode is accepted.

albercr3:
I think the Apple-side recovery path still has a selection bug when more than one profile is returned for the imported certificate.

`import-fetching-profile` synthesizes each Apple profile with `path: ''` because there is no local file, then appends all of them to `importMatches`. Later, `import-pick-profile` uses `value: p.path` for every option and resolves the selection with `matchedProfiles.find(p => p.path === value)`. For Apple-recovered profiles, every option therefore has the same empty value, so selecting the second/third profile still resolves to the first synthesized profile. The log can show the chosen row, but `chosenProfile` will be whichever empty-path entry appears first.

This matters for the fallback flow because `listProfilesForCert()` can return multiple profiles for different bundle IDs or distribution types, and the user may be unable to pick the one that matches the app. Please use a stable unique picker value such as `p.path || p.uuid` (or an `apple:${p.uuid}` synthetic path) and resolve by the same key. A regression can seed two synthesized profiles with empty `path` and assert selecting the second one sets `chosenProfile.uuid` to the second profile.

cli/src/build/onboarding/ui/app.tsx
```diff
@@ -138,6 +140,23 @@ const OnboardingApp: FC<AppProps> = ({ appId, initialProgress, iosDir, apikey })
...
 
+  // Import-existing sub-flow state (macOS only)
+  const [importMatches, setImportMatches] = useState<IdentityProfileMatch[]>([])
+  const [importProfiles, setImportProfiles] = useState<DiscoveredProfile[]>([])
+  const [chosenIdentity, setChosenIdentity] = useState<SigningIdentity | null>(null)
+  const [chosenProfile, setChosenProfile] = useState<DiscoveredProfile | null>(null)
+  const [importDistribution, setImportDistribution] = useState<'app_store' | 'ad_hoc' | null>(null)
+  const [importedP12Password, setImportedP12Password] = useState<string>('')
+  /** Tracks whether we're in the import flow so saving-credentials knows which writer to use */
+  const [importMode, setImportMode] = useState(false)
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 81a6f1c74

digzrow-coder:
I reproduced the current CI blockers locally and pushed a small patch branch against this PR head:

- Branch: https://github.com/digzrow-coder/capgo/tree/codex/capgo-pr2211-ci-fixes
- Commit: `477a23ad5` (`Fix Capgo import PR lint blockers`)

What it fixes:
- `lint:deadcode`: removes the unused `DEFAULT_LOGIN_KEYCHAIN` export and unused `NoIdentitiesError` class from `cli/src/build/onboarding/macos-signing.ts`.
- `typos`: changes `unparseable` -> `unparsable`, avoids the `mis-route` spelling hit in the UI comment, and changes test fixture `fake-udid` to `fake-device-id`.

Validation I ran on the patch:
- `bun run lint:deadcode`
- `typos v1.45.1 --config ./.typos.toml` on the changed CLI files/tests
- `bun run typecheck` from `cli/`
- `bun run test:macos-signing` from `cli/`
- `bun run test:apple-api-import-helpers` from `cli/`
- `bun test/test-mobileprovision-parser.mjs` from `cli/`
- `git diff --check`

I did not open a competing upstream PR since this is just a helper patch for #2211.

digzrow-coder:
Update: I extended the helper branch to cover the CodeQL failures too and force-updated it to a single current commit:

- Branch: https://github.com/digzrow-coder/capgo/tree/codex/capgo-pr2211-ci-fixes
- Current commit: `a0279a763` (`Fix Capgo import PR lint blockers`)

Additional fix included:
- Adds narrow `lgtm[js/weak-cryptographic-algorithm]` suppressions in the SHA1 certificate-fingerprint tests. These tests are intentionally checking Apple/Keychain SHA1 fingerprint compatibility, not using SHA1 for password/security hashing.

Validation rerun on `a0279a763`:
- `bun run lint:deadcode`
- `typos v1.45.1 --config ./.typos.toml cli`
- `bun run typecheck` from `cli/`
- `bun run test:macos-signing` from `cli/`
- `bun run test:apple-api-import-helpers` from `cli/`
- `bun test/test-mobileprovision-parser.mjs` from `cli/`
- `git diff --check`

digzrow-coder:
I put together a helper branch for the current CI/review blockers:

https://github.com/digzrow-coder/capgo/tree/codex/capgo-pr2211-helper-forkbase

What it addresses:
- Fixes `lint:deadcode` by removing the unused macOS signing exports.
- Fixes the current CLI typo failures (`udid`, `unparseable`, `mis-route`).
- Avoids the CodeQL SHA1 test alerts by replacing test-side SHA1 recomputation with known expected hashes.
- Preserves duplicate-profile retry context so import recovery returns to `import-create-profile-only` instead of always retrying `creating-profile`.
- Makes import profile creation honor `importDistribution` (`app_store` vs `ad_hoc`) in the ASC profile type and synthesized profile metadata.
- Filters import profile selection by both `appId` and `importDistribution`, and validates the selected UUID against the same filtered set.

Validated locally:
- `bun run lint:deadcode`
- `typos ./cli --config ./.typos.toml`
- `bun run --cwd cli test:macos-signing`
- `bun run --cwd cli test:apple-api-import-helpers`
- `bun test/test-mobileprovision-parser.mjs` from `cli/`
- `bun run --cwd cli typecheck`
- `bun run --cwd cli lint`
- `git diff --check`

I also tried full `bun run --cwd cli test`, but on this Windows shell it stops at `test:version-detection:setup` because Bun cannot execute `./test/fixtures/setup-test-projects.sh` via the POSIX path. The targeted tests touched by this patch pass.

digzrow-coder:
Follow-up: the fork CI for the helper branch is green now: https://github.com/digzrow-coder/capgo/actions/runs/25905147169

radiantjade:
I think the helper cache path is too trusting for a flow that exports signing private keys.

`compiledHelperPath()` resolves to a predictable file in the shared OS temp directory (`/tmp/capgo-keychain-export-v<version>`), and `ensureSwiftHelper()` returns that path immediately when `existsSync(cached)` is true. On macOS `/tmp` is world-writable, so another local account or a prior compromised process can pre-create an executable at that versioned path before the CLI runs. The onboarding flow will then execute that binary as the current user and pass it the selected identity SHA1, output path, and freshly generated P12 passphrase. Because this helper is the component that receives Keychain export prompts and handles the P12, executing an unverified temp-file cache is a pretty sensitive trust boundary.

I would avoid a shared predictable executable here. Safer options are: compile into a per-user 0700 cache directory such as `~/Library/Caches/capgo/` and verify owner/mode/non-symlink before reuse, or skip the persistent cache and compile into a fresh `mkdtemp()` directory for each run. If caching remains, I would also hash the bundled Swift source/compiler output metadata and reject any cached helper whose owner/mode/hash do not match expectations. A regression can pre-create the cache path with a fake executable and assert `ensureSwiftHelper()` does not execute/reuse it.

nasuf:
I think there is still a resume regression in the import path after the API key is verified.

`import-distribution-mode` persists `setupMethod = 'import-existing'` and `importDistribution`, and the component hydrates `importMode` / `importDistribution` from those fields on startup. But the `verifying-key` handler constructs a fresh `OnboardingProgress` object and saves it with only `p8Path` and `completedSteps.apiKeyVerified`; it does not preserve the existing `setupMethod` or `importDistribution` values.

That means an app-store import session can follow the intended path initially, then if the CLI is interrupted after `verifying-key` succeeds but before credentials are saved, the next run loads progress without `setupMethod`. `getResumeStep()`/state hydration then treats it as the create-new flow, which can send the user back toward certificate creation instead of `import-pick-identity`.

The fix should be to merge the existing progress record when saving `apiKeyVerified` here, or explicitly carry forward `setupMethod` and `importDistribution` into the new progress object. A small regression test can simulate: choose import-existing + app_store, run the verifying-key save, reload progress, and assert `setupMethod === 'import-existing'` and `importDistribution === 'app_store'` are still present.

PR #2213, opened 2026-05-11
style(ui): polish shared console surfaces

riderx:
## Summary (AI generated)

- Polished shared console chrome: sidebar, navbar, tab rows, app/settings/admin layout backgrounds, table controls, and chart cards.
- Refined auth page surfaces by reducing oversized radii, removing decorative orb backgrounds, and improving card/background contrast.
- Fixed the app table footer range from zero-based display to human-readable pagination and included before/after screenshots for every changed page reviewed.

## Motivation (AI generated)

The console already has a strong base, but several repeated surfaces felt heavier or less tactile than they needed to be: inactive nav items had low contrast, active tabs blended into the page background, table controls had inconsistent target sizes, and the auth pages leaned on decorative blobs. These tweaks make the product feel cleaner without changing the underlying structure or workflows.

- Improves perceived quality across the main Capgo dashboard without a full redesign.
- Makes repeated admin and app-management workflows easier to scan and tap.
- Keeps the work low-risk by focusing on shared visual primitives and small behavior polish.

## Screenshots (AI generated)

| Page | Before | After |
| --- | --- | --- |
| Login | Login before | Login after |
| Dashboard | Dashboard before | Dashboard after |
| Apps | Apps before | Apps after |
| App overview | App overview before | App overview after |
| App bundles | App bundles before | App bundles after |
| Settings organization | Settings organization before | Settings organization after |

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun typecheck`
- [x] `bun run build`
- [x] `bun test:front` - 19 passed

## Summary by CodeRabbit

* **New Features**
  * Added configurable page size option for tables, with default of 10 items per page.

* **Style**
  * Refreshed UI color theme from blue to slate across layouts and backgrounds.
  * Updated component styling for tables, navigation, cards, and buttons for improved visual consistency.
  * Enhanced font rendering with improved antialiasing.

* **Bug Fixes**
  * Simplified authentication for replication status loading to use session-based tokens.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2213)


riderx:
@coderabbitai review

albercr3:
﻿I think the new pagination math is only partially wired up. `DataTable` now uses `props.pageSize ?? offset.value`, but most existing server-paginated callers still pass `:offset="offset"` instead of `:page-size`, for example `BuildTable.vue`, `BundleTable.vue`, `ChannelTable.vue`, and `DeviceTable.vue`. Since `offset` is not a declared `DataTable` prop, Vue treats it as an extra attribute on the root element and `paginationPageSize` falls back to `elementList.length`.

That fallback breaks on a short final page. If a table has 25 total rows, a real page size of 10, and page 3 returns 5 rows, the range becomes `11-15` instead of `21-25`, `canNext()` still compares `3 < Math.ceil(25 / 5)`, and the next/fast-forward controls can move past the real last page.

Could we either keep `offset` as a backwards-compatible alias in `DataTable` or update all the server-paginated callers to pass `:page-size="offset"` / their fixed page size? A regression around a 25-row table with a 5-row last page would catch this.

KCDaemon (changes requested):
Rechecked the current diff and the issue is still present.

`DataTable` now computes pagination from `props.pageSize ?? offset.value`, but most server-paginated callers in this PR still do not pass `page-size`. I confirmed `BuildTable.vue` still passes the old `:offset="offset"` attribute, while `BundleTable.vue`, `ChannelTable.vue`, and `DeviceTable.vue` do not pass a page size at all even though their data loaders fetch fixed-size pages (`offset = 10` or `20`). `offset` is not declared as a `DataTable` prop, so it does not feed `paginationPageSize`; the component falls back to its internal default.

That breaks the new range / next-page math whenever the real server page size differs from the fallback. For example, `BuildTable` fetches 20 rows per page but `DataTable` will calculate with the default 10, so page 2 of 25 rows is displayed as `11-25` instead of `21-25`, and fast-forward / next availability can target the wrong last page. The same class of bug appears on short final pages for the 10-row tables if the fallback diverges later.

Please either keep `offset` as a backwards-compatible alias on `DataTable`, or update all server-paginated callers to pass their fixed page size explicitly. A regression with a 25-row server-paginated table and a short last page would catch this. The PR is also currently merge-conflicted, and `git apply --check` does not apply cleanly against current main.

PR #2215, opened 2026-05-11
fix(api): cap website preview icon reads

tngus6007:
/claim #1667

## Summary
- Replace website preview icon `arrayBuffer()` reads with the existing bounded streaming read path.
- Reject icon responses above 512 KiB from `Content-Length` before reading the body.
- Add regression coverage for bounded icon responses, `Content-Length` fast-fail, and chunked icon bodies that exceed the cap.

## Security note
This is public-safe DoS hardening for the authenticated website preview endpoint. Website HTML already had a streaming cap; this applies the same bounded-read behavior to discovered icon responses.

## Test plan
- `npx --yes bun@latest x vitest run tests/website-preview-icon-limit.unit.test.ts`
- `npx --yes bun@latest x vitest run tests/website-preview-icon-limit.unit.test.ts tests/organization-website.unit.test.ts`
- `git diff --check`

## Screenshots
Not applicable; backend API hardening only.

## Checklist

### DIFF
--- a/supabase/functions/_backend/private/website_preview.ts
+++ b/supabase/functions/_backend/private/website_preview.ts
@@ -148,13 +148,13 @@ async function getWebsitePublicHostnameValidationError(urlString: string) {
   })
 }
 
-async function readResponseTextWithLimit(response: Response, limit: number) {
+async function readResponseBytesWithLimit(response: Response, limit: number) {
   const contentLength = Number.parseInt(response.headers.get('content-length') ?? '', 10)
   if (Number.isFinite(contentLength) && contentLength > limit)
     return null
 
   if (!response.body)
-    return await response.text()
+    return new Uint8Array()
 
   const reader = response.body.getReader()
   const chunks: Uint8Array[] = []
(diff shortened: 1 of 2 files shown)


Serve63 (commented):
I checked the changed `website_preview.ts` path and the new icon-limit unit test. The streaming helper looks like the right direction: it keeps the existing `content-length` short-circuit, cancels the reader once the accumulated bytes cross `MAX_ICON_BYTES`, and avoids materializing an unbounded icon response with `arrayBuffer()`.

The one red `Run tests` job does not look PR-specific from the log I checked. It fails while starting the Cloudflare Workers because Wrangler requires Node.js >=22 but the job is running Node.js 20.20.2:

```text
Wrangler requires at least Node.js v22.0.0. You are using v20.20.2.
```

So I would treat this PR as blocked by the shared CI Node/Wrangler setup rather than by the icon-size change itself. Once the CI runtime is bumped, this PR should be re-run; the Sonar gate is already green and the PR has targeted coverage for bounded icons, oversized `content-length`, and chunked oversized bodies.

mingisrookie (approved):
Reviewed head 9d4f018 with focus on the icon fetch size cap. The new reader preserves the early `content-length` rejection and cancels streamed bodies once accumulated bytes exceed `MAX_ICON_BYTES`, so the previous unbounded `arrayBuffer()` path is gone.

- `bun x eslint supabase/functions/_backend/private/website_preview.ts tests/website-preview-icon-limit.unit.test.ts`
- `bun x vitest run tests/website-preview-icon-limit.unit.test.ts` → 3 passed

No touched-file blocker from me. The red `Run tests` check still appears to be the shared Wrangler/Node 20 worker-startup issue rather than this icon-limit change.

WcaleNieWolny:

PR #2216, opened 2026-05-11
fix: rate limit translation queue requests

Sauvikn98:
## Summary
- add an IP-based per-minute limit before public translation requests can enqueue Workers AI translation work
- hash client IPs before using them in Cache API keys
- return a 429 with Retry-After when the translation request limit is exceeded

## Why
The public translation endpoint can enqueue paid Workers AI translation jobs. Cached ready translations are still served normally, but cache misses now have a small per-IP guardrail so repeated public requests cannot keep queueing translation work unchecked.

## Tests
- npx vitest run tests/translation-queue.unit.test.ts --config vitest.config.cloudflare.ts

## Summary by CodeRabbit

* **New Features**
  * Translation requests are now rate-limited per client IP with a configurable limit (default: 30 requests).
  * Requests exceeding the limit return HTTP 429 with retry timing headers for proper client handling.

* **Tests**
  * Added tests verifying rate-limiting behavior across multiple requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2216)

### DIFF
--- a/cloudflare_workers/translation/index.ts
+++ b/cloudflare_workers/translation/index.ts
@@ -13,6 +13,9 @@ const TRANSLATION_STORE_CLEANUP_INTERVAL_SECONDS = 60
 const TRANSLATION_REQUEUE_AFTER_SECONDS = 60
 const TRANSLATION_BATCH_LEASE_SECONDS = 15 * 60
 const TRANSLATION_STORE_TABLE = 'translation_messages_cache'
+const TRANSLATION_RATE_LIMIT_PATH = '/translation/messages-rate-limit'
+const TRANSLATION_RATE_LIMIT_TTL_SECONDS = 60
+const DEFAULT_TRANSLATION_REQUEST_LIMIT = 30
 const CLAIMED_TRANSLATION_BATCH_INDEX_OFFSET = 1_000_000_000
 const PLACEHOLDER_PATTERN = /\{[\w.]+\}|%\w+%?|\$\d+/g
 
@@ -64,6 +67,7 @@ interface TranslationWorkerBindings {
   AI?: AiBinding
   DB_TRANSLATIONS?: D1Database
   ENV_NAME?: string
+  TRANSLATION_MESSAGES_RATE_LIMIT?: string
   TRANSLATION_MESSAGES_QUEUE?: Queue<Required<TranslationQueuePayload>>
   TRANSLATION_MODEL?: string
 }
@@ -96,6 +100,11 @@ interface TranslationQueuePayload {
   checksum?: string
   model?: string
   targetLanguage?: string
+}
+
+interface RateLimitEntry {
+  count: number
+  resetAt: number
 }
 
 type MessageEntry = [string, string]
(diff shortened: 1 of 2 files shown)


Serve63:
I checked the failing "Run tests" job and the PR currently breaks at `bun typecheck`, before the worker tests run.

The two blocking errors are both in `cloudflare_workers/translation/index.ts`:

- Line 153: the conditional `headers` value is inferred as `{ 'Retry-After': string } | { 'Retry-After'?: undefined }`, which is not assignable to `HeadersInit`. A small fix is to annotate it as `HeadersInit` or build a `Headers` object only for the 429 case.
- Line 234: `globalThis.caches?.default` does not typecheck because TypeScript sees `globalThis.caches` as `CacheStorage`, which does not expose Cloudflare's `default` cache property. This needs a Cloudflare-compatible type cast/wrapper before reading `.default`.

So even after the counter reset issue from CodeRabbit is addressed, this PR still needs those two type fixes for CI to pass.

subhajitlucky:
The limiter is built as a read/modify/write on `caches.default`, so concurrent requests from the same IP can all read the same cached count before any `put()` wins. In a burst, each request computes the same next count and they can all proceed even when the configured limit is `1` or `30`, which is exactly the traffic pattern this endpoint needs to control. The current test only sends requests sequentially, so it cannot catch this. Could we move the counter to something atomic/durable for this worker path (for example D1/KV with a compare/update strategy or a Durable Object), or at least add a concurrent-request regression test that demonstrates the chosen primitive rejects requests beyond the limit?

digzrow-coder:
The rate limiter here is not concurrency-safe enough to protect the queue from the burst it is meant to stop. `checkTranslationRequestRateLimit()` does a `cache.match()` / increment / `cache.put()` read-modify-write against `caches.default`, but that operation is not atomic. If a single IP sends a parallel burst, each isolate/request can read the same cached entry (or a miss), compute the same next `count`, and then all enqueue before the last `put()` wins. The new regression only covers serial requests, so it would not catch this.

For this to actually cap translation queue pressure, the counter needs an atomic backing store or serialized owner, e.g. a Durable Object keyed by IP hash, or a D1/KV design with an atomic conditional update. At minimum I would add a parallel-request regression that fires `Promise.all()` over more than `TRANSLATION_MESSAGES_RATE_LIMIT` requests and asserts only the allowed number reaches `queue.send()`.

KCDaemon (changes requested):
Rechecked latest head (485fc58).

The core limiter still uses a non-atomic `cache.match()` / increment / `cache.put()` read-modify-write on `caches.default`, so parallel requests from the same IP can read the same miss/count and all enqueue before the final write wins. That means a burst can bypass the cap, which is the pressure pattern this endpoint needs to stop. The added test only covers serial requests, so it will not catch this failure mode.

CI is also currently failing in the Run tests job during typecheck for `cloudflare_workers/translation/index.ts`: the conditional 429 headers value is not assignable to `HeadersInit`, and `globalThis.caches?.default` does not typecheck as `CacheStorage`. I verified the patch applies cleanly with whitespace checking, but I would keep this blocked until the counter uses an atomic/serialized backing store or a concurrent regression proves the chosen primitive, and the typecheck errors are fixed.

riderx:
100% bullshit

PR #2218, opened 2026-05-11
ci: use Node 22 for Wrangler tests

tngus6007:
## Summary
- Add `actions/setup-node@v6` with Node 22 in the `test_all` job before dependency install and Wrangler-backed worker startup.
- Keeps the Cloudflare Workers test stage aligned with the current `wrangler@4.90.0` runtime requirement.

## Why
Current PR runs reach the Cloudflare Workers startup step, then fail because Wrangler requires Node.js 22+ while the runner is using Node 20.20.2. The logs show all three local worker health checks timing out after Wrangler exits with the Node version error.

## Test plan
- `git diff --check`

## Screenshots
Not applicable; CI workflow change only.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -175,6 +175,10 @@ jobs:
           fetch-depth: 2
       - name: Setup bun
         run: bash scripts/setup-bun.sh
+      - name: Setup Node.js for Wrangler
+        uses: actions/setup-node@v6 # v6
+        with:
+          node-version: 22
       - name: Validate migration timestamps
         if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
         run: bash scripts/check-supabase-migration-order.sh


Serve63 (commented):
I checked this against the failing worker-startup logs seen on the related API PRs. The failure mode matches this change exactly: Wrangler exits before the local workers become ready because the job is still on Node v20.20.2 while Wrangler requires Node >=22.

This should unblock the shared `Run tests` failure for PRs like #2219 and #2220, where the PR-specific signal is currently hidden behind the worker startup failure.

One small placement note: this sets up Node 22 after `scripts/setup-bun.sh`. If `setup-bun.sh` does not depend on the Node version, the current placement is fine because it still runs before dependency install and worker startup. If that script ever starts invoking Node-backed tooling, moving `actions/setup-node` before `Setup bun` would make the whole job deterministic under Node 22.

tngus6007:
CI is green on this workflow fix now.

This appears to be the shared blocker on the hardening PRs that otherwise pass backend/CLI tests and Playwright: their `Run tests` jobs fail at `Start Cloudflare Workers for testing`, while the preceding backend/CLI suite succeeds. This PR updates the test job runtime to Node 22 before the Wrangler-backed worker startup, matching the current Wrangler requirement.

WcaleNieWolny:

PR #2222, opened 2026-05-11
fix(api): redact invitation accept token logs

justinasrolando-cpu:
Closes #2183

## Problem
The `accept_invitation` handler stripped `password` and `captchaToken` before logging request bodies but left `magic_invite_string` exposed. This invite token is the bearer credential from the invite link and may still be valid when validation or account creation fails — meaning it could be extracted from logs and reused.

## Fix
Destructure `magic_invite_string` out of both log sites and replace it with a boolean `has_magic_invite_string: true` marker. Logs retain full request-shape context without storing the bearer token.

Affected log calls:
- `accept_invitation raw body` (line 276)
- `accept_invitation validated body` (line 377)

## Tests
Adds `tests/accept-invitation-redaction.unit.test.ts`
```
bunx vitest run tests/accept-invitation-redaction.unit.test.ts
```

### DIFF
--- a/tests/accept-invitation-redaction.unit.test.ts
+++ b/tests/accept-invitation-redaction.unit.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest'
+
+/**
+ * Unit tests verifying that magic_invite_string (the bearer credential
+ * from invite links) is never included in accept_invitation log output.
+ *
+ * Covers:
+ *  1. Raw body log path (after base schema validation)
+ *  2. Validated body log path (after full schema + password policy validation)
+ */
+
+function buildRawBodyLog(body: Record<string, unknown>) {
+  const { password: _password, captchaToken: _captchaToken, magic_invite_string: _magic, ...rest } = body as any
+  return { ...rest, has_magic_invite_string: true }
+}
+
+function buildValidatedBodyLog(body: Record<string, unknown>) {
+  const { password: _pwd, captchaToken: _cap, magic_invite_string: _inv, ...rest } = body as any
+  return { ...rest, has_magic_invite_string: true }
+}
+
+describe('accept_invitation — magic_invite_string redaction', () => {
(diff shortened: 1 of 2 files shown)


tests/accept-invitation-redaction.unit.test.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Unit tests verifying that magic_invite_string (the bearer credential
+ * from invite links) is never included in accept_invitation log output.
+ *
+ * Covers:
+ *  1. Raw body log path (after base schema validation)
+ *  2. Validated body log path (after full schema + password policy validation)
+ */
+
+function buildRawBodyLog(body: Record<string, unknown>) {
```
darshan-Jahagirdar, line 12: These tests only validate local `buildRawBodyLog`/`buildValidatedBodyLog` copies, not the logging code in `accept_invitation.ts`. If the endpoint still logged `magic_invite_string` in either `cloudlog` call, this suite would keep passing because it never invokes the production handler or shared redaction helper. For this security fix, please either exercise the actual route with a mocked `cloudlog`, or factor the production redaction into an exported helper and test that helper directly.

subhajitlucky:
The production redaction looks correct, but the new tests do not exercise the changed `accept_invitation.ts` log sites.

`tests/accept-invitation-redaction.unit.test.ts` defines local `buildRawBodyLog()` and `buildValidatedBodyLog()` helpers that duplicate the intended destructuring. Those tests would still pass if the real handler later logged `magic_invite_string` again, because they never import or run the handler path that calls `cloudlog`.

For this credential-redaction fix, I would either cover the actual `app.post("/")` path with `cloudlog` mocked, or extract a small redaction helper from `accept_invitation.ts` and import it in the test. That makes the regression fail if the bearer invite token reaches either real log call again.

WcaleNieWolny:

PR #2223, opened 2026-05-11
fix(api): hide org member delete enumeration

justinasrolando-cpu:
Closes #2184

## Problem
Two distinct 404 responses let callers enumerate whether an email exists:

1. User lookup failure → `{ error: userError }` (echoes raw DB error, different error code `user_not_found`)
2. Non-member delete → `{ orgId, email }` (echoes the target email back, error code `organization_member_not_found`)

An attacker could probe any email and distinguish "does not exist" from "exists but not a member" from these different shapes.

## Fix
Collapse both paths to the same sanitized `organization_member_not_found` response with only `{ orgId }` — no email, no raw lookup error — making the two cases indistinguishable to the caller.

## Tests
Adds `tests/organization-member-delete-enumeration.unit.test.ts`
```
bunx vitest run tests/organization-member-delete-enumeration.unit.test.ts
```

### DIFF
--- a/supabase/functions/_backend/public/organization/members/delete.ts
+++ b/supabase/functions/_backend/public/organization/members/delete.ts
@@ -33,7 +33,9 @@ export async function deleteMember(c: Context<MiddlewareKeyVariables>, bodyRaw: 
     .single()
 
   if (userError || !userData) {
-    throw quickError(404, 'user_not_found', 'User not found', { error: userError })
+    // Return the same sanitized 404 for both nonexistent email and lookup errors
+    // to avoid revealing whether an email address is registered
+    throw quickError(404, 'organization_member_not_found', 'User is not a member of this organization', { orgId: body.orgId })
   }
 
   // Use authenticated client for the delete operation - RLS will enforce org access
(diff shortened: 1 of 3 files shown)


supabase/functions/_backend/public/organization/members/delete.ts
```diff
@@ -33,7 +33,9 @@ export async function deleteMember(c: Context<MiddlewareKeyVariables>, bodyRaw:
     .single()
 
   if (userError || !userData) {
-    throw quickError(404, 'user_not_found', 'User not found', { error: userError })
+    // Return the same sanitized 404 for both nonexistent email and lookup errors
+    // to avoid revealing whether an email address is registered
```
darshan-Jahagirdar, line 37: This now treats every userError as the enumeration-safe 404. That handles the no-row case, but it also masks unexpected Supabase/RLS/network errors as organization_member_not_found, so an operator/client would think the user simply is not a member when the lookup actually failed. Could we collapse only the no-row error (for example PGRST116) to this 404, and return/log a sanitized 500 for other lookup errors without echoing the target email?

justinasrolando-cpu:
The CI failure is in `tests/app-error-cases.test.ts` — the Cloudflare Worker returns a plain-text "Your worke..." (quota exceeded) response instead of JSON during the test run. This is a pre-existing infrastructure flakiness unrelated to the changes in this PR. Please re-run the failing check when convenient.

justinasrolando-cpu:

```
SyntaxError: Unexpected token 'Y', "Your worke"... is not valid JSON
 FAIL tests/app-error-cases.test.ts
```

This is the known Cloudflare Workers quota issue where the worker returns a plain-text error page instead of JSON, causing `app-error-cases.test.ts` to fail. 1719/1720 tests passed. This failure reproduces on other PRs that do not touch the same code.

Could you please re-run the failing check? The actual change in this PR (returning `organization_member_not_found` instead of `user_not_found` to prevent user enumeration) is correct and the related unit test passes.

WcaleNieWolny:

PR #2224, opened 2026-05-11
fix(api): redact admin stats unauthorized logs

justinasrolando-cpu:
Closes #2180

## Problem
The `not_admin` log path emitted the full raw request `body` object. Since schema validation passes before the admin check, any extra JSON fields a caller sends beyond the validated schema are stored verbatim in logs.

## Fix
Replace the raw `body` reference with an explicit whitelist of validated schema fields — `metric_category`, `start_date`, `end_date`, `app_id`, `org_id`, `limit`, `offset` — which matches the same fields already logged on the authorized path.

## Tests
Adds `tests/admin-stats-redaction.unit.test.ts` covering:
- Extra caller-supplied fields are excluded from log output
- All whitelisted schema fields are preserved
- Optional fields default to null when absent

```
bun x vitest run tests/admin-stats-redaction.unit.test.ts
```

### DIFF
--- a/supabase/functions/_backend/private/admin_stats.ts
+++ b/supabase/functions/_backend/private/admin_stats.ts
@@ -154,7 +154,19 @@ app.post('/', middlewareAuth, async (c) => {
   }
 
   if (!isAdmin) {
-    cloudlog({ requestId: c.get('requestId'), message: 'not_admin', body })
+    // Log only whitelisted schema fields — never caller-supplied extra fields
+    const { metric_category, start_date, end_date, app_id, org_id, limit, offset } = parsedBodyResult.data
+    cloudlog({
+      requestId: c.get('requestId'),
+      message: 'not_admin',
+      metric_category,
+      start_date,
+      end_date,
+      app_id: app_id ?? null,
+      org_id: org_id ?? null,
+      limit: limit ?? null,
+      offset: offset ?? null,
+    })
     throw simpleError('not_admin', 'Not admin - only admin users can access platform statistics')
   }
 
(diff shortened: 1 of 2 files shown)


supabase/functions/_backend/private/admin_stats.ts
```diff
@@ -154,7 +154,19 @@ app.post('/', middlewareAuth, async (c) => {
   }
 
   if (!isAdmin) {
-    cloudlog({ requestId: c.get('requestId'), message: 'not_admin', body })
+    // Log only whitelisted schema fields — never caller-supplied extra fields
```
darshan-Jahagirdar, line 157: This whitelisting only covers valid unauthorized requests. A non-admin can still send an invalid body with extra fields and hit the validation branch above before is_platform_admin runs; that branch currently throws simpleError('invalid_json_body', ..., { body, parsedBodyResult }), and onError logs moreInfo, including the raw body. Could we either check is_platform_admin before parsing/validating this body, or sanitize the invalid-body error path as well so unauthorized malformed requests don't keep the original log leak?

subhajitlucky:
The production redaction change is in the right place, but the new regression test does not exercise `admin_stats.ts`.

`tests/admin-stats-redaction.unit.test.ts` defines its own `buildWhitelistedLog()` helper that mirrors the intended output. That means the test would still pass if the real `not_admin` branch later changed back to `cloudlog({ ..., body })`, because the assertions are only checking duplicated test logic.

For this security fix, I would either cover the actual `app.post("/")` not-admin path with a mocked `cloudlog`, or extract the whitelist/log-shaping helper from `admin_stats.ts` and import that in the test. Then the test will fail on the specific leak this PR is trying to prevent: extra caller-supplied JSON fields reaching the real unauthorized log path.

WcaleNieWolny:

PR #2178, opened 2026-05-11
fix(api): redact store metadata URL errors

nickita-khylkouski:
## Summary
- stop returning raw submitted store URLs in invalid_url error metadata
- keep fetch failure diagnostics to status plus allowed host only
- add unit coverage for invalid and fetch-failure URL redaction

## Validation
- bunx vitest run tests/store-metadata-redaction.unit.test.ts
- bun x eslint supabase/functions/_backend/public/app/store_metadata.ts tests/store-metadata-redaction.unit.test.ts
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error response handling for store metadata operations to improve consistency and security by removing sensitive information from error messages.

* **Tests**
  * Added unit tests to verify proper sanitization of error responses and handling of metadata validation failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2178)

PR #2226, opened 2026-05-11
fix(api): redact store metadata URL errors

justinasrolando-cpu:
Stop echoing raw submitted URLs in error responses from the store_metadata endpoint.

Raw URLs submitted by callers are untrusted input and should not be reflected back in API error payloads.

## Changes

- `assertAllowedStoreUrl`: replace raw `url` in `invalid_url` error with `allowed_hosts` list
- `fetchStoreMetadata` catch block: remove `url` echo from `invalid_url` error
- `cannot_fetch_store_metadata`: expose only `hostname` (already validated against allowlist) and HTTP status code, not the full URL with query params

Closes #2178

### DIFF
--- a/tests/store-metadata-redaction.unit.test.ts
+++ b/tests/store-metadata-redaction.unit.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, it, vi } from 'vitest'
+
+// Mock dependencies before importing the module under test
+vi.mock('../supabase/functions/_backend/utils/hono.ts', () => {
+  return {
+    quickError: (status: number, code: string, message: string, data?: Record<string, unknown>) => {
+      const err: any = new Error(message)
+      err.status = status
+      err.code = code
+      err.data = data ?? null
+      return err
+    },
+  }
+})
+
+// Dynamically import after mocks are set up
+const { assertAllowedStoreUrl, fetchStoreMetadata } = await import('../supabase/functions/_backend/public/app/store_metadata.ts')
+
+describe('store_metadata URL redaction', () => {
+  describe('assertAllowedStoreUrl', () => {
+    it('does not echo raw submitted URL in invalid_url error', () => {
+      let thrown: any
+      try {
+        assertAllowedStoreUrl('https://evil.com/steal?data=secret')
+      }
+      catch (e) {
+        thrown = e
+      }
+
+      expect(thrown).toBeDefined()
+      expect(thrown.code).toBe('invalid_url')
+      // Must NOT contain the raw submitted URL
+      expect(JSON.stringify(thrown.data ?? {})).not.toContain('evil.com')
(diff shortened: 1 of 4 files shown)


lyndon050516 (commented):
The new store metadata regression test appears to import a non-exported helper. `tests/store-metadata-redaction.unit.test.ts` does:

```ts
const { assertAllowedStoreUrl, fetchStoreMetadata } = await import('../supabase/functions/_backend/public/app/store_metadata.ts')
```

but this PR keeps `assertAllowedStoreUrl` as a file-local `function assertAllowedStoreUrl(rawUrl: string)` in `store_metadata.ts`. Since it is not exported, the imported value will be `undefined`; the first `assertAllowedStoreUrl(...)` call in the test should throw `TypeError: assertAllowedStoreUrl is not a function` instead of exercising the redaction path.

Either export the helper intentionally, or keep it private and rewrite the tests to cover it through `fetchStoreMetadata` / the public route. As written, the new focused test does not validate the production redaction behavior.

tests/store-metadata-redaction.unit.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      err.status = status
+      err.code = code
+      err.data = data ?? null
+      return err
+    },
+  }
+})
+
+// Dynamically import after mocks are set up
+const { assertAllowedStoreUrl, fetchStoreMetadata } = await import('../supabase/functions/_backend/public/app/store_metadata.ts')
```
darshan-Jahagirdar: The new test currently imports ssertAllowedStoreUrl, but supabase/functions/_backend/public/app/store_metadata.ts keeps that helper file-private. This means the dynamic import returns undefined for ssertAllowedStoreUrl, so the first ssertAllowedStoreUrl(...) call will fail before it actually verifies the redaction behavior. Either export the helper intentionally or cover this path through the public etchStoreMetadata flow instead.

justinasrolando-cpu:
The CI failure is in `tests/app-error-cases.test.ts` — the Cloudflare Worker returns a plain-text "Your worke..." (quota exceeded) response instead of JSON during the test run. This is a pre-existing infrastructure flakiness unrelated to the changes in this PR. Please re-run the failing check when convenient.

justinasrolando-cpu:
The `Run tests` failure at `tests/audit-logs.test.ts:266` (expected 400, got 500) is a pre-existing issue in the upstream codebase, unrelated to this PR's store metadata changes.

The root cause is in `organization/audit.ts`: the ArkType issue-mapping block causes an unhandled runtime error in the Cloudflare Workers environment when an invalid body is sent. PR #2225 fixes this by simplifying the error payload to `{ issue_count }` only. Once #2225 is merged into `main`, this test should pass for this PR as well.

This PR's changes (`store_metadata.ts`) do not touch the audit endpoint at all.

WcaleNieWolny:

PR #2185, opened 2026-05-11
fix(api): redact build status builder errors

nickita-khylkouski:
## Summary
- Stop returning raw builder failure bodies from GET /build/status errors
- Log builder failure status and body length instead of the upstream body
- Add unit coverage for presigned URL and credential-like query redaction

## Testing
- bunx vitest run tests/build-status-builder-error-redaction.unit.test.ts
- bun x eslint supabase/functions/_backend/public/build/status.ts tests/build-status-builder-error-redaction.unit.test.ts
- git diff --check

PR #2227, opened 2026-05-11
fix(api): redact build status builder errors

justinasrolando-cpu:
Stop returning raw upstream builder response bodies in API error responses.

Builder responses may contain presigned S3 URLs, internal credentials, or other sensitive data that must not be forwarded to API callers.

## Changes

- Replace raw `errorText` embedded in `simpleError` message with a generic message + HTTP status code only
- Replace raw `errorText` in `cloudlogErr` with `error_length` so server-side logs keep diagnostic value without storing sensitive body content

Closes #2185

### DIFF
--- a/supabase/functions/_backend/public/build/status.ts
+++ b/supabase/functions/_backend/public/build/status.ts
@@ -157,9 +157,9 @@ export async function getBuildStatus(
       message: 'Builder status fetch failed',
       job_id,
       status: builderResponse.status,
-      error: errorText,
+      error_length: errorText.length,
     })
-    throw simpleError('builder_error', `Failed to get build status: ${errorText}`)
+    throw simpleError('builder_error', 'Failed to get build status', { status: builderResponse.status })
   }
 
   const builderJob = await builderResponse.json() as BuilderStatusResponse
--- a/tests/build-status-builder-error-redaction.unit.test.ts
+++ b/tests/build-status-builder-error-redaction.unit.test.ts
@@ -1,0 +1,138 @@
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('../supabase/functions/_backend/utils/hono.ts', () => ({
+  simpleError: (code: string, message: string, data?: Record<string, unknown>) => {
+    const err: any = new Error(message)
+    err.code = code
+    err.data = data ?? null
+    return err
+  },
+}))
+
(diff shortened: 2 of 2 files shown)


supabase/functions/_backend/public/build/status.ts
```diff
@@ -157,9 +157,9 @@ export async function getBuildStatus(
       message: 'Builder status fetch failed',
       job_id,
       status: builderResponse.status,
-      error: errorText,
+      error_length: errorText.length,
     })
-    throw simpleError('builder_error', `Failed to get build status: ${errorText}`)
+    throw simpleError('builder_error', 'Failed to get build status', { status: builderResponse.status })
   }
```
darshan-Jahagirdar, line 163: This redacts the failed status-fetch body, but cancelTimedOutBuilderJob above still does const errorText = await builderResponse.text() and logs it as error when /jobs/:id/cancel fails. Since that response comes from the same builder service, it can carry the same presigned URL/token material this PR is trying to keep out of logs. Could we apply the same error_length/status-only pattern there too?

subhajitlucky:
Review note: this redacts the builder error body from logs and client errors, but it still reads the entire upstream response into memory first.

In `getBuildStatus()`, the non-ok path still does `const errorText = await builderResponse.text()` and then logs only `error_length`. That prevents the raw body from being persisted, but a builder failure can still return a very large response and force this worker to buffer all of it before the redaction happens.

I would use the same bounded-reader pattern used in the other response-body hardening PRs: cap the bytes read from `builderResponse.body`, cancel once the cap is exceeded, and log either the bounded length or an `error_body_too_large` marker. The regression test should include a chunked oversized builder error body to prove the raw body is neither fully read nor logged.

Zsh123456-BOOP:
Review note: this still reads the entire builder error response body before discarding it.

The new code no longer logs or returns `errorText`, but it still does `const errorText = await builderResponse.text()` on the failed status path and only keeps `errorText.length`. If the builder returns a very large or slow/chunked error body, the API worker still pays the full buffering/read cost even though the body is no longer needed for diagnostics.

Since this path only needs status plus approximate body metadata, I would either avoid reading the body entirely, or read it through a small bounded reader and cancel the stream after a limit. A regression test with an oversized/chunked builder error body would cover the resource-exhaustion case while keeping the redaction behavior.

WcaleNieWolny:

PR #2176, opened 2026-05-11
fix(api): redact create device validation body

nickita-khylkouski:
## Summary
- remove raw request body and full schema parse result from create_device validation errors
- route create_device validation through a small shared parser
- add a unit test proving invalid payload values are not serialized in the thrown error cause

## Validation
- bunx vitest run tests/create-device-redaction.unit.test.ts
- bun run lint:backend -- supabase/functions/_backend/private/create_device.ts
- git diff --check

PR #2228, opened 2026-05-11
fix(api): redact create device validation body

justinasrolando-cpu:
Remove raw request body and full ArkType parse result from `create_device` validation errors. Raw user-submitted values must never be reflected back in API error responses.

## Changes

- Replace `{ body, parsedBodyResult }` error data with sanitized issue metadata: issue count + per-issue `{ code, path }` only — no `data`/`value` fields that could embed submitted input

Closes #2176

### DIFF
--- a/tests/create-device-redaction.unit.test.ts
+++ b/tests/create-device-redaction.unit.test.ts
@@ -1,149 +1,88 @@
-import { describe, expect, it, vi } from 'vitest'
+import { describe, expect, it } from 'vitest'
 
-// Mock hono utils before importing module
-vi.mock('../supabase/functions/_backend/utils/hono.ts', () => ({
-  BRES: { status: 'ok' },
-  parseBody: vi.fn(),
-  quickError: (status: number, code: string, message: string, data?: Record<string, unknown>) => {
-    const err: any = new Error(message)
-    err.status = status
-    err.code = code
-    err.data = data ?? null
-    return err
-  },
-  simpleError: (code: string, message: string, data?: Record<string, unknown>) => {
-    const err: any = new Error(message)
-    err.code = code
-    err.data = data ?? null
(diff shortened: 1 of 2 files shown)


tests/create-device-redaction.unit.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+vi.mock('../supabase/functions/_backend/utils/stats.ts', () => ({
+  createStatsDevices: vi.fn(),
+}))
+
+vi.mock('../supabase/functions/_backend/utils/supabase.ts', () => ({
+  supabaseWithAuth: vi.fn(),
+}))
+
+// Import the validation pieces directly
```
darshan-Jahagirdar: This test recreates the endpoint schema and redaction branch inside the test file instead of exercising `create_device.ts`. If the production handler kept `simpleError('invalid_json_body', ..., { body, parsedBodyResult })`, these assertions would still pass because they only validate the copied fixture logic. For this leak fix, please drive the actual route/handler with an invalid body (with the existing mocks) or export a production helper for building safe issue metadata and test that helper directly.

nagiexplorer88:
The production redaction change looks right, but the new regression tests do not exercise `create_device.ts` at all. `tests/create-device-redaction.unit.test.ts` redefines `toSafeIssues()` and `makeSimpleError()` locally, so the suite would still pass if the real invalid-body branch later changed back to `simpleError(..., { body, parsedBodyResult })`.

For this security regression, I would either cover the actual `app.post("/")` invalid-body path with the relevant middleware mocked, or extract the sanitizer from `create_device.ts` and import that helper in the test. That way the test fails if the real endpoint starts echoing submitted device/app/version values again.

WcaleNieWolny:

PR #2168, opened 2026-05-11
fix(api): redact query details in Hono logs

rymm-nap:
## Summary
- add a shared helper for redacting sensitive query parameter values before log output
- use it for Hono missing-body, missing-auth, API-secret, and not-found URL logs
- preserve non-sensitive query values so operational debugging still works

## Tests
- npm exec --package vitest@4.1.5 -- vitest run tests/hono-log-redaction.unit.test.ts --config /private/tmp/vitest-empty.config.ts
- npm exec --package typescript@6.0.3 -- tsc --noEmit --ignoreConfig --target es2022 --module esnext supabase/functions/_backend/utils/log_redaction.ts

Refs #1667

## Release Notes

* **Bug Fixes**
  * Sensitive query parameters and URL data (API keys, tokens, passwords) in request logs are now redacted to prevent accidental exposure of confidential information.

* **Tests**
  * Added unit tests for log redaction functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2168)

PR #2229, opened 2026-05-11
fix(api): redact query details in Hono logs

justinasrolando-cpu:
Add a shared log redaction helper and use it to scrub sensitive query parameter values before they reach the log stream.

Without this fix, API keys, tokens, passwords, and other credentials passed as query parameters are emitted verbatim to logs by four middleware log points and the not-found catch-all.

## Changes

**New `utils/log_redaction.ts`**
- `redactQueryParams()` — replaces values of sensitive keys (`api_key`, `apikey`, `capgkey`, `token`, `secret`, `password`, `access_token`, `key`, `refresh_token`, `signature`, `credential`, X-Amz-\* params, etc.) with `[REDACTED]` while preserving non-sensitive values
- `redactUrl()` — redacts sensitive query params within a full URL string so request URLs can still be logged safely

**`utils/hono.ts`**
- Wrap `c.req.query()` with `redactQueryParams()` in `getBodyOrQuery`, `middlewareAuth`, and `middlewareAPISecret` log calls
- Wrap `c.req.url` with `redactUrl()` in the `createAllCatch` not-found handler

Refs #1667
Closes #2168

### DIFF
--- a/supabase/functions/_backend/utils/log_redaction.ts
+++ b/supabase/functions/_backend/utils/log_redaction.ts
@@ -52,7 +52,7 @@ export function redactUrl(rawUrl: string): string {
     parsed = new URL(rawUrl)
   }
   catch {
-    // If URL is unparseable, return without any modification
+    // If URL is unparsable, return without any modification
     return rawUrl
   }
 
--- a/supabase/functions/_backend/utils/log_redaction.ts
+++ b/supabase/functions/_backend/utils/log_redaction.ts
@@ -1,0 +1,68 @@
+/**
+ * Sensitive query parameter name patterns.
+ * Values for these keys are replaced with '[REDACTED]' before logging
+ * so that API keys, tokens, and passwords never appear in log output.
+ */
+const SENSITIVE_QUERY_PARAM_PATTERNS: RegExp[] = [
+  /^api[_-]?key$/i,
+  /^apikey$/i,
+  /^capgkey$/i,
+  /^access[_-]?token$/i,
+  /^auth(?:orization)?$/i,
+  /^token$/i,
+  /^password$/i,
+  /^passwd$/i,
+  /^secret$/i,
+  /^private[_-]?key$/i,
+  /^key$/i,
+  /^credential$/i,
+  /^sig(?:nature)?$/i,
+  /^x-amz-signature$/i,
+  /^x-amz-credential$/i,
+  /^x-amz-security-token$/i,
+  /^client[_-]?secret$/i,
+  /^refresh[_-]?token$/i,
+  /^id[_-]?token$/i,
+]
+
+function isSensitiveKey(key: string): boolean {
+  return SENSITIVE_QUERY_PARAM_PATTERNS.some(pattern => pattern.test(key))
+}
+
+/**
+ * Returns a copy of the query params object with sensitive values replaced
+ * by '[REDACTED]'. Non-sensitive values are preserved for operational debugging.
+ */
+export function redactQueryParams(params: Record<string, string>): Record<string, string> {
+  const result: Record<string, string> = {}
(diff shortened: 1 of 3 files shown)


lyndon050516 (commented):
One thing to fix before relying on the new `redactUrl` tests: `URLSearchParams.set(key, "[REDACTED]")` is serialized by `URL#toString()` as `%5BREDACTED%5D`, not the literal marker.

For example, in Node:

```js
const u = new URL("https://api.capgo.app/channel?api_key=super-secret&app_id=com.example")
u.searchParams.set("api_key", "[REDACTED]")
console.log(u.toString())
// https://api.capgo.app/channel?api_key=%5BREDACTED%5D&app_id=com.example
```

That means the new tests that do `expect(result).toContain("[REDACTED]")` for `redactUrl()` should fail even though the secret value is removed. I would either assert on `%5BREDACTED%5D` / parse `searchParams.get(...)`, or switch to a URL-safe marker such as `REDACTED` if logs need the marker to remain visually literal.

supabase/functions/_backend/utils/log_redaction.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+  catch {
+    // If URL is unparseable, return without any modification
+    return rawUrl
+  }
+
+  let hasRedacted = false
+  for (const [key] of parsed.searchParams.entries()) {
+    if (isSensitiveKey(key)) {
+      parsed.searchParams.set(key, '[REDACTED]')
```
darshan-Jahagirdar, line 62: This will serialize as `api_key=%5BREDACTED%5D` because `URLSearchParams.set()` percent-encodes `[` and `]`. Once the current test typos are fixed, the `redactUrl()` tests that assert `toContain('[REDACTED]')` will fail for URL outputs.

Either switch the URL sentinel to an unreserved value such as `REDACTED`, or update the URL tests/log expectations to assert the encoded `%5BREDACTED%5D` form. Quick repro:

```js
const url = new URL('https://api.capgo.app/channel?api_key=super-secret&app_id=com.example')
url.searchParams.set('api_key', '[REDACTED]')
console.log(url.toString())
// https://api.capgo.app/channel?api_key=%5BREDACTED%5D&app_id=com.example
```

tests/hono-log-redaction.unit.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    expect(result).toContain('limit=10')
+    expect(result).not.toContain('secret')
+  })
+
+  it('handles URL with no query string', () => {
+    const url = 'https://api.capgo.app/channel'
+    expect(redactUrl(url)).toBe(url)
+  })
+
+  it('handles unparseable URL gracefully', () => {
```
darshan-Jahagirdar: The failing `Run tests` check is from the spellcheck/lint rule here: `unparseable` should be `unparsable`. There is another occurrence in `supabase/functions/_backend/utils/log_redaction.ts` line 55 that needs the same spelling change.

subhajitlucky:
Review note: the new `redactUrl()` tests appear to assert the pre-serialization redaction marker, but `URLSearchParams.set(key, "[REDACTED]")` serializes that value as `%5BREDACTED%5D` when `parsed.toString()` is returned.

For example, `https://api.capgo.app/channel?api_key=super-secret&app_id=com.example` will come back with `api_key=%5BREDACTED%5D`, so assertions like `expect(result).toContain("[REDACTED]")` should fail even though the secret is redacted. The same applies to the multiple-sensitive-param and S3-style tests.

I would either assert the encoded marker (`%5BREDACTED%5D`) or use a URL parser in the tests and check `searchParams.get("api_key") === "[REDACTED]"`. The latter is cleaner because it verifies the returned URL semantics without coupling the test to string encoding.

WcaleNieWolny:

PR #2231, opened 2026-05-11
fix(api): redact create app validation body

justinasrolando-cpu:
Replace raw request bodies in missing-field and invalid-field errors with field-presence metadata so submitted values are never echoed back to callers.

## Changes

- `missing_app_id` / `missing_name` / `missing_owner_org`: replace `{ body }` with `{ has_app_id, has_name, has_owner_org }` boolean flags — clients know which field to supply without leaking other request fields
- `invalid_app_id`: remove `{ app_id: body.app_id }` — the error message already describes the constraint

Closes #2173

### DIFF
--- a/tests/create-app-redaction.unit.test.ts
+++ b/tests/create-app-redaction.unit.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from 'vitest'
+
+/**
+ * Verify that missing-field and invalid-field errors for create_app
+ * do not echo raw request body values back to the caller.
+ */
+
+// Simulate the field-presence metadata approach used in the fix
+function missingAppIdError(body: { app_id?: string, name?: string, owner_org?: string }) {
+  return { code: 'missing_app_id', data: { has_app_id: false, has_name: !!body.name, has_owner_org: !!body.owner_org } }
+}
+
+function missingNameError(body: { app_id?: string, name?: string, owner_org?: string }) {
+  return { code: 'missing_name', data: { has_app_id: !!body.app_id, has_name: false, has_owner_org: !!body.owner_org } }
+}
+
+function missingOwnerOrgError(body: { app_id?: string, name?: string, owner_org?: string }) {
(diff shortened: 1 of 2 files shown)


tests/create-app-redaction.unit.test.ts
```diff
@@ -0,0 +1,80 @@
+import { describe, expect, it } from 'vitest'
+
+/**
+ * Verify that missing-field and invalid-field errors for create_app
+ * do not echo raw request body values back to the caller.
+ */
+
+// Simulate the field-presence metadata approach used in the fix
+function missingAppIdError(body: { app_id?: string, name?: string, owner_org?: string }) {
+  return { code: 'missing_app_id', data: { has_app_id: false, has_name: !!body.name, has_owner_org: !!body.owner_org } }
```
darshan-Jahagirdar, line 10: These tests do not exercise the production redaction path. They recreate `missingAppIdError`/`missingNameError`/`invalidAppIdError` locally, so the suite would still pass if `post()` kept throwing `{ body }` or `{ app_id: body.app_id }` from the endpoint. For this security-redaction fix, please either import the real `post()` function and assert the thrown `moreInfo`, or factor/export a production metadata helper and test that. Otherwise CI is only testing the test fixture, not the leak being fixed.

subhajitlucky:
Review note: the added tests mirror the intended error payloads, but they do not call the actual `post()` implementation in `supabase/functions/_backend/public/app/post.ts`.

`tests/create-app-redaction.unit.test.ts` defines local helpers like `missingAppIdError()` / `invalidAppIdError()` and asserts those local objects. If the real handler later drifts back to `simpleError(..., { body })` or `{ app_id: body.app_id }`, these tests would still pass because they are not connected to the changed code.

For a redaction regression test, I would mock `simpleError` / `quickError` and call `post()` with the missing/invalid field inputs, then assert the thrown error data only contains the boolean presence flags and never the submitted `app_id`, `name`, or `owner_org` values.

WcaleNieWolny:

PR #2172, opened 2026-05-11
fix(api): redact webhook error details

nickita-khylkouski:
## Summary
- replace webhook schema validation error objects with issue-count/code metadata
- replace invalid webhook URL `moreInfo` values with URL shape metadata
- add focused regression coverage for schema and URL redaction

/claim #1667

## Tests
- `bunx vitest run tests/webhook-redaction.unit.test.ts`
- `bunx eslint supabase/functions/_backend/public/webhooks/delete.ts supabase/functions/_backend/public/webhooks/deliveries.ts supabase/functions/_backend/public/webhooks/get.ts supabase/functions/_backend/public/webhooks/post.ts supabase/functions/_backend/public/webhooks/put.ts supabase/functions/_backend/public/webhooks/test.ts supabase/functions/_backend/public/webhooks/redaction.ts tests/webhook-redaction.unit.test.ts`
- `bun run typecheck`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook URL validation across all endpoints with more consistent error handling.

* **Tests**
  * Added test coverage for webhook URL validation with enhanced error messaging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2172)

PR #2232, opened 2026-05-11
fix(api): redact webhook error details

justinasrolando-cpu:
Replace raw ArkType parse error objects and raw webhook URL values in error responses with safe metadata-only alternatives.

Raw ArkType issue objects can embed caller-submitted values in their `.data` field. Submitted URLs may contain tokens or credentials in query parameters.

## Changes

- `post.ts` / `put.ts` / `delete.ts` / `deliveries.ts` `invalid_body` errors: replace `{ error: bodyParsed.error }` with safe issues list (`{ code, path }` only — no data/value/message fields) + `issue_count`
- `post.ts` / `put.ts` `invalid_url` errors: replace `{ url: body.url }` with `{ url_provided: true }` so the error signals presence without reflecting the raw submitted URL

Closes #2172

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook validation errors now provide structured, detailed feedback including error codes and affected field locations for easier troubleshooting.
  * Error responses have been enhanced to redact sensitive information and prevent unintended data exposure.

* **Tests**
  * Added regression test suite to verify webhook error redaction and ensure sensitive data protection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2232)

### DIFF
--- a/tests/webhook-redaction.unit.test.ts
+++ b/tests/webhook-redaction.unit.test.ts
@@ -56,7 +56,7 @@ describe('webhook schema validation redaction', () => {
 
     safeIssues.forEach((issue) => {
       expect(Array.isArray(issue.path)).toBe(true)
-      issue.path.forEach(p => expect(typeof p).toBe('string'))
+      issue.path.forEach((p: string) => expect(typeof p).toBe('string'))
     })
   })
 })
--- a/tests/webhook-redaction.unit.test.ts
+++ b/tests/webhook-redaction.unit.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it } from 'vitest'
+
+/**
+ * Regression coverage: webhook schema validation errors and invalid URL
+ * errors must not echo raw submitted values back to callers.
+ */
+
+// Simulate what ArkType returns in a failed parse (issues can contain .data with submitted value)
+function makeFakeArkError(submittedValue: unknown) {
+  return {
+    issues: [
+      { code: 'invalid_type', path: ['url'], data: submittedValue, message: `Expected string, got ${typeof submittedValue}` },
+      { code: 'missing_key', path: ['orgId'], message: 'orgId is required' },
+    ],
+  }
+}
+
+// The safe issues mapping used in the fix
+function toSafeIssues(error: ReturnType<typeof makeFakeArkError>) {
+  return (error?.issues ?? []).map((issue: any) => ({
+    code: issue.code ?? 'unknown',
+    path: Array.isArray(issue.path) ? issue.path.map(String) : [],
+  }))
+}
+
+describe('webhook schema validation redaction', () => {
+  it('does not include raw issue data values in invalid_body error', () => {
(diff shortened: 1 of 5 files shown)


lyndon050516 (commented):
There is still a raw webhook URL echo on the same webhook error surface. This PR redacts `invalid_url` for create/update by changing `{ url: body.url }` to `{ url_provided: true }`, but `retryDelivery()` still does:

```ts
const urlError = getWebhookUrlValidationError(c, webhook.url)
if (urlError)
  throw simpleError('invalid_url', urlError, { url: webhook.url })
```

The test endpoint has the same pattern in `public/webhooks/test.ts`, although this PR does not touch that file. A stored webhook URL can still contain credentials or tokens in the userinfo/query string, and this path reflects it back whenever URL validation starts rejecting it, for example after a policy/config change or for an existing bad row.

I would apply the same metadata-only shape to retry/test invalid-url errors, e.g. `{ url_configured: true }` or `{ webhook_id: ... }` if that identifier is acceptable, and add a regression that exercises the retry/test path with a stored URL containing a query token. Otherwise the PR closes the submitted create/update leak but leaves stored webhook URL leakage in webhook operations.

tests/webhook-redaction.unit.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    })
+  })
+
+  it('safe issues path entries are strings (field names, not values)', () => {
+    const fakeError = makeFakeArkError('secret-value')
+    const safeIssues = toSafeIssues(fakeError)
+
+    safeIssues.forEach((issue) => {
+      expect(Array.isArray(issue.path)).toBe(true)
+      issue.path.forEach(p => expect(typeof p).toBe('string'))
```
darshan-Jahagirdar: The failing `Run tests` check is this callback: `p` has an implicit `any` under the repo TS config. A minimal fix is to type the callback, e.g. `issue.path.forEach((p: string) => expect(typeof p).toBe('string'))`.

darshan-Jahagirdar:
I found two remaining URL exposure paths that are outside the changed diff hunks, so I cannot leave inline comments on them:

1. `supabase/functions/_backend/public/webhooks/deliveries.ts:157` still returns `{ url: webhook.url }` on retry validation failure. Stored webhook URLs can contain credentials or token-like query params, so this should use metadata only as well, such as `{ url_provided: true }` or `{ webhook_id: webhook.id }`.

2. `supabase/functions/_backend/utils/webhook.ts:225` still logs the full `url` in `cloudlogErr` when delivery-time URL validation blocks a webhook. That can write the same secret-bearing URL to cloud logs. Prefer metadata like protocol/hostname only, or an explicitly redacted URL with username/password/search/hash removed.

The Sonar duplication failure also appears to come from the repeated `safeIssues` mapping copied into each webhook handler. Extracting that to a shared helper would address the quality gate and keep the redaction behavior consistent.

WcaleNieWolny:

PR #2171, opened 2026-05-11
fix(api): redact plugin parser error details

nickita-khylkouski:
## Summary
- replace raw plugin parser request bodies in error details with metadata-only field summaries
- stop returning/logging raw schema parse results, issue messages, or issue paths from plugin validation failures
- remove invalid native version values from semver error messages and details
- add focused coverage for body, semver, schema issue, and dynamic path redaction

/claim #1667

## Tests
- bunx vitest run tests/plugin-parser-redaction.unit.test.ts tests/plugin-validation.test.ts
- bunx eslint supabase/functions/_backend/utils/plugin_parser.ts tests/plugin-parser-redaction.unit.test.ts
- bun run typecheck
- git diff --check

PR #2233, opened 2026-05-11
fix(api): redact plugin parser error details

justinasrolando-cpu:
Replace raw request bodies, full schema parse results, and raw `version_build` values in plugin parser validation errors.

## Changes

- Empty body error: replace `{ body }` with `{ has_fields: false }`
- `missing_device_id`: replace `{ body }` with `{ has_device_id: false, has_app_id }`
- `missing_app_id`: replace `{ body }` with `{ has_device_id, has_app_id: false }`
- `semver_error`: strip raw `version_build` from error message template and data; emit `{ version_build_length }` instead
- Schema parse failure: replace `{ parseResult }` with safe issues list (`{ code, path }` only) + `issue_count`

Closes #2171

### DIFF
--- a/tests/plugin-parser-redaction.unit.test.ts
+++ b/tests/plugin-parser-redaction.unit.test.ts
@@ -1,0 +1,115 @@
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('../supabase/functions/_backend/utils/hono.ts', () => ({
+  simpleError: (code: string, message: string, data?: Record<string, unknown>) => {
+    const err: any = new Error(message)
+    err.code = code
+    err.data = data ?? null
+    return err
+  },
+}))
+
+vi.mock('../supabase/functions/_backend/utils/ark_validation.ts', async () => {
+  const actual = await import('../supabase/functions/_backend/utils/ark_validation.ts')
+  return actual
+})
+
+const { parsePluginBody } = await import('../supabase/functions/_backend/utils/plugin_parser.ts')
+const { type } = await import('arktype')
+
+const testSchema = type({
+  device_id: 'string',
+  app_id: 'string',
+  plugin_version: 'string',
+  version_build: 'string',
+  platform: 'string',
+})
+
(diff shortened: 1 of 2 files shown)


tests/plugin-parser-redaction.unit.test.ts
```diff
@@ -0,0 +1,115 @@
...
+vi.mock('../supabase/functions/_backend/utils/hono.ts', () => ({
+  simpleError: (code: string, message: string, data?: Record<string, unknown>) => {
+    const err: any = new Error(message)
+    err.code = code
+    err.data = data ?? null
+    return err
+  },
+}))
+
+vi.mock('../supabase/functions/_backend/utils/ark_validation.ts', async () => {
```
darshan-Jahagirdar, line 12: This mock factory imports the same rk_validation.ts module that it is mocking. In Vitest, wait import(...) here resolves through the mock system again, which can recurse or fail during module setup. If the goal is to pass the real module through unchanged, use wait vi.importActual<typeof import('../supabase/functions/_backend/utils/ark_validation.ts')>('../supabase/functions/_backend/utils/ark_validation.ts') instead, or remove this mock entirely.

subhajitlucky:
One regression gap in the new schema-parse redaction test: the last case passes even if `parsePluginBody()` does not throw.

`schema parse failure does not include raw parseResult with embedded values` wraps the call, stores `thrown`, and then guards all assertions with `if (thrown) { ... }`. If a future ArkType/schema change or fixture mistake makes that invalid payload parse successfully, this test still passes and the `parseResult` redaction path is no longer covered.

I would make the throw mandatory before the redaction assertions, for example:

```ts
expect(thrown).toBeDefined()
expect(thrown?.data).not.toHaveProperty("parseResult")
```

or use `expect(() => parsePluginBody(...)).toThrow()` and then inspect the captured error. That keeps the test aligned with the PR goal: schema parse failures should happen and should not carry raw parse details.

WcaleNieWolny:

PR #2170, opened 2026-05-11
fix(api): redact download link request logs

nickita-khylkouski:
## Summary
- replace raw private download_link request body logging with metadata-only booleans
- avoid returning/logging raw app ids on access-denied errors
- avoid returning/logging the raw bundle row when owner_org is missing
- add focused unit coverage for request and bundle metadata redaction

/claim #1667

## Tests
- bunx vitest run tests/download-link-redaction.unit.test.ts
- bunx eslint supabase/functions/_backend/private/download_link.ts tests/download-link-redaction.unit.test.ts
- git diff --check

PR #2234, opened 2026-05-11
fix(api): redact download link request logs

justinasrolando-cpu:
Stop logging raw request body and stop returning raw `app_id` / bundle row in download_link error responses.

## Changes

- **cloudlog**: replace `body` with metadata booleans (`has_app_id`, `has_id`, `has_manifest`) so no caller-supplied values reach the log stream
- **`app_access_denied`**: remove `{ app_id: body.app_id }` from error data
- **`cannot_get_owner_org`**: replace `{ bundle }` (full DB row) with `{ has_bundle, has_owner_org }` presence flags

Closes #2170

### DIFF
--- a/supabase/functions/_backend/private/download_link.ts
+++ b/supabase/functions/_backend/private/download_link.ts
@@ -20,7 +20,7 @@ app.use('/', useCors)
 
 app.post('/', middlewareAuth, async (c) => {
   const body = await parseBody<DataDownload>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'post download link body', body })
+  cloudlog({ requestId: c.get('requestId'), message: 'post download link body', has_app_id: !!body.app_id, has_id: !!body.id, has_manifest: !!body.isManifest })
   const authorization = c.get('authorization')
   if (!authorization)
     throw simpleError('cannot_find_authorization', 'Cannot find authorization')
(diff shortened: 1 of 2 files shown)


supabase/functions/_backend/private/download_link.ts
```diff
@@ -53,7 +53,7 @@ app.post('/', middlewareAuth, async (c) => {
   }
 
   if (!ownerOrg) {
-    throw simpleError('cannot_get_owner_org', 'Cannot get owner org', { bundle })
+    throw simpleError('cannot_get_owner_org', 'Cannot get owner org', { has_bundle: !!bundle, has_owner_org: !!bundle?.owner_org })
```
darshan-Jahagirdar, line 56: This sanitized error still may not run when the join is missing. const ownerOrg = bundle?.owner_org.created_by only guards undle; if undle exists but owner_org is null, it throws a TypeError before reaching this branch, which then bypasses the intended sanitized cannot_get_owner_org response. Could this be undle?.owner_org?.created_by so the new redacted error path handles the missing relation?

subhajitlucky:
Review note: the new redaction tests do not exercise the actual `download_link` code paths that changed.

`tests/download-link-redaction.unit.test.ts` builds local `logEntry`, `err`, and `errData` objects that mirror the intended output, but it never imports `supabase/functions/_backend/private/download_link.ts`, mocks `cloudlog` / `simpleError`, or calls the handler. That means the tests can stay green even if a future edit reintroduces `cloudlog({ body })`, `{ app_id: body.app_id }`, or `{ bundle }` in the real endpoint.

For this kind of security regression, I would add at least one handler-level or module-level test that drives the actual missing-permission and missing-owner-org branches with mocks and asserts the captured `cloudlog` / thrown error data does not contain the raw `app_id`, storage provider, or bundle row fields.

WcaleNieWolny:

PR #2235, opened 2026-05-11
fix(api): redact failed version delete logs

justinasrolando-cpu:
Stop logging raw request body and stop returning raw `app_id` / body in `delete_failed_version` error responses.

## Changes

- **cloudlog**: replace `body` with `{ has_app_id, has_name }` booleans — no caller-supplied values in log stream
- **`not_authorized`**: remove `{ app_id: body.app_id }` from error data
- **`error_app_id_missing` / `error_bundle_name_missing`**: replace `{ body }` with `{ has_app_id, has_name }` presence flags

Closes #2169

### DIFF
--- a/supabase/functions/_backend/private/delete_failed_version.ts
+++ b/supabase/functions/_backend/private/delete_failed_version.ts
@@ -17,7 +17,7 @@ export const app = new Hono<MiddlewareKeyVariables>()
 
 app.delete('/', middlewareKey(['all', 'write', 'upload']), async (c) => {
   const body = await parseBody<DataUpload>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'delete failed version body', body })
+  cloudlog({ requestId: c.get('requestId'), message: 'delete failed version body', has_app_id: !!body.app_id, has_name: !!body.name })
   const apikey = c.get('apikey')
   const capgkey = c.get('capgkey') as string
   if (apikey && typeof apikey === 'object') {
@@ -31,7 +31,7 @@ app.delete('/', middlewareKey(['all', 'write', 'upload']), async (c) => {
   }
   // Auth context is already set by middlewareKey
   if (!(await checkPermission(c, 'bundle.delete', { appId: body.app_id }))) {
-    return quickError(401, 'not_authorized', 'You can\'t access this app', { app_id: body.app_id })
+    return quickError(401, 'not_authorized', 'You can\'t access this app')
   }
 
   const { error: errorApp } = await supabaseApikey(c, capgkey)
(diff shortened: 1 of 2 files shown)


supabase/functions/_backend/private/delete_failed_version.ts
```diff
@@ -31,7 +31,7 @@ app.delete('/', middlewareKey(['all', 'write', 'upload']), async (c) => {
   }
   // Auth context is already set by middlewareKey
   if (!(await checkPermission(c, 'bundle.delete', { appId: body.app_id }))) {
```
darshan-Jahagirdar, line 33: This still leaks the raw app id through the permission check logs. `checkPermission` logs the full `scope` object in `utils/rbac.ts`, so a request with `body.app_id = "com.secret.app"` will still emit that value before this endpoint returns the now-redacted `not_authorized` error. That also means the new unit tests can pass while the runtime path still logs the sensitive identifier.

Can you validate `body.app_id`/`body.name` before calling `checkPermission`, or update the permission logging path to log presence/hashed scope fields instead of the raw `appId`?

justinasrolando-cpu:
The `Run Cloudflare Workers backend tests` failure is unrelated to this PR.

The failing test (`tests/app-error-cases.test.ts`) gets a plain-text `"Your worke..."` response from the Cloudflare Worker instead of JSON — this is a transient CI infrastructure issue (worker quota exceeded in the test runner environment). The test does not touch any of the files changed in this PR (`private/delete_failed_version.ts`, `tests/delete-failed-version-redaction.unit.test.ts`).

Please re-run the failing check when convenient.

justinasrolando-cpu:

```
SyntaxError: Unexpected token 'Y', "Your worke"... is not valid JSON
 FAIL tests/app-error-cases.test.ts
```

This is the known Cloudflare Workers quota issue. 1719/1720 tests passed. Could you please re-run the failing check?

subhajitlucky:
One real handler path still looks uncovered here: the new `error_app_id_missing` response is after `checkPermission(c, 'bundle.delete', { appId: body.app_id })`.

For a request with no `app_id`, the handler calls `checkPermission` with `appId: undefined` before it reaches:

```ts
if (!body.app_id) {
  return quickError(400, 'error_app_id_missing', ...)
}
```

If the permission helper rejects that undefined scope, the endpoint returns `not_authorized` instead of the intended 400 missing-app-id error, and this PR’s new redacted missing-field response is effectively unreachable. I would move the `!body.app_id` / `!body.name` validation before the permission check, then add an endpoint-level regression. The current tests recreate the expected log/error objects locally, so they would still pass even if the production handler never reaches this branch.

WcaleNieWolny:

PR #2163, opened 2026-05-11
fix(api): redact stripe billing logs

HappyReni:
/claim #1667

## Summary
- replace raw stripe portal and checkout request-body logging with metadata-only logging
- avoid logging raw auth user IDs and Stripe customer IDs during portal and checkout creation
- add unit coverage that callback URLs, checkout redirect URLs, org IDs, user IDs, customer IDs, price IDs, client references, and attribution IDs are not retained in logs

## Test plan
- ./node_modules/.bin/vitest run tests/stripe-portal-log-redaction.unit.test.ts
- ./node_modules/.bin/eslint supabase/functions/_backend/private/stripe_portal.ts supabase/functions/_backend/private/stripe_checkout.ts tests/stripe-portal-log-redaction.unit.test.ts
- git diff --check

PR #2236, opened 2026-05-11
fix(api): redact stripe billing logs

justinasrolando-cpu:
Replace raw Stripe portal/checkout request body logging with metadata-only booleans.

Closes #2163

### DIFF
--- a/supabase/functions/_backend/private/stripe_checkout.ts
+++ b/supabase/functions/_backend/private/stripe_checkout.ts
@@ -23,7 +23,7 @@ app.use('/', useCors)
 
 app.post('/', middlewareAuth, async (c) => {
   const body = await parseBody<CheckoutData>(c)
-  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout body', body })
+  cloudlog({ requestId: c.get('requestId'), message: 'post stripe checkout body', has_org_id: !!body.orgId, has_price_id: !!body.priceId, has_success_url: !!body.successUrl, has_cancel_url: !!body.cancelUrl })
 
   if (!body.orgId)
     throw simpleError('no_org_id_provided', 'No org_id provided')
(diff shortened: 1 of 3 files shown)


lyndon050516 (commented):
One remaining logging path looks like it defeats part of this redaction for both billing endpoints. The endpoint-specific `cloudlog` calls no longer include `body.orgId` or `authContext.userId`, but both handlers still call `checkPermission(c, 'org.update_billing', { orgId: body.orgId })` immediately afterward. `checkPermission` logs `scope` and `userId` in its normal `checkPermission: checking` and `checkPermission: result` entries, so a Stripe checkout/portal request can still put the raw org id and user id into the log stream via the shared RBAC helper.

That means the PR removes the local `post stripe ... body` / `auth` leaks, but it does not actually keep org/user identifiers out of operational logs for this billing flow. If the intent is to stop retaining those identifiers here, the permission helper needs a metadata-only logging mode, or these endpoints need to avoid passing raw request IDs into a helper that logs the full scope. A focused regression would need to exercise the production handler through the permission check path; the current tests only recreate the sanitized local log objects, so they would not catch the remaining `checkPermission` logs.

supabase/functions/_backend/private/stripe_checkout.ts
```diff
@@ -54,7 +54,7 @@ app.post('/', middlewareAuth, async (c) => {
   if (!await checkPermission(c, 'org.update_billing', { orgId: body.orgId }))
     throw simpleError('not_authorize', 'Not authorize')
 
-  cloudlog({ requestId: c.get('requestId'), message: 'user', org })
+  cloudlog({ requestId: c.get('requestId'), message: 'user', has_customer_id: !!org.customer_id })
   const checkout = await createCheckout(c, org.customer_id, body.recurrence ?? 'month', body.priceId ?? 'price_1KkINoGH46eYKnWwwEi97h1B', body.successUrl ?? `${getEnv(c, 'WEBAPP_URL')}/app/usage`, body.cancelUrl ?? `${getEnv(c, 'WEBAPP_URL')}/app/usage`, body.clientReferenceId, body.attributionId)
```
darshan-Jahagirdar, line 58: This still leaves a raw URL leak on the invalid redirect path. `createCheckout` calls `getAllowedRedirectUrl`, and that helper currently throws `simpleError('invalid_redirect_url', ..., { field, value, expectedOrigin })` when parsing fails or the origin does not match. Since `value` is the user-provided `successUrl`/`cancelUrl`, a rejected URL with tokens in the query string can still end up in the error metadata even though the initial body log is now redacted.

Can you also sanitize `getAllowedRedirectUrl` so it logs presence/origin validity instead of the raw `value`? The same applies to `stripe_portal` through `callbackUrl`.

WcaleNieWolny:

PR #2237, opened 2026-05-12
fix(api): cap PostHog error response bodies

tngus6007:
/claim #1667

## Summary
- Add bounded reads for PostHog error responses before logging them.
- Apply the cap to capture, exception, and group-identify failure paths.
- Replace oversized PostHog error bodies with a stable placeholder instead of retaining the full upstream response.
- Add regression coverage for capture/exception failures and `Content-Length` fast-fail handling.

## Security note
This is public-safe DoS/log hardening for backend telemetry helpers. Normal small PostHog error messages are preserved, while oversized upstream error bodies no longer force unbounded buffering or log retention.

## Test plan
- `./node_modules/.bin/vitest.exe run tests/posthog.unit.test.ts --reporter=verbose`
- `git diff --check`

## Screenshots
Not applicable; backend helper hardening only.

## Checklist

## Release Notes

* **Bug Fixes**
  * Improved error handling to prevent oversized error responses from being excessively logged, ensuring better system performance and preventing potential data capture issues.

* **Tests**
  * Added comprehensive test coverage for error body size limits to verify proper handling of large error responses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2237)

### DIFF
--- a/supabase/functions/_backend/utils/posthog.ts
+++ b/supabase/functions/_backend/utils/posthog.ts
@@ -5,6 +5,7 @@ import { existInEnv, getEnv } from './utils.ts'
 
 const POSTHOG_CAPTURE_URL = 'https://eu.i.posthog.com/capture/'
 const POSTHOG_EXCEPTION_URL = 'https://eu.i.posthog.com/i/v0/e/'
+const MAX_POSTHOG_ERROR_BODY_BYTES = 4 * 1024
 
 export type PostHogGroups = Record<string, string>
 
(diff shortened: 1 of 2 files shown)


tngus6007:
Addressed the CodeRabbit stream-cancel nit in the latest commit.

The `Content-Length` fast-fail path now cancels the response body before returning the oversized marker path, so oversized PostHog error responses do not leave an unread stream around.

Validation:
- `vitest run tests/posthog.unit.test.ts --reporter=verbose`
- `bun typecheck`
- `git diff --check`

tngus6007:
Pushed a follow-up aimed at the Sonar duplication signal while preserving the PostHog error-body behavior.

Changes:
- rewrote the bounded PostHog response reader to avoid mirroring the repeated response-reader structure
- collapsed repeated oversized-error-body test assertions into a shared helper

Validation:
- `vitest run tests/posthog.unit.test.ts --reporter=verbose`
- `bun x eslint supabase/functions/_backend/utils/posthog.ts tests/posthog.unit.test.ts`
- `bun typecheck`
- `git diff --check`

subhajitlucky:
Review note: the bounded PostHog error-body reader is now used in three call sites, but the new endpoint-level tests only cover `trackPosthogEvent()` and `capturePosthogException()`.

`groupIdentifyPosthog()` also switched from `await res.text()` to `formatPostHogErrorBody(await readPostHogErrorBody(res))`, but there is no regression test that exercises an oversized `$groupidentify` error response through that public helper. A future change could accidentally reintroduce an unbounded read or raw log body there while the current tests still pass.

I would add one test mirroring the existing oversized helper for `groupIdentifyPosthog()`, asserting it returns `false`, logs `PostHog $groupidentify error`, and does not include the oversized response body in `cloudlogErr`.

WcaleNieWolny:

PR #2238, opened 2026-05-12
fix(api): cap LogSnag error response bodies

tngus6007:
/claim #1667

## Summary
- bound LogSnag insight error response reads to 4 KiB before logging
- avoid logging oversized external response bodies by replacing them with a fixed marker
- add unit coverage for small, content-length oversized, and chunked oversized LogSnag error responses

## Tests
- `vitest run tests/logsnag.unit.test.ts --reporter=verbose`
- `bun typecheck`

## Release Notes

* **New Features**
  * Added size-limiting support for error body handling with automatic overflow detection and safe stream processing.

* **Tests**
  * Added comprehensive test suite for error logging with coverage for size limits, overflow detection, and response stream handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2238)

### DIFF
--- a/supabase/functions/_backend/utils/logsnag.ts
+++ b/supabase/functions/_backend/utils/logsnag.ts
@@ -4,6 +4,9 @@ import { LogSnag } from '@logsnag/node'
 import { cloudlog, cloudlogErr, serializeError } from './logging.ts'
 import { getEnv } from './utils.ts'
 
+const MAX_LOGSNAG_ERROR_BODY_BYTES = 4 * 1024
+const LOGSNAG_ERROR_BODY_TOO_LARGE = '[logsnag_error_body_too_large]'
+
 function logsnag(c: Context) {
   const ls = getEnv(c, 'LOGSNAG_TOKEN')
     ? new LogSnag({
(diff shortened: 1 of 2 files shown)


Fleehopper (approved):
LGTM from a targeted review of the LogSnag error-body cap.

The important failure modes look covered here: a declared oversized `content-length` short-circuits before reading the body, chunked/unknown-length responses cancel the stream once the byte cap is crossed, small upstream errors remain available for diagnostics, and the log assertion checks the oversized payload is not present in the serialized log call. I also checked the call site: replacing the old `response.text()` read with `readLimitedResponseText()` keeps the existing `false` return behavior for failed LogSnag calls.

One minor follow-up, not blocking this PR: the same bounded-read helper pattern now exists in a few integrations, so after this batch lands it may be worth centralizing the helper to keep future caps consistent. I could not run the Bun/Vitest command locally because this Windows environment does not have `bun` on PATH, so this approval is based on manual diff review and the GitHub checks.

supabase/functions/_backend/utils/logsnag.ts
```diff
@@ -21,6 +24,51 @@ function logsnag(c: Context) {
...
+  const contentLength = response.headers.get('content-length')
+  if (!contentLength)
+    return false
+
+  const parsed = Number.parseInt(contentLength, 10)
+  return Number.isFinite(parsed) && parsed > maxBytes
+}
+
+async function readLimitedResponseText(response: Response, maxBytes: number) {
+  if (isOversizedContentLength(response, maxBytes))
```
darshan-Jahagirdar: Same cleanup gap as in the other capped-read helpers: when content-length is already over the limit, this returns the marker without cancelling 
esponse.body. That avoids buffering the oversized error, but the stream/connection can remain open until runtime cleanup. Since the streaming oversize path cancels the reader below, the content-length fast path should also wait response.body?.cancel().catch(() => undefined) before returning.

tngus6007:
Fixed in the latest pushed commit (`4e0233cc1`). The content-length fast path now cancels the response body before returning the capped LogSnag error marker, and I added a unit test covering that cancellation path.

Validation run locally:
- `npx --yes bun@latest x vitest run tests/logsnag.unit.test.ts --reporter=verbose`
- `npx --yes bun@latest x eslint supabase/functions/_backend/utils/logsnag.ts tests/logsnag.unit.test.ts`
- `git diff --check`

tngus6007:
Addressed the latest CodeRabbit test-style comment in the latest pushed commit (`05a3731f0`). The LogSnag unit tests now use `it.concurrent()` while keeping the test bodies unchanged.

Validation run locally:
- `npx --yes bun@latest x vitest run tests/logsnag.unit.test.ts --reporter=verbose`
- `npx --yes bun@latest x eslint tests/logsnag.unit.test.ts`
- `git diff --check`

nagiexplorer88:
One test reliability issue with the latest `it.concurrent()` change: these tests still share the same hoisted `fetchMock`, `cloudlogErrMock`, and global `fetch` stub across all cases.

Because each concurrent test mutates `fetchMock.mockResolvedValue(...)` and then awaits `logsnagInsights()`, another test can overwrite the shared mock before the first one performs its fetch/assertions. The same applies to assertions over `cloudlogErrMock.mock.calls`, which can now include calls from sibling tests.

I would either keep this suite serial, or make each case use isolated mocks/modules so the concurrent tests cannot observe each other's fetch responses or log calls. Otherwise the suite can become order-dependent or flaky while still testing security-sensitive redaction behavior.

WcaleNieWolny:

PR #2240, opened 2026-05-12
fix(api): cap Bento error response bodies

tngus6007:
/claim #1667

## Summary
- bound Bento API error response reads before including them in thrown errors/logs
- replace oversized Bento error bodies with a fixed marker
- add unit coverage for small, chunked oversized, and content-length oversized Bento error responses

## Tests
- `vitest run tests/bento.unit.test.ts --reporter=verbose`
- `bun typecheck`

### DIFF
--- a/supabase/functions/_backend/utils/bento.ts
+++ b/supabase/functions/_backend/utils/bento.ts
@@ -2,6 +2,9 @@ import type { Context } from 'hono'
 import { cloudlog, cloudlogErr, serializeError } from './logging.ts'
 import { getEnv } from './utils.ts'
 
+const MAX_BENTO_ERROR_BODY_BYTES = 4 * 1024
+const BENTO_ERROR_BODY_TOO_LARGE = '[bento_error_body_too_large]'
+
 export function isBentoConfigured(c: Context) {
   const publishableKey = (getEnv(c, 'BENTO_PUBLISHABLE_KEY') || '').trim()
   const secretKey = (getEnv(c, 'BENTO_SECRET_KEY') || '').trim()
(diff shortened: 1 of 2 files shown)


Serve63 (approved):
LGTM. I did a targeted review of the bounded Bento error-body path: oversized content-length short-circuits before reading, streamed bodies cancel once they exceed the byte cap, small error bodies still flow through for useful diagnostics, and oversized bodies are replaced with the fixed marker before logging. The unit coverage checks the small-body case, chunked oversized body case, and content-length oversized case, and the GitHub checks are green. I couldn't run the Bun test command locally in this environment, so this approval is based on manual diff review plus the passing CI.

supabase/functions/_backend/utils/bento.ts
```diff
@@ -37,6 +40,51 @@ function getBentoHeaders(c: Context) {
...
+  const contentLength = response.headers.get('content-length')
+  if (!contentLength)
+    return false
+
+  const parsed = Number.parseInt(contentLength, 10)
+  return Number.isFinite(parsed) && parsed > maxBytes
+}
+
+async function readLimitedBentoErrorBody(response: Response, maxBytes: number) {
+  if (isOversizedContentLength(response, maxBytes))
```
darshan-Jahagirdar: On the content-length fast path we return the marker without cancelling the response body. That avoids reading the oversized payload, but it can leave the underlying stream/connection open until the runtime cleans it up. Since this is the resource-cap path, it should cancel 
esponse.body before returning, matching the streaming oversize branch below.

tngus6007:
Fixed in the latest pushed commit (`6cc9a638b`). The content-length fast path now cancels the response body before returning the capped Bento error marker, and I added a unit test covering that cancellation path.

Validation run locally:
- `npx --yes bun@latest x vitest run tests/bento.unit.test.ts --reporter=verbose`
- `npx --yes bun@latest x eslint supabase/functions/_backend/utils/bento.ts tests/bento.unit.test.ts`
- `git diff --check`

Serve63:
Rechecked 6cc9a638b against the follow-up concern: the content-length fast path now cancels response.body before returning the capped Bento error marker, and the added unit test asserts that cancellation path. That addresses the resource-cleanup gap while preserving the no-read behavior for oversized responses, so my approval still stands.

WcaleNieWolny:

PR #2241, opened 2026-05-12
fix: redact sensitive Discord alert fields

darshan-Jahagirdar:
## Summary
- move Discord alert request-body/header sanitization into a focused helper
- recursively sanitize JSON request bodies so password-like fields are removed and token-like fields are partially redacted
- cover access/refresh tokens, session keys, password confirmation fields, and sensitive query parameters with unit tests

## Validation
- `npx --yes tsx -` smoke test for JSON body, query-string body, and header sanitization
- `git diff --check`

Could not run the repo Vitest command locally because this checkout has no `node_modules` and `bunx` is not installed; `npx vitest` cannot load the project config without the repo's local Vite dependency.

/claim #1667

## Summary by CodeRabbit

* **Bug Fixes**
  * Discord error alerts now automatically redact sensitive data (passwords, tokens, API keys) before posting to prevent credential exposure.

* **Tests**
  * Added comprehensive test coverage for Discord alert data sanitization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2241)

### DIFF
--- a/supabase/functions/_backend/utils/discord.ts
+++ b/supabase/functions/_backend/utils/discord.ts
@@ -46,7 +46,7 @@ export function sendDiscordAlert500(c: Context, functionName: string, body: stri
   const userAgent = c.req.header('user-agent') ?? 'unknown'
   const ip = c.req.header('cf-connecting-ip') ?? c.req.header('x-forwarded-for') ?? 'unknown'
   const method = c.req.method
-  const url = c.req.url
+  const url = sanitizeSensitiveFromString(c.req.url)
   const rawHeaders = Object.fromEntries((c.req.raw.headers as any).entries())
   const headers = sanitizeSensitiveHeaders(rawHeaders)
   const errorMessage = sanitizeSensitiveFromString(e?.message ?? 'Unknown error')
--- a/supabase/functions/_backend/utils/discord.ts
+++ b/supabase/functions/_backend/utils/discord.ts
@@ -2,66 +2,9 @@ import type {
   RESTPostAPIWebhookWithTokenJSONBody,
 } from 'discord-api-types/v10'
 import type { Context } from 'hono'
+import { sanitizeSensitiveFromString, sanitizeSensitiveHeaders } from './discord_sanitization.ts'
 import { cloudlog, cloudlogErr } from './logging.ts'
 import { getEnv } from './utils.ts'
-
-// Fields that should be completely removed from logs (never logged)
-const REMOVED_FIELDS = ['password']
-// Fields that should show first 4 and last 4 characters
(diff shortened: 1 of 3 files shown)


lyndon050516 (commented):
One failure mode to cover before merge: `sanitizeSensitiveFromString` can now throw while sanitizing non-JSON text if a sensitive query parameter has malformed percent encoding.

The replacement callback does `decodeURIComponent(value)` unguarded. A payload such as `url=https://example.test/cb?access_token=%E0%A4%A` will raise `URIError: URI malformed`, which means a best-effort Discord alert sanitizer can fail the alert path instead of returning a redacted string.

I would wrap the decode in a small helper that falls back to the raw value when decoding fails, then still redact that raw value. A focused regression around malformed encoded `access_token` / `refresh_token` values would make sure the sanitizer remains non-throwing for arbitrary error text.

lyndon050516:
This is still present on the current head (`fe49e39`): the non-JSON query-param replacement callback still does `decodeURIComponent(value)` unguarded.

```ts
return `${separator}${key}=${encodeURIComponent(partialRedact(decodeURIComponent(value)))}`
```

So `sanitizeSensitiveFromString("?access_token=%E0%A4%A")` can still throw `URIError` before returning a sanitized string. The new tests cover normal encoded/query values, but not malformed percent-encoding. A tiny helper such as `safeDecodeURIComponent(value)` with a raw-value fallback would keep the alert sanitizer best-effort for arbitrary error text.

darshan-Jahagirdar:
Follow-up pushed in `98eb71d` for the malformed percent-encoding case: `decodeComponent` catches invalid encodings, and `tests/discord-alert-sanitization.unit.test.ts` now has an explicit non-throwing malformed `%` sequence test.

Current checks are green, including `Run tests`, Playwright, SonarCloud, dead-code, and benchmarks.

darshan-Jahagirdar:
Pushed one more hardening pass in `017a868`: the same sanitizer is now applied to error messages, stacks, and serialized error objects, and JSON string values containing token-like query parameters are covered by tests.

Current checks are green again, including `Run tests`, Playwright, SonarCloud, dead-code, and benchmarks.

Zsh123456-BOOP:
Review note: `sendDiscordAlert500()` still includes the raw request URL in the Discord alert.

The PR now sanitizes the body, headers, error message, stack, and serialized error object, but `c.req.url` is still embedded directly in the "Request Details" field. Callback/error URLs commonly carry sensitive query parameters such as `access_token`, `refresh_token`, `code`, `api_key`, or password-reset tokens, so this can still leak secrets into Discord logs.

I would pass the URL through the same sanitizer before rendering it, e.g. `const safeUrl = sanitizeSensitiveFromString(c.req.url)`, and add a regression test with `/callback?access_token=secret&next=/dashboard` asserting the raw token is absent from the alert payload.

WcaleNieWolny:

PR #2242, opened 2026-05-12
fix(webhooks): route dashboard writes through API

BrianChung95:
## Summary

- route dashboard webhook create/update/delete actions through the existing `webhooks` edge-function API
- allow webhook write endpoints to use the same unified JWT/API-key auth path already used by test/retry routes
- add regression coverage for JWT create/update/delete webhook API access

## Why

The dashboard was still writing webhook rows directly through the browser-side Supabase client while the backend already has centralized webhook permission checks for org admin access, API-key scope policy, and URL/event validation. This keeps the UI behavior the same while putting write operations through the backend authorization path.

This complements the existing webhook API-key/RLS hardening work by removing the remaining direct browser-side write path for webhook management.

Related: #1667
/claim #1667

## Testing

- `git diff --check`
- Not run locally: this checkout does not have `bun`, `vitest`, or installed `node_modules`; CI should run `tests/webhooks.test.ts`.

## Security note

Public-safe hardening only. No production systems, credentials, or private data were accessed.

## Summary by CodeRabbit

* **New Features**
  * Webhook operations (create, update, delete) now support JWT-based authentication.

* **Tests**
  * Expanded webhook test suite with JWT authentication workflows and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2242)

### DIFF
--- a/src/stores/webhooks.ts
+++ b/src/stores/webhooks.ts
@@ -34,8 +34,71 @@ const DELIVERIES_PER_PAGE = 50
 
 const supabase = useSupabase()
 
+type WebhookRow = Database['public']['Tables']['webhooks']['Row']
+type WebhookWriteData = Partial<{
+  name: string
+  url: string
+  events: string[]
+  enabled: boolean
+}>
+
+function getCurrentOrgId(): string | undefined {
+  return useOrganizationStore().currentOrganization?.gid
+}
+
+function validateWebhookUrl(url: string): string | undefined {
+  try {
+    const parsedUrl = new URL(url)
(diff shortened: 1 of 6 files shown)


riderx:
Write from Supabase is supported and expected.whats the point of doing this if the webhooks rls are not changed ?

BrianChung95:
Thanks for clarifying. I misframed this: direct Supabase writes are supported here, and without a corresponding webhook RLS policy change this PR does not address the right layer.

I’m closing it to avoid adding churn. I’ll keep any further security/audit work scoped to the advisory/rules process rather than discussing vulnerability framing publicly.

PR #2246, opened 2026-05-12
fix(api): stop 429 handler from hiding the real error cause

WcaleNieWolny:
## Summary

The 429 branch in `onError` ([supabase/functions/_backend/utils/on_error.ts](supabase/functions/_backend/utils/on_error.ts)) unconditionally rewrote every response body to:

```json
{ "error": "too_many_requests", "message": "You are being rate limited" }
```

This discarded the actual `error` code, `message`, and `moreInfo` from the thrower. Several genuinely different 429-producing conditions all collapsed into the same opaque message:

| Thrower | Real `error` code | Real `moreInfo` |
|---|---|---|
| `simpleRateLimit` (IP failed-auth lockout) | `too_many_requests` | `reason: too_many_failed_auth_attempts`, `rateLimitResetAt`, `retryAfterSeconds` |
| `simpleRateLimit` (API-key flood) | `too_many_requests` | `reason: api_key_rate_limit_exceeded`, `apikey_id`, `rateLimitResetAt`, `retryAfterSeconds` |
| `reserveNativeBuildSlot` | `native_build_concurrency_limit_exceeded` | `activeBuilds`, `limit`, `planName` |
| `cron_clear_versions` guards | `rate_limit_exceeded` | — |

The real-world impact: an org that exhausted its plan's native build concurrency saw `"You are being rate limited"` from `/build/start`, with no signal that the resolution was to free a build slot or upgrade the plan, not to wait for a rate-limit window to roll over. Tracking that down required reading the source. (Discovered while debugging a `Failed to start build: 429 …` report — the rate-limit error reported by the CLI was a `native_build_concurrency_limit_exceeded` in disguise, compounded by orphaned `build_requests` rows that `cron_reconcile_build_status` had not reaped.)

## Change

Keep the header derivation (`Retry-After`, `X-RateLimit-Reset`) — that path is still useful whenever `moreInfo` carries `rateLimitResetAt` / `retryAfterSeconds` — and drop the body override so control falls through to the existing \`return c.json(res, e.status)\` below.

\`\`\`diff
       if (e.status === 429) {
+        // Set rate-limit headers from moreInfo when available, but DO NOT
+        // overwrite the response body. […] Fall through to
+        // \`return c.json(res, e.status)\` below so the thrower's real error
+        // payload is preserved.
         const rateLimitResetAt = …
         …
-        return c.json({ error: 'too_many_requests', message: 'You are being rate limited' }, e.status)
       }
\`\`\`

## Backward-compat / test plan

- [x] `simpleRateLimit` callers still emit \`error: "too_many_requests"\` (only the message changes from \"You are being rate limited\" to \"Too many requests\", which is what the thrower originally set).
- [x] The only existing test that asserts on the body, [tests/account-rate-limit.unit.test.ts:145](tests/account-rate-limit.unit.test.ts), stubs its own \`app.onError\` and asserts on \`error\` and \`reason\` from the thrown cause — unchanged by this diff.
- [x] CLI [error-recovery hint matcher](https://github.com/Cap-go/CLI/blob/main/src/build/onboarding/recovery.ts) checks the lowercased message for \`'429'\` (still present in CLI-side wrapping like \`Failed to start build: 429 - …\`) or \`'rate limit'\`. The substring \`'rate limit'\` no longer appears in `simpleRateLimit` bodies — but that matcher is on the Apple onboarding error path, not the Capgo backend response body, so this is not a regression.

## Follow-ups (not in this PR)

1. Add a regression unit test that throws a 429 with a non-\`too_many_requests\` errorCode (e.g. \`native_build_concurrency_limit_exceeded\`) and asserts the response body still carries that code + moreInfo through \`onError\`.
2. Investigate why \`cron_reconcile_build_status\` is not reaping orphaned native build rows that never reach a terminal status — separate bug surfaced by the same incident.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rate-limit (429) responses now preserve structured error details and set appropriate rate-limit headers for better client handling.
  * Rate-limit metadata is curated to avoid echoing full request payloads, ensuring only intended fields are returned in the error `moreInfo`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2246)

### DIFF
--- a/supabase/functions/_backend/plugins/channel_self.ts
+++ b/supabase/functions/_backend/plugins/channel_self.ts
@@ -505,7 +505,10 @@ async function parseChannelSelfPluginRequest(
   cloudlog({ requestId: c.get('requestId'), message: logMessage, body })
 
   if (isLimited(c, body.app_id)) {
-    return { response: simpleRateLimit(body) }
+    // Pass curated metadata only — see simpleRateLimit contract in hono.ts.
+    // Reflecting the raw `body` would echo the client's full DeviceLink
+    // payload back inside the 429 response's `moreInfo`.
+    return { response: simpleRateLimit({ app_id: body.app_id, device_id: body.device_id }) }
   }
 
   const bodyParsed = parsePluginBody<DeviceLink>(c, body, schema, requireDevice)
--- a/supabase/functions/_backend/plugins/updates.ts
+++ b/supabase/functions/_backend/plugins/updates.ts
@@ -21,7 +21,10 @@ app.post('/', async (c) => {
   const body = await parseBody<AppInfos>(c)
   cloudlog({ requestId: c.get('requestId'), message: 'post updates body', body })
   if (isLimited(c, body.app_id)) {
-    return simpleRateLimit(body)
+    // Pass curated metadata only — see simpleRateLimit contract in hono.ts.
+    // Reflecting the raw `body` would echo the client's full update payload
+    // back inside the 429 response's `moreInfo`.
+    return simpleRateLimit({ app_id: body.app_id })
   }
   return update(c, parsePluginBody<AppInfos>(c, body, updateRequestSchema))
 })
--- a/supabase/functions/_backend/utils/hono.ts
+++ b/supabase/functions/_backend/utils/hono.ts
@@ -314,6 +314,17 @@ export function quickError(status: number, errorCode: string, message: string, m
   })
 }
 
+/**
+ * Throw a 429 "too_many_requests" HTTPException.
+ *
+ * IMPORTANT: `moreInfo` is reflected to the client as the `moreInfo` field of
+ * the 429 response body (see `onError` in `on_error.ts`). Pass curated
+ * diagnostic metadata only — fields like `app_id`, `device_id`, `reason`,
+ * `apikey_id`, `rateLimitResetAt`, `retryAfterSeconds`. Do NOT pass the raw
+ * parsed request body: that turns the rate-limit response into a reflective
+ * echo of whatever the client submitted and means any future field added to
+ * the request schema (sensitive or not) silently lands in the error payload.
+ */
 export function simpleRateLimit(moreInfo: any = {}, cause?: any): never {
   const status = 429
   const message = 'Too many requests'
--- a/supabase/functions/_backend/utils/on_error.ts
+++ b/supabase/functions/_backend/utils/on_error.ts
@@ -119,6 +119,16 @@ export function onError(functionName: string) {
         && typeof e.cause === 'object'
         && (e.cause as { suppressDiscordAlert?: unknown }).suppressDiscordAlert === true
       if (e.status === 429) {
+        // Set rate-limit headers from moreInfo when available, but DO NOT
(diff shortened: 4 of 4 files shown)


subhajitlucky:
This should include the regression test before merge. The PR changes the shared `onError()` behavior, but the cited `tests/account-rate-limit.unit.test.ts` stubs its own `app.onError`, so it cannot fail if this branch regresses or if a future edit reintroduces the generic 429 body.

A focused test can call `onError(functionName)` with an `HTTPException(429, ...)` whose response/cause carries `error: "native_build_concurrency_limit_exceeded"` and `moreInfo: { activeBuilds, limit, planName }`, then assert the returned JSON preserves that code and metadata while still setting `Retry-After` / `X-RateLimit-Reset` when those fields are present. That would cover the exact behavior this PR is changing rather than relying on manual reasoning.

nagiexplorer88:
Preserving the 429 body globally appears to re-expose request data for existing `simpleRateLimit(body)` callers. `simpleRateLimit()` passes its argument through `quickError(..., moreInfo)`, and after this change `onError()` falls through to `c.json(res, e.status)`, so `res.moreInfo` becomes client-visible.

Concrete examples are `plugins/updates.ts`, which calls `simpleRateLimit(body)` when `isLimited(c, body.app_id)` is true, and `plugins/channel_self.ts`, which does the same in `parseChannelSelfPluginRequest()`. Before this PR those 429s returned the generic `{ error: "too_many_requests", message: "You are being rate limited" }`; after the PR they can echo the submitted plugin/device payload under `moreInfo`.

I think the preservation path needs either an allowlist for safe 429 causes (`native_build_concurrency_limit_exceeded`, sanitized rate-limit metadata, etc.) or the `simpleRateLimit` call sites need to stop passing raw request bodies before the generic 429 body override is removed. A regression test that throws `simpleRateLimit({ app_id, device_id })` through `onError()` would catch this.

WcaleNieWolny:
Addressed the bot's concern about reflected request bodies by curating the `moreInfo` at the two affected call sites instead of allowlisting in the handler.

- [`plugins/updates.ts`](supabase/functions/_backend/plugins/updates.ts) — `simpleRateLimit(body)` → `simpleRateLimit({ app_id: body.app_id })`
- [`plugins/channel_self.ts`](supabase/functions/_backend/plugins/channel_self.ts) — `simpleRateLimit(body)` → `simpleRateLimit({ app_id: body.app_id, device_id: body.device_id })` (matches the curated shape already used elsewhere in the same file)

Plus documented the contract on `simpleRateLimit` itself ([`utils/hono.ts`](supabase/functions/_backend/utils/hono.ts)) so future callers don't reintroduce the pattern — `moreInfo` is reflected to the client and must be curated diagnostic metadata, never the raw request body.

Chose the call-site-fix path over an allowlist in the handler because:

1. The handler stays a pure pass-through — no growing list of "safe 429 causes" to maintain.
2. Each call site declares its diagnostic intent locally, where the schema is known.
3. The 5 other `simpleRateLimit` sites in the codebase already follow this curated-object pattern; this aligns the outliers.

No behavior change for any other 429 path.

PR #2247, opened 2026-05-12
fix(updater): support encoded manifest asset downloads

riderx:
## Summary (AI generated)

- Normalize legacy percent-encoded manifest `file_name` values at `/updates` response time while keeping the response shape unchanged.
- Normalize manifest rows during `on_version_update` so future uploads from old CLI versions save raw installer paths server-side.
- Keep new CLI uploads correct by returning raw `file_name` values and URL-safe `s3_path` object keys.
- Read attachment files using same-scope decoded/raw R2 key candidates so existing encoded object keys can still download.
- Add a guarded audit/apply script for optional database cleanup, without making it part of the customer fix path.

## Motivation (AI generated)

Customers on `@capgo/capacitor-updater` 8.45.11 can fail manifest downloads for assets containing characters like `@`. The visible encoded filename in logs is only a symptom: old uploads can have encoded local manifest names and encoded object keys, while the file route may decode before the R2 lookup. This fixes the actual download and install path without changing the endpoint contract or requiring a response version bump.

After the backend deploy, old affected uploads should start returning updater-safe manifest paths without customers re-uploading, reinstalling, or shipping a native release. Old CLI users get corrected manifest rows on future uploads because the backend normalizes at insert time, and new CLI uploads remain correct by preserving raw local names. `/updates` only gets cheap string normalization per manifest row; no R2, DB, or cache lookup is added to the hot update-check path.

## Test Plan (AI generated)

- [x] `bunx vitest run tests/manifest-path-encoding.unit.test.ts tests/upload-path-encoding.unit.test.ts`
- [x] `bun run --cwd cli test:manifest-path-encoding`
- [x] `bun scripts/audit_manifest_filename_encoding.ts --help`
- [x] `bun lint:backend`
- [x] `bun lint --quiet`
- [x] `bun typecheck`
- [x] `bun run cli:typecheck`
- [x] `bun run cli:build`
- [x] `git diff --check`

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of special characters in manifest filenames, particularly the `@` symbol in file paths.
  * Improved file upload and retrieval logic to correctly process legacy and current file path encoding formats.

* **Tests**
  * Added comprehensive test coverage for manifest path encoding and special character handling.
  * Updated CLI test suite with new path encoding validation tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2247)


nagiexplorer88:
The range preflight still only checks the decoded key with `head(fileId)` before the new fallback keys are built. For a legacy object that exists only at the raw percent-encoded key, e.g. `.../sad_post_grey%402x.png`, a `Range` request will skip this file-size check because `.../sad_post_grey@2x.png` has no head result, then the later `get()` can still serve the raw fallback key.

That means out-of-range legacy range requests will not return the intended `Content-Range: bytes */<actual size>` response from lines 454-463. I would compute `rawFileId`/`candidateKeys` before the range block and run the range `head()` against the same candidate set used for reads, stopping on the first existing object. A regression test with only the `%40` object present plus `Range: bytes=<size>-` would cover this.

riderx:
Fixed in cbbd60e16: range preflight now heads the same decoded/raw candidate key set used by downloads, and the unit test covers the legacy `%40` object-only case.

PR #2258, opened 2026-05-13
fix(files): cache TUS upload access checks

riderx:
## Summary (AI generated)

- Added a tracked migration for the production `app_versions.r2_path` index that Supabase recommended for the TUS/files ready-bundle guard.
- Cached successful and ready-bundle TUS attachment write-access checks for exact API key, org, app, and file path matches so repeated `HEAD`/`PATCH` calls do not repeat the same DB work during an active upload.
- Fixed `/stats` fallback version lookup so a deleted `unknown` placeholder can be reused instead of repeatedly scheduling no-op placeholder inserts.
- Added focused regressions for the TUS write-access cache and the deleted `unknown` stats fallback.

## Motivation (AI generated)

Supabase reported heavy production volume from the ready-bundle `app_versions.r2_path` lookup. That lookup comes from `checkWriteAppAccess` in the files/TUS upload route. The repeated placeholder insert comes from `/stats` falling back to `unknown` with the wrong deleted filter, which can miss the placeholder and schedule `ensurePlaceholderVersions` again on every request.

This reduces unnecessary database pressure on hot upload and stats paths, keeps the production index tracked in migrations, and lowers the risk of DB saturation from resumable-upload loops or stats traffic. Upload integrity stays intact because TUS create requests still perform the full primary DB validation and only repeated chunk/progress checks reuse a short-lived exact-match cache.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun lint:backend`
- [x] `bun test tests/files-upload-access-cache.unit.test.ts`
- [x] `bun test tests/stats.test.ts --test-timeout 20000`
- [ ] `bunx supabase migration list --local` blocked because local Supabase Postgres is not running on `127.0.0.1:54322` in this worktree.

Generated with AI

## Summary by CodeRabbit

* **Performance Improvements**
  * Implemented caching mechanism for upload authorization decisions to minimize database queries during file transfer operations.
  * Added database index to optimize app version lookups.

* **Bug Fixes**
  * Corrected fallback behavior for deleted app versions in the statistics endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2258)

### DIFF
--- a/supabase/functions/_backend/files/files.ts
+++ b/supabase/functions/_backend/files/files.ts
@@ -7,6 +7,7 @@ import { Hono } from 'hono/tiny'
 import { app as download_link } from '../private/download_link.ts'
 import { app as upload_link } from '../private/upload_link.ts'
 import { app as ok } from '../public/ok.ts'
+import { CacheHelper } from '../utils/cache.ts'
 import { sendDiscordAlert } from '../utils/discord.ts'
 import { quickError, simpleError } from '../utils/hono.ts'
 import { middlewareKey } from '../utils/hono_middleware.ts'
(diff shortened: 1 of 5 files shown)


digzrow-coder:
The allowed-write cache looks like it can bypass the ready-bundle immutability guard during the 60s TTL.

`POST /upload/attachments` performs the primary DB checks and then stores an `allowed` cache entry for the exact `file_id`. Subsequent `HEAD`/`PATCH` requests hit `getCachedAttachmentWriteAccess()` before opening the primary connection, so an `allowed` cache hit skips the `app_versions` lookup that blocks finalized bundle paths. If the upload is finalized and an `app_versions` row is created for the same `r2_path` while the cache entry is still alive, a later resumable `PATCH` for that same path can continue through `next()` instead of returning `bundle_already_ready`.

That seems to weaken the new ready-bundle guard exactly when the state changes from upload-in-progress to finalized. I would avoid caching `allowed` across mutating `PATCH` requests, or include an upload/session state that is invalidated when the bundle becomes ready. A regression can reproduce this by doing the POST to seed `allowed`, changing the mocked ready-bundle query to return a row, then asserting a following PATCH does not reuse the allowed cache and returns 409.

PR #2262, opened 2026-05-13
[codex] hard delete old soft-deleted bundles

riderx:
## Summary (AI generated)
- Add a database cleanup function that permanently deletes soft-deleted `app_versions` after 90 days.
- Update the existing `delete_old_versions` cron task to run the cleanup daily at 03:00.
- Add backend coverage for the 90-day retention boundary and cron task wiring.
- Add a narrow typo-ignore rule to preserve the existing `files.ts` no-DB-read warning wording without modifying that file.
- Update file-read tests to assert the existing no-DB-read download contract.

## Motivation (AI generated)
Soft-deleted bundles currently stay in `app_versions` indefinitely, which leaves useless records and related storage metadata around long after the retention window. A 90-day hard-delete window keeps recovery room while clearing old deleted bundles automatically.

This reduces long-term database/storage clutter from deleted bundles and keeps cleanup automatic through the existing cron task system. The file-read test updates protect the intended high-availability download behavior.

## Test Plan (AI generated)
- [x] `typos .`
- [x] `bunx eslint tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/files-security.test.ts tests/delete-old-deleted-versions.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/delete-old-deleted-versions.test.ts`
- [x] `bun lint:deadcode`
- [x] CI `Run tests`
- [x] CI `Run Playwright tests`
- [x] CI `Run benchmarks`

## Summary by CodeRabbit

* **New Features**
  * Scheduled maintenance now permanently cleans up app versions soft-deleted for 90+ days (runs daily at 3:00 AM).

* **Tests**
  * Added integration tests covering version cleanup and cron behavior.
  * Improved unit/regression tests for file reads, local proxy behavior, and caching.

* **Chores**
  * CI workflow cleanup/robustness improvements.
  * Typo checker ignore list updated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2262)


digzrow-coder:
One thing to tighten before this runs in production: the hard-delete path can still strand partial-update manifest objects when the earlier soft-delete cleanup did not finish. `delete_old_deleted_versions()` now issues a raw `DELETE FROM public.app_versions`; that fires the async `on_version_delete` trigger, but by the time the queued handler runs, the `manifest_app_version_id_fkey ... ON DELETE CASCADE` has already removed the manifest rows. `deleteIt()` can still delete `OLD.r2_path`, but `deleteManifest()` no longer has the `manifest.s3_path` rows needed to remove the per-file delta objects or to check shared references.

That means a version whose `deleted_at` was set 90+ days ago while `on_version_update` was delayed/failed can be permanently removed from the DB and lose the last pointers to its manifest files. The new regression only asserts the `app_versions` rows disappear, so it would pass while R2 cleanup becomes impossible. Consider collecting/deleting manifest objects before the hard `DELETE`, or only hard-delete rows once their manifest rows are already gone/known-clean, with a regression that seeds a stale deleted version with manifest rows and verifies storage cleanup or retry state is preserved.

PR #2263, opened 2026-05-13
fix(backend): keep manifest cleanup on soft delete

riderx:
## Summary (AI generated)

- Keep soft-delete manifest cleanup running even when bundle deletion needs a retry or app_versions_meta is missing.
- Keep shared manifest storage safe by checking the exact s3_path before deleting a manifest object from storage.
- Add an admin dry-run/apply script to clean inactive manifest rows and matching R2 objects only when no active version references the same s3_path.
- Align stale file-read tests with the current no-database-read file availability behavior from main.

## Motivation (AI generated)

A soft-deleted bundle could leave manifest rows behind when earlier cleanup steps returned or threw before deleteManifest ran. The cleanup script gives us a controlled way to audit and remove already-stuck inactive manifest rows without changing retention policy or hard-deleting app_versions.

This helps keep public.manifest size under control, reduces wasted Supabase/R2 storage, and avoids deleting files still used by active app versions. It also preserves update availability by not introducing a database read into file downloads.

## Test Plan (AI generated)

- [x] bunx vitest run tests/on-version-update-cleanup.unit.test.ts
- [x] bunx vitest run tests/on-version-update-cleanup.unit.test.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/files-security.test.ts
- [x] bunx eslint --no-ignore scripts/cleanup_inactive_manifest.ts
- [x] bun lint:backend
- [x] bunx typos package.json scripts/cleanup_inactive_manifest.ts supabase/functions/_backend/triggers/on_version_update.ts tests/on-version-update-cleanup.unit.test.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts tests/files-security.test.ts
- [x] bun run supabase:with-env -- bun scripts/cleanup_inactive_manifest.ts --batch-size=1 --max-batches=1
- [x] bun test:backend

## Summary by CodeRabbit

* **New Features**
  * Added administrative cleanup tool to remove inactive manifest entries and unreferenced S3 objects.

* **Bug Fixes**
  * Improved S3 asset deletion logic to prevent accidental cleanup of shared files.
  * Enhanced error handling during manifest cleanup to defer and report failures.

* **Tests**
  * Updated test expectations for file caching and availability behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2263)


digzrow-coder:
I think the cleanup ordering here can permanently strand R2 objects on a transient delete failure.

Both `deleteManifest()` and the new `admin:cleanup-inactive-manifest` path delete the `manifest` DB rows before deleting the corresponding R2 keys. In the admin script, `deleteManifestRows()` commits the row deletion and then `deleteR2Objects()` runs; if any `DeleteObjectCommand` fails, the script only increments `totalR2Failed`. On the next run, those paths are no longer returned by `fetchCandidatePaths()` because the manifest rows are gone, so there is no database reference left to retry the failed object delete. The trigger path has the same shape: it deletes each manifest row first, then calls `s3.deleteObject`; a failed object delete is logged but the row has already been removed.

For production cleanup, I would avoid dropping the retry source before the external delete succeeds. One option is to select candidate rows/paths, delete R2 first for unreferenced paths, and only delete the manifest rows for paths whose object deletion succeeded, with a final active-reference check before the DB delete. At minimum this needs a regression where the first R2 delete fails and a second cleanup run can still find and retry the same path.

KCDaemon (changes requested):
Rechecked the latest head (f01ac91) and the object-lifecycle ordering issue is still present in both cleanup paths.

In `scripts/cleanup_inactive_manifest.ts`, `runApply()` calls `deleteManifestRows()` first; that function deletes `public.manifest` rows inside a committed transaction and only afterwards does `deleteR2Objects()` attempt `DeleteObjectCommand` for the returned paths. If any R2 delete fails, the script only increments `totalR2Failed` / prints the failed paths. On the next run, those paths are no longer returned by `fetchCandidatePaths()` because the manifest rows are gone, so there is no durable retry source left for the stranded objects.

The trigger path in `deleteManifest()` has the same shape for each manifest entry: delete the manifest DB row, then check for remaining references, then call `s3.deleteObject()`. A transient object-delete failure leaves the object behind after the DB row has already been removed.

`git diff --check origin/main...origin/pr-2263` passes locally, but CI currently has failing Run tests jobs. I would keep this blocked until object deletion happens before dropping the last manifest reference, or failed object deletes keep durable retry state. A regression where the first R2 delete fails and a second cleanup run can still find/retry the same path would cover the production failure mode.

digzrow-coder:
I put together a helper branch that fixes the lifecycle ordering blocker and the current typo CI failure:

https://github.com/digzrow-coder/capgo/tree/codex/capgo-pr2263-helper-forkbase

What it changes:
- `scripts/cleanup_inactive_manifest.ts` now deletes R2 objects before deleting the matching `public.manifest` rows, and only deletes DB rows for paths whose R2 deletion succeeded. Failed R2 deletes leave the manifest rows in place so the next run can find/retry them.
- `deleteManifest()` in `on_version_update.ts` now checks for other references first; if this is the last reference, it deletes the R2 object before deleting the manifest row. If the object delete fails, the row is kept as retry state.
- Fixes the `typos` failures in `supabase/functions/_backend/files/files.ts`.
- Adds regressions for both retry paths, including an admin-script test where the first R2 delete fails and the second run can still retry/delete that path.

Validated locally:
- `bunx vitest run tests/cleanup-inactive-manifest.unit.test.ts tests/on-version-update-cleanup.unit.test.ts tests/files-app-read-guard.unit.test.ts tests/files-local-read-proxy.unit.test.ts`
- `bunx eslint --no-ignore scripts/cleanup_inactive_manifest.ts tests/cleanup-inactive-manifest.unit.test.ts supabase/functions/_backend/triggers/on_version_update.ts supabase/functions/_backend/files/files.ts tests/on-version-update-cleanup.unit.test.ts`
- `bun run lint:backend`
- `bun run typecheck`
- `typos` on the changed files
- `git diff --check`

I could not run the service-backed `tests/files-security.test.ts` locally because Postgres/Supabase was not listening on `127.0.0.1:54322`; the non-service unit tests above pass.

digzrow-coder:
Follow-up: the fork CI rerun for the helper branch is green now: https://github.com/digzrow-coder/capgo/actions/runs/25904826715

digzrow-coder:
I rebased the manifest cleanup retryability fix onto current `main` because this PR is currently conflicting and the same cleanup-order bug is still present there.

Clean branch/commit:
- https://github.com/digzrow-coder/capgo/tree/codex/manifest-cleanup-retry-current
- https://github.com/digzrow-coder/capgo/commit/bf243802532cceab54d9b242e6f0bdf89016ded9

What changed:
- Manifest R2 objects are deleted before their durable `manifest` rows are removed when the deleted version is the last reference.
- If manifest object deletion fails, the row and manifest counters are kept so the trigger can retry instead of stranding storage state.
- Bundle deletion failures no longer prevent manifest cleanup from being attempted, but the trigger still throws afterward so the deleted version remains retryable.
- Same-object cleanup is serialized with a transaction-scoped advisory lock to preserve the shared-file race protection.

Validation on the clean branch:
- `bunx vitest run tests/on-version-update-cleanup.unit.test.ts`
- `bunx eslint --no-ignore supabase/functions/_backend/triggers/on_version_update.ts tests/on-version-update-cleanup.unit.test.ts`
- `bun run lint:backend`
- `bun run typecheck`
- `git diff --check` (only Git CRLF conversion warnings)

I tried opening this as a fresh PR, but GitHub shows: "An owner of this repository has limited the ability to open a pull request to users that are collaborators on this repository." This branch is ready to cherry-pick or use as the replacement for the conflicting cleanup slice.

digzrow-coder:
Follow-up: fork CI for the clean current-main branch is green now.

Run: https://github.com/digzrow-coder/capgo/actions/runs/25919249303

Completed successfully:
- `Run tests`
- `Run Playwright tests`
- `Check dead code`
- path/change detection jobs

This is the branch from my previous comment: https://github.com/digzrow-coder/capgo/tree/codex/manifest-cleanup-retry-current

PR #2265, opened 2026-05-13
[codex] Store uploaded manifest file sizes

riderx:
## Summary (AI generated)

- Add a service-role-only `uploaded_file_sizes` table populated by backend upload/storage observation, not CLI manifest payloads.
- Hydrate `manifest.file_size` from backend-observed upload records and ignore any client-provided manifest size values.
- Keep the legacy R2 HEAD fallback for old or missing upload records, then cache the observed fallback size for future manifests.
- Gate the manifest insert trigger so only rows matching a backend-observed size skip `on_manifest_create`; forged or missing sizes still queue verification.
- Isolate queue-consumer per-message failures so one thrown message becomes one failed result instead of failing the whole batch.

## Motivation (AI generated)

`on_manifest_create` was doing one follow-up R2 metadata lookup per manifest row. Large bundles can contain thousands of files, which creates queue/archive growth and risks worker limits. The fix moves size capture into backend-controlled upload handling while preserving the R2 fallback for legacy rows.

The trust boundary matters here: the CLI can send malformed or malicious manifest data, but `manifest.file_size` is now written from sizes observed by the backend upload/storage layer only.

The queue consumer should also keep draining successful messages when one message handler throws before returning an HTTP response. That failure now stays attached to the one message, while successful messages can still be deleted.

This reduces new `on_manifest_create` queue pressure for modern uploads, makes large delta bundles more reliable, and avoids turning manifest size into a client-trusted billing/storage signal. It also reduces queue blast radius when one message handler fails unexpectedly.

## Test Plan (AI generated)

- [x] `bun lint:backend`
- [x] `bun typecheck`
- [x] `bun run --cwd cli lint`
- [x] `bun run supabase:with-env -- bunx vitest run tests/queue-consumer-message-shape.unit.test.ts tests/manifest-uploaded-size.unit.test.ts tests/files-r2-error.test.ts`
- [x] `bun run supabase:db:reset`
- [x] `bash scripts/check-supabase-migration-order.sh`
- [x] `git diff --check`
- [x] GitHub CI after force-push

Generated with AI

### DIFF
--- a/cli/src/bundle/partial.ts
+++ b/cli/src/bundle/partial.ts
@@ -187,6 +187,12 @@ interface PartialEncryptionOptions {
   ivSessionKey: string
 }
 
+interface PartialUploadManifestEntry {
+  file_name: string
+  s3_path: string
+  file_hash: string
+}
+
 export async function uploadPartial(
   apikey: string,
   manifest: manifestType,
@@ -195,7 +201,7 @@ export async function uploadPartial(
   orgId: string,
   encryptionOptions: PartialEncryptionOptions | undefined,
   options: OptionsUpload,
-): Promise<any[] | null> {
+): Promise<PartialUploadManifestEntry[] | null> {
   const spinner = spinnerC()
   spinner.start('Preparing partial update with TUS protocol')
   const startTime = performance.now()
@@ -238,7 +244,7 @@ export async function uploadPartial(
     spinner.message(`Uploading ${totalFiles} files using TUS protocol`)
 
     // Helper function to upload a single file
-    const uploadFile = async (file: manifestType[number]) => {
+    const uploadFile = async (file: manifestType[number]): Promise<PartialUploadManifestEntry> => {
       const finalFilePath = join(path, file.file)
       const filePathUnix = convertToUnixPath(file.file)
 
@@ -285,7 +291,7 @@ export async function uploadPartial(
         })
       }
 
-      return new Promise((resolve, reject) => {
+      return new Promise<PartialUploadManifestEntry>((resolve, reject) => {
         spinner.message(`Prepare upload partial file: ${filePathUnix}`)
         // Get the MIME type for this file (based on original filename, not the R2 path)
         const filetype = getContentType(uploadPathUnix)
@@ -341,9 +347,9 @@ export async function uploadPartial(
       })
     }
 
-    // Process files in batches of 1000 to avoid overwhelming the server
+    // Process files in batches to avoid overwhelming the server
     const BATCH_SIZE = 500
-    const results: any[] = []
+    const results: PartialUploadManifestEntry[] = []
 
     for (let i = 0; i < manifest.length; i += BATCH_SIZE) {
       const batch = manifest.slice(i, i + BATCH_SIZE)
--- a/cli/src/types/supabase.types.ts
+++ b/cli/src/types/supabase.types.ts
@@ -2330,6 +2330,33 @@ export type Database = {
           },
         ]
       }
+      uploaded_file_sizes: {
+        Row: {
+          app_id: string
+          created_at: string
+          file_size: number
+          owner_org: string
+          s3_path: string
+          updated_at: string
+        }
+        Insert: {
+          app_id: string
+          created_at?: string
+          file_size: number
+          owner_org: string
+          s3_path: string
+          updated_at?: string
+        }
+        Update: {
+          app_id?: string
+          created_at?: string
+          file_size?: number
+          owner_org?: string
+          s3_path?: string
+          updated_at?: string
+        }
+        Relationships: []
+      }
       usage_credit_consumptions: {
         Row: {
           applied_at: string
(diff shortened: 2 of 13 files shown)


digzrow-coder:
This looks like it turns `manifest.file_size` into a client-trusted value for new uploads.

The migration changes `on_manifest_create` to only enqueue the R2 HEAD fallback when `NEW.file_size IS NULL OR NEW.file_size <= 0`, and `on_version_update` then copies `entry.file_size` into the version manifest via `normalizeManifestFileSize(entry.file_size)`. The new CLI sends the real `finalBuffer.length`, but that value is still supplied by the client/upload request path, not derived server-side from R2.

A modified CLI or direct API caller can set any small positive `file_size` for uploaded manifest rows. Because the trigger skips `on_manifest_create`, the server never re-HEADs the object and the bogus size is persisted into the app version manifest. Any downstream storage/bandwidth/accounting, quota, or audit code that trusts the manifest size can then be under-reported while the actual R2 object is larger.

I think the optimization needs a trust boundary: either keep the R2 HEAD verification for client-provided positive sizes, or only skip the trigger when the row is written by a trusted backend path that has independently measured the uploaded object size. At minimum, the DB trigger should not let arbitrary positive `file_size` bypass server-side verification.

riderx:
Addressed in the latest commit. The CLI no longer sends `file_size` in manifest entries, and `on_version_update` ignores any client-provided size value. New manifest rows are hydrated only from `public.uploaded_file_sizes`, which is populated by backend upload/storage observation or by the legacy R2 HEAD fallback. The manifest trigger also only skips `on_manifest_create` when the inserted `manifest.file_size` matches that service-role-owned observed-size row; forged or missing values still enqueue the R2 verification path.

nagiexplorer88:
One blocker remains in the forged/untrusted positive `file_size` path.

The migration now does the right first step: `trigger_verified_manifest_create()` only skips the queue when `NEW.file_size` matches `public.uploaded_file_sizes`; otherwise it enqueues `on_manifest_create` so the R2 HEAD fallback can verify it.

But the queued handler still exits immediately for any positive value:

```ts
if (record.file_size && record.file_size > 0) {
  return c.json(BRES)
}
```

So a manifest row inserted with `file_size = 123` and no matching `uploaded_file_sizes` row follows this path:

1. migration trigger enqueues it because the size is not backend-observed;
2. `on_manifest_create` receives the row;
3. `updateManifestSize()` returns before calling `getManifestSizeWithRetry()`.

That means forged or stale positive sizes are still accepted instead of being revalidated, which contradicts the migration comment that “Legacy/missing or forged records still keep the fallback R2 HEAD.” The early return needs to distinguish verified backend-observed sizes from queued/unverified positive sizes, or the queued path should ignore the existing positive value and run the R2 size check before deciding whether to keep it.

digzrow-coder (commented):
I think the trusted-size shortcut still needs to bind the observed upload row to the app/version being materialized, not only to the raw `s3_path`.

`handleManifest()` now hydrates `file_size` from `uploaded_file_sizes` with `WHERE s3_path = ANY(...)`, and `trigger_verified_manifest_create()` skips the fallback queue when any observed row has the same `s3_path`/size. But the manifest being inserted is scoped by `app_version_id` while `uploaded_file_sizes` also records `owner_org`/`app_id`; neither path verifies that the observed row's app/owner matches the `app_versions` row currently being updated. Since `app_versions.manifest` is API-key/user writable through the existing update policies, a manifest payload can point at an already-observed path outside the current app and get a trusted non-zero size without revalidation for this version.

Please either reject manifest entries whose parsed scoped path does not match `record.owner_org`/`record.app_id`, or constrain the uploaded-size lookup/trigger check through the owning `app_versions`/`apps` scope. A regression should seed an `uploaded_file_sizes` row for app A, update an app B version with that `s3_path`, and assert the manifest insert is rejected or gets `file_size = 0`/queues verification instead of taking the trusted size.

KCDaemon (changes requested):
Rechecked the latest head (eff81cb).

The trusted uploaded-size path still is not scoped tightly enough to the version/app being materialized. `getUploadedFileSizes()` loads rows only with `WHERE s3_path = ANY(...)`, and `trigger_verified_manifest_create()` skips the fallback verification when any `uploaded_file_sizes` row has the same `s3_path` and `file_size`. Neither path checks that the observed row's `owner_org` / `app_id` matches the `app_versions` row receiving the manifest.

Because the manifest payload is still user/API-key writable and `uploaded_file_sizes` stores the owning scope separately, a version can point at an already observed path from another app/org and receive a trusted non-zero `file_size` without verification for the current app version. I would constrain the lookup and trigger through the owning `app_versions`/`apps` scope, or reject entries whose parsed scoped path does not match the record being updated.

The CI checks are green and the patch applies cleanly with whitespace checking, but I would keep this blocked until cross-app/org trusted-size reuse is covered by a regression test.

RobinWitch (changes requested):
Rechecked latest head (eff81cb). The new queue trigger still does not actually revalidate untrusted positive `file_size` rows.

`trigger_verified_manifest_create()` now enqueues rows whose `file_size` does not match `uploaded_file_sizes`, which is the right direction. However, the queued payload still contains the original positive `record.file_size`, and `on_manifest_create.updateManifestSize()` immediately returns for any positive value before doing the R2 `HEAD` fallback. That means a client/API-key write with a forged positive `file_size` that is not in `uploaded_file_sizes` will be queued, but the worker will log "manifest file_size already set" and leave the forged size untouched.

Please remove the unconditional positive-size short-circuit from the worker, or include a trusted/verified flag in the queue payload and only skip R2 verification for backend-observed sizes. A regression should insert a manifest row with a positive `file_size` that has no matching `uploaded_file_sizes` row, run the trigger consumer, and assert the worker overwrites it from storage (or at least attempts the R2 size lookup) instead of returning early.

PR #2266, opened 2026-05-13
fix(plugin): read channel self get from replica

riderx:
## Summary (AI generated)

- Route old-plugin `/channel_self` PUT read requests through the read-replica Postgres client.
- Keep POST and DELETE on the primary connection because they can write `channel_devices`.

## Motivation (AI generated)

Supabase reported plugin workers still opening direct primary Postgres sessions. The legacy `/channel_self` PUT handler is a read action, so it should use the existing read-replica path instead of the write/default connection.

This reduces primary database connection pressure for old plugin channel reads without changing public API responses or legacy write behavior.

## Test Plan (AI generated)

- [x] `bun lint:backend`
- [x] `bun typecheck`
- [x] `bun run supabase:with-env -- bunx vitest run tests/channel_self.test.ts --fileParallelism=false --hookTimeout=60000`

Generated with AI

## Summary by CodeRabbit

* **Chores**
  * Updated PostgreSQL client configuration for channel self operations endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2266)

### DIFF
--- a/supabase/functions/_backend/plugins/channel_self.ts
+++ b/supabase/functions/_backend/plugins/channel_self.ts
@@ -593,7 +593,7 @@ app.put('/', async (c) => {
     return simpleRateLimit({ app_id: bodyParsed.app_id, device_id: bodyParsed.device_id, ...buildRateLimitInfo(rateLimitStatus.resetAt) })
   }
 
-  const pgClient = getPgClient(c)
+  const pgClient = getPgClient(c, true)
 
   return await runChannelSelfWithPgClient(
     c,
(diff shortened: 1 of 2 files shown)


digzrow-coder:
I think this should stay on the primary for the legacy override path, or split the handler so only the local-storage/new-plugin branch can use the replica.

`POST /channel_self` writes/removes `channel_devices` on the primary for older clients, and those clients commonly call `PUT /channel_self` right after setting the channel to resolve the effective channel. With this PR, `PUT` now uses `getPgClient(c, true)`, but `put()` still reads `getChannelDeviceOverridePg()` and `getChannelsPg()` from that client for the pre-vX.34.0 path.

That makes the immediate read-after-write path replica-lag dependent: a device can successfully self-assign to a channel, then its next `PUT` can miss the just-written override and fall back to the public/default channel. During staged rollouts this can send the device back to production or the wrong channel until replication catches up.

A narrow fix would be to keep `PUT` on the primary when `isChannelSelfLocalChannelStorageVersion(c, body, 'PUT')` is false, and only use the read replica for the new local-storage path/listing reads that do not depend on `channel_devices` read-your-write consistency.

PR #2282, opened 2026-05-18
fix: pass appid through plan-check RPC so RBAC keys can upload

WcaleNieWolny:
## Summary

- Adds a 3-arg overload (`appid` parameter) to `is_allowed_action_org_action` and `is_paying_and_good_plan_org_action`, threading `appid` into `check_min_rights`.
- Updates the CLI's `checkPlanValidUpload` to pass `appid` (it already received it but was discarding it).
- Updates the generated Supabase types to reflect the new overload.

## Bug

The CLI's `checkPlanValidUpload` calls `is_allowed_action_org_action(orgid, ['storage'])` without an `app_id`. The call chain reaches `rbac_check_permission_direct` with `p_app_id = NULL`. For an API key with `limited_to_apps` set (typical for RBAC-managed keys), the app-scope restriction block:

```sql
IF array_length(v_api_key.limited_to_apps, 1) > 0 THEN
  IF v_effective_app_id IS NULL OR NOT (v_effective_app_id = ANY(v_api_key.limited_to_apps)) THEN
    -- deny: RBAC_CHECK_PERM_APIKEY_APP_RESTRICT
```

…denies because `array_length > 0 AND v_effective_app_id IS NULL`. The CLI surfaces this RBAC denial as the misleading `"Plan upgrade required for upload"` error, even though the org's plan is perfectly healthy.

## Safety invariant preserved

A key with `limited_to_apps = [A]` still cannot perform app-B operations: `rbac_check_permission_direct` still enforces `limited_to_apps` whenever an `app_id` is supplied. Org-scope reads now correctly thread the caller's `app_id` (when known) so the restriction has the context it needs.

## Backward compatibility

Both `is_allowed_action_org_action(orgid, actions)` and `is_paying_and_good_plan_org_action(orgid, actions)` keep their existing 2-arg signatures. The new 3-arg overloads sit alongside them, so older deployed clients keep working unchanged.

## Workaround for affected users on old CLI

Recreate the API key **without** `limited_to_apps` (scope by org only), or remove the app restriction on the existing key. This unblocks them until they upgrade.

## Test plan

- [x] `bun run cli:lint` — clean
- [x] `bun run cli:typecheck` — clean
- [x] `bun run cli:build` — succeeds
- [x] `bun run cli:test` — 13/13 passing
- [ ] Manual smoke test: upload a bundle with an RBAC-managed key that has `limited_to_apps` set to confirm the upload now succeeds
- [ ] Verify the same RBAC key is still denied when calling for a different app (safety invariant)
- [ ] Confirm a key with `limited_to_apps = [A]` is still denied if attempting to use it for app B's upload (`app.upload_bundle` path remains restricted)

## Out of scope

Other call-sites of `is_allowed_action_org_action` / `is_allowed_action_org` (bandwidth, MAU, dashboard reads) may have the same trap. Worth a follow-up audit, but not bundled here to keep this PR surgical.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error handling and messaging when upload plan validation fails.

* **Improvements**
  * Plan validation now respects app-specific checks for uploads and access, enforcing API keys limited to a specific app.

* **Tests**
  * Added integration tests covering app-scoped plan checks and RBAC behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2282?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


WcaleNieWolny:
Triaged CodeRabbit's review and addressed in `fc2af0feb`:

| # | Comment | Action | Where |
|---|---|---|---|
| 1 | Outside-diff: `checkPlanValidUpload` ignores RPC error, treats falsy `validPlan` as bad plan | ✅ Fixed — destructure `error`, throw with `formatError()` so permission/migration failures are distinguishable from billing failures | `cli/src/utils.ts` |
| 2 | Types: companion overload for `is_paying_and_good_plan_org_action` missing | ✅ Fixed — added the 3-arg overload union to mirror `is_allowed_action_org_action` | `cli/src/types/supabase.types.ts` |
| 3 | Migration: REVOKE from `anon` blocks CLI which runs as anon | ✅ Fixed — replaced the REVOKE with `GRANT EXECUTE ... TO anon`, matching the existing 2-arg overload's grants (see `20260427105151_harden_security_definer_execute_grants.sql`). Without this the upload path would 403 before the function body runs. Good catch — would have shipped a broken fix. | `supabase/migrations/...passthrough_appid.sql` |
| 4 | Migration: `pg_catalog.current_setting` / `pg_catalog.now()` qualification | ⏭️ Skipped intentionally | — |

**On #4**: there's no explicit repo rule for `pg_catalog` qualification, and the function I'm overloading (`is_paying_and_good_plan_org_action` from `20260427105817`) uses bare `current_setting('role', true)` and `now()`. Matching the existing style preserves consistency with the function we're extending. Since the function declares `SET search_path = ''`, `pg_catalog` is still implicitly first in the resolution order, so the unqualified calls always resolve to the built-ins regardless of session search_path. Happy to add the prefix as a follow-up if a maintainer wants the whole RPC chain hardened in one pass.

CI re-running on `fc2af0feb`.

WcaleNieWolny:
Follow-up on types: I started a local Supabase, applied all migrations including the new one, and ran `bunx supabase gen types typescript --local`. Confirmed empirically that the generator does **not** emit the new 3-arg overload for either `is_allowed_action_org_action` or `is_paying_and_good_plan_org_action`, even though:

- Both 3-arg functions exist in `pg_proc` (verified via `\df` inside the local DB).
- PostgREST routes correctly to either overload depending on whether `appid` is in the request body (tested with curl against both shapes; both returned `false` for a bogus orgid, proving the function ran).
- `pg-meta`'s `/pg/functions` endpoint returns both overloads.
- Function grants are identical to `is_allowed_capgkey`, which **does** get its overload emitted.

Looks like a `supabase gen types` quirk specific to certain overload shapes — I couldn't determine the trigger from the surface.

Since the repo's `chore(auto-sync)` job runs the same generator, any manual overload I add to `supabase.types.ts` will be reverted on the next sync. To avoid that fragility, in `d2b554e3a` I:

1. **Reverted the manual type overloads** so `cli/src/types/supabase.types.ts` matches what the generator produces (no diff vs `origin/main`).
2. **Cast the 3-arg args to `never`** at the single call site in `cli/src/utils.ts`, so TS accepts the extra `appid` while the runtime call still routes to the 3-arg DB function via PostgREST.

This keeps the build green regardless of what the auto-sync does next, and isolates the gen-types quirk to one well-commented line of code. The DB migration is unchanged — the 3-arg overload is the source of truth.

Lint / typecheck / build / tests all green locally.

WcaleNieWolny:
Added a regression test in `tests/plan-check-appid-passthrough.test.ts` (commit `f693c3e61`). Three assertions, all running against PostgREST as anon with a `capgkey` header — the exact path the CLI takes:

1. **2-arg call denies** — original bug repro. Confirms `is_allowed_action_org_action(orgid, ['storage'])` returns `false` for an RBAC-managed key with `limited_to_apps` set, against an org with `use_new_rbac = true`.
2. **3-arg call with matching appid allows** — the fix. Confirms the appid passthrough lets RBAC's app-scope restriction find the right context.
3. **3-arg call with non-matching appid still denies** — safety invariant. Two same-org apps; the key is restricted to app A, the call asks for app B, still rejected. Proves the rejection comes from `limited_to_apps`, not from the cross-org check in `check_min_rights`.

### Validation that the test is actually testing the fix

I rolled the migration back locally (dropped the 3-arg overload) and re-ran the test. PostgREST returned `PGRST202: Could not find the function public.is_allowed_action_org_action(actions, appid, orgid)` for both 3-arg cases — i.e., the test would catch a regression if the migration were ever reverted. With the migration applied, all 3 pass.

### Test setup gotcha worth noting

The `apikeys_force_server_key` BEFORE trigger overrides whatever you `insert` for `key` with its own UUID. The plaintext is only readable from `INSERT ... RETURNING` (which runs before the DEFERRABLE INITIALLY DEFERRED `apikeys_strip_plain_key_for_hashed` trigger fires at commit). The test reads the real key back from the RETURNING row before sending it via the capgkey header.

Local: `bun run supabase:with-env -- bunx vitest run tests/plan-check-appid-passthrough.test.ts` → 3/3 passing in ~400ms.

PR #2279, opened 2026-05-17
feat(apikeys): migrate legacy keys to v2

riderx:
## Summary (AI generated)

- Migrates legacy API keys into V2 role bindings and removes V1 columns/RPC creation paths.
- Makes legacy rights, org membership checks, org listing, API-key scope checks, and notification/admin helpers RBAC-backed so `org_users.user_right` is compatibility data, not an authority source.
- Moves API key management to the global sidebar page and shows RBAC app/group access without the old `use_new_rbac` UI split.
- Preserves old CLI/client compatibility by keeping legacy RPC names, key mode inputs, invite role names, and org_users compatibility rows mapped into RBAC.

## Motivation (AI generated)

Legacy API keys and legacy org rights created two parallel permission systems. This consolidates runtime authorization on RBAC/V2 while keeping old clients working through compatibility wrappers, so users do not need to update the CLI for the migration to work.

## Business Impact (AI generated)

- Reduces API-key and org-role permission ambiguity for users with multiple organizations.
- Makes API-key management easier to discover from the sidebar.
- Removes old permission decision paths that could drift from RBAC behavior.
- Preserves existing users and old CLI versions by migrating current keys 1:1 and accepting legacy inputs through RBAC-backed wrappers.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun lint:backend`
- [x] `bun typecheck`
- [x] `bun test tests/invites.unit.test.ts`
- [ ] GitHub Actions `Run tests` / CodeQL / CodeRabbit on this PR update
- [ ] Local `bun run supabase:db:reset` could not complete in this worktree because the Docker/Supabase reset process timed out after 45s before producing SQL output; CI is being used for migration-backed validation.

PR #2302, opened 2026-05-20
feat(cli-warnings): fail fast when CLI < 7.107.0 hits the PR #2282 RBAC bug

WcaleNieWolny:
## Summary

Follow-up to #2282. That PR fixed the appid-passthrough bug on the server side and shipped the CLI fix in `@capgo/cli@7.107.0`. This PR adds a fail-fast **fatal** CLI warning for users still running older CLIs whose API keys would trigger the bug, so they see an actionable message instead of the misleading `"Plan upgrade required for upload"` billing error.

## The warning fires only when ALL are true

1. Caller's API key is **RBAC v2** (`apikeys.mode IS NULL` with role_bindings).
2. Key has `limited_to_apps` set (non-empty).
3. Org has `use_new_rbac = true`.
4. `cli_version` parses as `X.Y.Z` AND sits below `7.107.0`.

These are the exact preconditions for the [appid-passthrough denial in `rbac_check_permission_direct`](https://github.com/Cap-go/capgo/pull/2282). If any are false, the warning stays silent — including unparseable version strings like `dev`/`next` (we'd rather not nag on local builds).

## Why fatal and why now

- `checkRemoteCliMessages` runs **before** `checkPlanValidUpload` in `cli/src/bundle/upload.ts:846` vs `:850`. A fatal warning throws and stops the upload before the misleading billing error can fire.
- The bug currently surfaces as `Error: Plan upgrade required for upload` + an auto-opened plans page in the browser. Users waste time inspecting Stripe / their plan when the real issue is RBAC denying for app-context reasons.
- The replacement message names the bug, the upgrade target (`@capgo/cli@7.107.0`), and the workaround (drop `limited_to_apps` on the key, leaving only the org restriction).

## Why scoped to RBAC v2 keys only

Legacy keys (`mode IN ('read','write','upload','all')`) with `limited_to_apps` on a `use_new_rbac` org **also** hit the bug. I narrowed scope to RBAC v2 keys per the request — start tight, broaden later if support reports the same misclassification for legacy keys. Easy to extend the condition.

## How the version check works

```sql
cli_version_match := regexp_match(cli_version, '^([0-9]+)\.([0-9]+)\.([0-9]+)');
IF cli_version_match IS NOT NULL THEN
  cli_version_parts := ARRAY[m[1]::int, m[2]::int, m[3]::int];
  IF cli_version_parts < ARRAY[7, 107, 0]::int[] THEN
    -- append fatal warning
  END IF;
END IF;
```

Postgres `int[]` compares lexicographically, so `[7,106,0] < [7,107,0]` and `[7,107,0] < [7,107,0]` is false. Pre-release suffixes (`7.107.0-beta.1`) parse to `[7,107,0]` and fall the right side of the cutoff.

## What the user sees

Before (on, say, `@capgo/cli@7.106.0`):
```
Error: Plan upgrade required for upload
You need to upgrade your plan to continue to use capgo.
Upgrade here: https://capgo.app/settings/organization/plans
```
(browser opens the plans page — completely unrelated to the actual problem)

After:
```
Found 1 cli warnings for your organization.
Your CLI version (7.106.0) has a known bug affecting RBAC-managed API keys restricted to specific apps.
Uploads with this key fail with "Plan upgrade required for upload" even when your plan is healthy.
Fix: upgrade to @capgo/cli@7.107.0 or newer:
    npm i -g @capgo/cli@latest
Workaround if you cannot upgrade: remove the app restriction (limited_to_apps) on this API key, leaving only the org restriction. See https://github.com/Cap-go/capgo/pull/2282 for context.
Please fix the warnings and try again.
```

## Test plan

`tests/cli-warning-rbac-appid-bug.test.ts` covers the full condition matrix end-to-end via PostgREST (anon + `capgkey` header — same path the CLI takes):

| mode  | limited_to_apps | use_new_rbac | cli_version       | fires? |
|-------|-----------------|--------------|-------------------|--------|
| NULL  | non-empty       | true         | 7.106.0           | ✅     |
| NULL  | non-empty       | true         | 7.107.0           | ❌ (cutoff) |
| NULL  | non-empty       | true         | 7.108.0           | ❌ (above) |
| 'all' | non-empty       | true         | 7.106.0           | ❌ (legacy) |
| NULL  | empty           | true         | 7.106.0           | ❌ (no app restriction) |
| NULL  | non-empty       | false        | 7.106.0           | ❌ (RBAC off) |
| NULL  | non-empty       | true         | dev/next/empty    | ❌ (unparseable) |

- [x] `bun typecheck` — clean
- [x] `bun run cli:lint` / `cli:typecheck` / `cli:build` — clean
- [x] `bun run supabase:with-env -- bunx vitest run tests/cli-warning-rbac-appid-bug.test.ts` — 7/7 passing
- [x] Sanity: replaced the function with a no-op stub locally and re-ran the test — the positive case fails as expected; restoring the migration makes everything green again. The test actually exercises the function logic, not just confirms current behavior.

## Out of scope

- Broadening to legacy keys (separate decision).
- Updating the wiki — will follow up once this lands.
- Updating the existing PR #2282 description with a pointer here once merged.

## Summary by CodeRabbit

**New Features**
- Introduced CLI warning system that provides real-time notifications about storage capacity constraints when limits are approached or exceeded, and alerts users to known RBAC v2 compatibility issues affecting specific older CLI versions. These warnings help users identify potential deployment issues early and take corrective action to maintain system stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2302?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/supabase/migrations/20260520040202_cli_warning_rbac_appid_bug.sql
+++ b/supabase/migrations/20260520040202_cli_warning_rbac_appid_bug.sql
@@ -1,0 +1,151 @@
+-- Warn users on @capgo/cli versions older than 7.107.0 when their API key
+-- would hit the appid-passthrough RBAC bug fixed by PR #2282.
+--
+-- The bug: is_allowed_action_org_action(orgid, actions) calls check_min_rights
+-- with app_id = NULL. For an RBAC-managed key (mode IS NULL) with
+-- limited_to_apps set, on an org with use_new_rbac = true,
+-- rbac_check_permission_direct denies the call because the key is restricted
+-- to apps but no app context was provided. The CLI surfaces this as the
+-- misleading "Plan upgrade required for upload" error.
+--
+-- The fix shipped in @capgo/cli@7.107.0. For users still running older CLIs,
+-- this warning fires fatally during checkRemoteCliMessages (which runs BEFORE
+-- checkPlanValidUpload), replacing the misleading billing error with an
+-- actionable one that explains the bug, the upgrade target, and the
+-- workaround.
+--
+-- Scope: RBAC v2 keys only (mode IS NULL with role_bindings). Legacy keys
+-- (mode IN ('read','write','upload','all')) with limited_to_apps would also
+-- hit the bug on use_new_rbac orgs, but the user-requested scope is RBAC v2.
+
+CREATE OR REPLACE FUNCTION public.get_organization_cli_warnings(
+    orgid uuid,
+    cli_version text
+) RETURNS jsonb []
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = ''
+AS $$
+DECLARE
+    messages jsonb[] := ARRAY[]::jsonb[];
+    request_apikey text;
+    api_key public.apikeys%ROWTYPE;
+    fallback_app_id text;
+    has_org_read boolean;
+    org_uses_new_rbac boolean;
+    cli_version_match text[];
+    cli_version_parts int[];
+    -- Lowest @capgo/cli release that contains the PR #2282 fix.
+    fix_cli_version constant int[] := ARRAY[7, 107, 0];
+BEGIN
+    PERFORM cli_version;
+
+    has_org_read := public.cli_check_permission(
+        permission_key := public.rbac_perm_org_read(),
+        org_id := orgid
+    );
+
+    SELECT public.get_apikey_header() INTO request_apikey;
+
+    IF request_apikey IS NOT NULL AND request_apikey <> '' THEN
+        SELECT * INTO api_key
+        FROM public.find_apikey_by_value(request_apikey)
+        LIMIT 1;
+    END IF;
+
+    IF NOT has_org_read THEN
+        IF api_key.id IS NOT NULL
+            AND COALESCE(array_length(api_key.limited_to_apps, 1), 0) > 0
+        THEN
+            SELECT public.apps.app_id INTO fallback_app_id
+            FROM public.apps
+            WHERE public.apps.owner_org = orgid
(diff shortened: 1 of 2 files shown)


WcaleNieWolny:
Closing in favor of #2279.

After tracing through #2279's migration carefully, the bug this PR warns about is structurally fixed by #2279 — not just patched. The relevant chain:

1. `apikeys.limited_to_apps` column is dropped entirely.
2. The migration backfills `role_bindings` so every existing app-restricted key gets an **org-scope `org_member` binding** in addition to its app-scope binding(s).
3. `rbac_check_permission_direct` is rewritten without the `limited_to_apps` gate.

For an old CLI (< 7.107.0) calling `is_allowed_action_org_action(orgid, ['storage'])` after #2279 lands, the trace goes:
- `check_min_rights('read', user_id, orgid, NULL /* app_id */, NULL)`
- → `rbac_check_permission_direct('org.read', ..., orgid, NULL, NULL, apikey)`
- → `rbac_has_permission(apikey, rbac_id, 'org.read', orgid, NULL, NULL)`
- → finds the `org_member` binding at org scope → returns true

So the bug can no longer manifest on the old CLI's call path. The 3-arg overload from #2282 stays harmless. The warning this PR adds becomes:
- **Broken** at runtime — references `apikeys.mode` and `apikeys.limited_to_apps` which #2279 drops.
- **Unnecessary** — the condition it warns about cannot occur after #2279.

There's a separate concern about a privilege-widening side effect of #2279's migration (previously app-restricted keys gain `org.read` / `org.read_members` / `org.create_app` via the `org_member` binding). I'll raise that on #2279 directly rather than gate it on this PR.

PR #2305, opened 2026-05-20
feat: open channel previews from QR deeplinks

riderx:
## What
- QR codes now open Capgo preview deep links that point to a preview payload URL.
- The scan page accepts channel preview links, bundle preview links, direct preview web URLs, and legacy channel links.
- New preview payload endpoint returns the bundle/manifest payload from the existing preview file worker.

## Why
- Preview should download and set the exact preview payload, not call the normal updater channel endpoint and hit native-version/channel rollback rules.
- Bundle and channel preview should follow the same mobile flow.

## How
- Channel QR uses `capgo://preview/channel?...&url=<preview.json>`.
- Bundle QR uses `capgo://preview/bundle?url=<preview.json>`.
- The app fetches the JSON payload, downloads it, starts preview mode, and sets the bundle directly.
- The existing preview worker serves `/.capgo/preview.json`; no update backend or channel-assignment backend code is changed.

## Testing
- `bun test tests/preview-links.unit.test.ts`
- `bun run typecheck:frontend`
- `bun run lint`
- `bun test tests/preview-links.unit.test.ts && bun run build`
- `bunx cap copy`
- `xcodebuild -workspace ios/App/App.xcworkspace -scheme App -configuration Debug -destination generic/platform=iOS CODE_SIGNING_ALLOWED=NO`

## Not Tested
- Live physical-device scan against production preview after deployment of the preview payload endpoint.


achimala (commented):
This branch depends on updater APIs that the app does not actually ship yet. `src/pages/scan.vue` calls `getLatest({ appId, channel })` and `startPreviewSession({ appId })`, but `bun.lock` still resolves `@capgo/capacitor-updater` to `8.45.11`.

In `8.45.11`, `GetLatestOptions` only contains `channel`, Android/iOS `getLatest` only read that `channel`, and there is no native `startPreviewSession` plugin method registered. The local type cast masks this from TypeScript, but the QR flow will run against the installed native plugin: the target app id will be ignored, and devices can fall back to the old `current()/next()/setShakeMenu()` branch instead of starting a real cross-app preview session.

Please land/release the updater-side preview APIs and bump `@capgo/capacitor-updater` plus `bun.lock` here before merging this app-side QR flow. At minimum, this should have a version guard or native smoke test so the scanner cannot silently advertise cross-app previews while the bundled native plugin is still on the old API surface.

src/pages/scan.vue
```diff
@@ -133,6 +149,51 @@ async function handleBarcodeScan(scannedValue: string) {
...
+    downloadProgress.value = 0
+
+    await removeDownloadListener()
+    downloadListener = await CapacitorUpdater.addListener('download', (state: DownloadEvent) => {
+      downloadProgress.value = state.percent || 0
+    })
+
+    toast.success(`Switching to channel: ${previewLink.channelName}`)
+
+    const latest = await CapacitorUpdater.getLatest({ appId: previewLink.appId, channel: previewLink.channelName })
```
hadessunxy-code: This native preview path never sets `preview: true` on the updater request. The PR description says the scan flow should call `getLatest({ appId, channel, preview: true })`, and the backend preview behavior is gated on that flag so it can bypass the normal channel/device restrictions while still checking `allow_preview`. With the current call, scanning a channel preview deep link for another user's app will still look like a normal update check, so the main use case can fail or hit normal stats/device side effects. Can we pass `preview: true` here and add a test/fixture that proves the generated QR/deep link path requests preview mode?

hadessunxy-code (commented):
Reviewed the QR channel preview flow and found one blocker in the native scan path.

thefuturrjfhejh:
I checked the latest scan flow after the updater bump. `getLatest` uses the preview override now, but `startPreviewSession(previewLink.appId)` still runs before `download()` / `set()`.

Risk: if the download or `set` fails, the catch only shows a toast, while the native preview session has already stored the preview app id and remains active until the user leaves preview. That can leave the container using the target app id even though no preview bundle was applied.

Verification: `src/pages/scan.vue` calls `startPreviewSession` before `CapacitorUpdater.download()` and there is no cleanup call in the catch path. I think downloading first, then starting the preview session immediately before `set(bundle)`, avoids this without changing the API contract.

astrobrownmusic:
Reviewed current head 5dfdb294 with focus on the QR preview deep-link contract and the native scan entry path. I also checked that the earlier updater API compatibility concern is no longer present in this branch: package.json/bun.lock now resolve @capgo/capacitor-updater to 8.47.4, and the scan page calls the typed startPreviewSession/getLatest preview APIs instead of relying on the older fallback path for payload-backed links.

Local verification from this branch:

- bun test tests/preview-links.unit.test.ts -> 4 tests passed

The parser/builders cover payload-backed channel links, legacy web channel links, native capgo:/ variations, and bundle payload links. In scan.vue, payload-backed links go through startPreviewPayload, while the getLatest({ preview: true }) path is kept for legacy channel links without a payload URL. No blocker from this focused pass. I did not do the physical-device production scan listed under Not Tested.

PR #2308, opened 2026-05-21
fix(cli/build init): route to recovery when no profiles match this app

WcaleNieWolny:
## Summary

Two UX fixes for `capgo build init --platform ios` (import-existing path):

### 1. Empty picker → no-match-recovery (the main fix)

When the user has profiles for the identity they picked but **none match the current app's bundle ID** (or distribution mode), the picker showed:

```
Pick a provisioning profile (0 matching this app's bundle ID and app_store distribution):
(1 other profile hidden — wrong bundle ID or distribution mode)
↩️   Back to identity selection
```

That's a dead end. The user already has the `import-no-match-recovery` screen (with **Fetch from Apple**, **Create new App Store profile via Apple**, and **Open Developer Portal** options) — we just weren't routing them there.

Now the `filterProfilesForApp(profiles, appId, importDistribution)` filter runs at:
- **identity selection** — the natural choke point
- **`import-pick-profile` entry** — defense-in-depth covering Apple-fetch results, resume, and back-navigation

If nothing survives, we jump straight to `import-no-match-recovery`. The alert wording there now distinguishes the two cases so the user understands what's actually wrong:

- *No profiles at all* → "No provisioning profile on this Mac is linked to '<identity>'."
- *Profiles exist but none match* → "No provisioning profile on this Mac matches this app (**com.example.app**) for **app_store** distribution under '<identity>'."

This is what the user asked for: when the local list is empty for this app, offer to create the profile via the Apple API using their `.p8` key (already implemented; we just needed to send users there).

### 2. Dedupe consecutive identical log lines

`✔ Distribution · app_store` could appear 3× in the rendered log. `addLog` now collapses consecutive identical entries — covers this symptom plus any other repeats from resume / back-navigation.

### Refactor

Extracted the `bundleId + distribution` filter (previously inline in 3 places in `app.tsx`) as `filterProfilesForApp` in `macos-signing.ts`. Added 5 unit tests covering matching, mismatched bundleId, mismatched distribution, null/undefined distribution, and empty input.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bunx eslint src/build/onboarding/ui/app.tsx src/build/onboarding/macos-signing.ts` — only 7 pre-existing errors remain on main, none introduced by this PR
- [x] `bun test/test-macos-signing.mjs` — 23 tests pass (including 5 new `filterProfilesForApp` tests)
- [x] `bun test/test-onboarding-recovery.mjs` — all pass
- [x] `bun test/test-apple-api-import-helpers.mjs` — all pass
- [ ] Manual: run `capgo build init` on a Mac with a distribution cert whose only profiles are for a different bundle ID — verify the flow lands at `import-no-match-recovery` with the new alert wording and "Create a new App Store profile via Apple" option
- [ ] Manual: repeatedly enter / re-enter the Distribution selector — verify only one `✔ Distribution · …` line appears

## Summary by CodeRabbit

* **New Features**
  * Added iOS bundle ID detection and mismatch checks during onboarding, with a confirmation step when the app identifier appears different.
  * Improved macOS signing import with smarter certificate/profile matching, Apple-side validation, and support for selecting a local `.mobileprovision` file.
  * Updated onboarding UI with clearer progress states, better resume behavior, and improved terminal tables/input handling.
* **Bug Fixes**
  * Fixed resume routing so import flows continue from the right step based on selected distribution.
  * Improved exit hints and reduced duplicate onboarding log entries.
* **Tests**
  * Expanded coverage for bundle ID detection, signing checks, resume paths, and UI behavior.


cli/src/build/onboarding/macos-signing.ts
```diff
@@ -251,6 +251,27 @@ export function matchIdentitiesToProfiles(
...
+ * `importDistribution` is null/undefined when the user hasn't picked yet —
+ * in that case any profileType is accepted.
+ */
+export function filterProfilesForApp(
+  profiles: readonly DiscoveredProfile[],
+  appId: string,
+  importDistribution: 'app_store' | 'ad_hoc' | null | undefined,
+): DiscoveredProfile[] {
+  return profiles.filter(p =>
+    p.bundleId === appId
```
hadessunxy-code: This exact match hides valid wildcard profiles in the import flow. A profile with `application-identifier` like `TEAM.com.example.*` (or `TEAM.*`) is usable for an app such as `com.example.myapp`, especially for non-App-Store distributions, but `p.bundleId === appId` filters it out and routes the user to `import-no-match-recovery`. That recreates the empty/dead-end recovery path for users who already have a wildcard profile installed. Could we treat wildcard suffixes as matches here, at least when `importDistribution` is not `app_store`, and add a regression case alongside the new `filterProfilesForApp` tests?

hadessunxy-code (commented):
Reviewed the import-existing recovery path and found one edge case worth fixing before merge.

Lukaa91:
Heads up, I think `listProfilesForCert` only looks at the first page of profiles now, which can bite on bigger teams.

Old query was `filter[certificates]=X&limit=50`. Server side filter, so up to 50 matching profiles. The new one (`/profiles?include=certificates,bundleId&limit=200` + client side filter on `relationships.certificates.data[].id`) caps at 200 total profiles for the team, not 200 matches.

The comment right above on `listDistributionCerts` says `limit=200` is "wildly higher than the per-team cert limits, so pagination is unnecessary". Profiles don't really fit that argument though — they grow with apps × distribution types (Development, Ad Hoc, App Store, Enterprise), and any team with more than a handful of apps is going to blow past 200 pretty fast. If the matching profile happens to land on page 2 the user gets bounced to no-match recovery even though the cert is wired up to a perfectly usable profile in ASC.

Walking `links.next` until null before doing the filter would cover it. Unless there's some ordering on `/profiles` I missed that already puts the cert's profiles first?

PR #2315, opened 2026-05-21
feat(cli): generate GitHub Actions workflow + auto-push CAPGO_TOKEN

WcaleNieWolny:
## Summary

When `capgo build init` detects a GitHub remote after a successful first build, the wizard now offers a 3-option choice replacing the existing yes/no secrets upload:

- ✅  Yes — set the secrets AND create a workflow file
- 🔒  Yes — set ONLY the secrets
- ❌  No

The "No" branch then offers a `.env` export fallback so the user can wire up CI later via `gh secret set -f`, reusing the renderer from `build credentials manage`.

cc @riderx

### 1. CAPGO_TOKEN auto-push
`createCiSecretEntries` now takes an optional API key. When provided, the bundle pushed to GitHub/GitLab includes `CAPGO_TOKEN` alongside build credentials — the generated workflow can authenticate without the user manually running `gh secret set CAPGO_TOKEN` afterward.

### 2. Workflow generator (`cli/src/build/onboarding/workflow-generator.ts`)
Pure function that produces `.github/workflows/capgo-build.yml`:

- `workflow_dispatch` trigger with `platform` + `build_mode` inputs (defaults to the platform being onboarded)
- One install + one build + one `capgo build request` step, branched on package manager
- `env:` block enumerates the exact secret names that were pushed — no drift between what we set and what we reference
- Final step posts the artifact URL to `$GITHUB_STEP_SUMMARY`

**Per maintainer convention, the bun template includes BOTH `oven-sh/setup-bun@v2` AND `actions/setup-node@v4`** — bun's Node compat isn't perfect and many build pipelines still need `node` on PATH.

12 unit tests in `cli/test/test-workflow-generator.mjs` covering all four PMs, custom commands, skip-build mode, monorepo-friendly secrets enumeration, and the `workflow_dispatch`-only constraint.

### 3. Workflow writer (`cli/src/build/onboarding/workflow-writer.ts`)
Thin file-I/O wrapper. Returns `kind: 'exists'` with both contents when the target file is already present, so the wizard can show a line-count comparison and ask for explicit overwrite confirmation before clobbering.

### 4. .env export reuse (`cli/src/build/onboarding/env-export.ts`)
Reuses `renderEnvFile` from `build credentials manage` (refactored to take `({ appId, local, platform, creds })` — one minimal signature change, same comment header / `.gitignore` reminder / provisioning-map base64 fallback). Writes mode 0600, refuses to silently overwrite.

### 5. Build script picker (both wizards)
When the user picks "secrets + workflow", the wizard always prompts for which `package.json` script builds the web assets BEFORE invoking `capgo build request` — never auto-picks blindly. Lists every script in `scripts{}`, surfaces a "recommended" hint sourced from `findBuildCommandForProjectType()` when the matching script exists, plus escape hatches:

- `Type a custom command…` (text input, e.g. `make web`, `bash scripts/build.sh`)
- `Skip build step (my app is raw HTML)` for plain HTML/JS Capacitor apps

## Routing

- **GitHub-only for v1** — GitLab keeps the existing 2-option `ask-ci-secrets` flow. In the multi-target picker, picking GitHub routes to the new 3-option prompt; GitLab routes to the legacy prompt.
- `uploading-ci-secrets` branches on `setupMode`:
  - `with-workflow` → loads scripts + project-type recommendation → `pick-build-script` → `writing-workflow-file` (with overwrite confirm if file exists)
  - `secrets-only` / `undecided` (GitLab) → `build-complete`
- The "No → .env export" path is reachable only from the new 3-option prompt. `ci-secrets-target-select` "skip" still goes straight to `build-complete` (no second-chance prompt).

## What v1 deliberately doesn't do

- GitLab `.gitlab-ci.yml` generation (structurally different, follow-up)
- Push / pull_request triggers (only `workflow_dispatch` — manual until the user trusts it)
- Monorepo subdirectory detection (`working-directory`)
- Modifying / merging into existing non-Capgo workflows
- webDir verification after the build step

## Test plan

- [x] `bun run cli:check` (lint + typecheck + build + test) green locally
- [x] `bun test/test-workflow-generator.mjs` — 12 new tests pass
- [x] `bun test/test-ci-secrets.mjs` — 11 tests pass (8 existing + 3 new CAPGO_TOKEN cases)
- [ ] Manual: `capgo build init --platform ios` in a repo with a GitHub remote — confirm new 3-option prompt fires; pick "secrets + workflow"; verify a `.github/workflows/capgo-build.yml` lands with the right PM template
- [ ] Manual: same with `--platform android` — confirm the workflow defaults to `android`
- [ ] Manual: pick "secrets only" — confirm secrets are pushed but no workflow file written
- [ ] Manual: pick "No" → "Yes export .env" — confirm `.env.capgo.<appId>.<platform>` written at cwd with 0600
- [ ] Manual: re-run wizard with existing `capgo-build.yml` → confirm overwrite prompt shows line counts
- [ ] Manual: run in a repo with only a GitLab remote — confirm existing `ask-ci-secrets` 2-option flow still fires (no GitLab workflow generation)
- [ ] Manual: `bun.lock` repo — confirm bun template includes both setup-bun + setup-node

## Summary by CodeRabbit

* **New Features**
  * Expanded Android/iOS onboarding: GitHub Actions setup, secret push confirmation, workflow generation/preview/write with overwrite handling, package-manager and build-script selection, and `.env` export with overwrite/exists/empty handling
  * CI secret upload with async progress and optional API-key injection; secrets confirmation table
  * Terminal-height aware fullscreen diff viewer with keyboard exit and workflow preview telemetry
  * Workflow file writer and secure `.env` renderer (permission-tightening)

* **Tests**
  * New tests covering CI secrets, workflow generator, and diffing utilities

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2315?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


BataraSurya (commented):
I think the generated Yarn template currently straddles two Yarn generations in a way that can make the workflow fail for common Yarn Classic projects.

The install/build pieces are Classic-compatible (`yarn install --frozen-lockfile` and `yarn <script>`), and the test even notes "yarn classic invokes scripts without `run`". But the Capgo steps use `yarn dlx`, which is a Yarn Berry command; GitHub-hosted runners commonly provide Yarn 1.x unless the project enables Corepack / ships a Berry release. For a repo detected as Yarn purely from `yarn.lock`, the generated workflow can therefore reach the Capgo build step and fail with `Command "dlx" not found`.

Could this either use `npx @capgo/cli@latest ...` for the Yarn template, or explicitly enable/pin Corepack/Yarn Berry before emitting `yarn dlx`? A regression test that models a Yarn Classic workflow would catch the mismatch.

PR #2318, opened 2026-05-21
fix(script): fan out full manifest backfill batches

riderx:
## Summary (AI generated)
- Remove the storage metadata concurrency throttle from the manifest file-size backfill.
- Let each worker fan out the full fixed 1000-row candidate batch at once.
- Keep `--concurrency` accepted as a backward-compatible no-op so older commands still run.

## Motivation (AI generated)
The backfill needs to process millions of manifest rows quickly. The DB/API candidate read remains bounded at 1000 rows, but storage metadata checks should use the full batch capacity per worker instead of a smaller extra throttle.

This increases production backfill throughput so missing manifest file sizes can be corrected faster, while preserving indexed manifest.id scans and bulk DB updates.

## Test Plan (AI generated)
- [x] node --check scripts/backfill_manifest_file_sizes.mjs
- [x] bun scripts/backfill_manifest_file_sizes.mjs --help
- [x] bunx eslint --no-ignore scripts/backfill_manifest_file_sizes.mjs
- [x] git diff --check origin/main...HEAD
- [x] bun lint

## Summary by CodeRabbit

* **Chores**
  * Updated manifest file size backfill script's concurrency model to improve parallel processing efficiency of storage operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2318?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/backfill_manifest_file_sizes.mjs
+++ b/scripts/backfill_manifest_file_sizes.mjs
@@ -651,7 +651,7 @@ Options:
   --limit              Max rows to scan without --all. Default: 1000.
   --batch-size         Backward-compatible alias; candidate reads are always 1000.
   --workers            Parallel shared-cursor workers. Default: 8 for --all, 1 otherwise.
-  --concurrency        Backward-compatible no-op; each worker checks its full 1000-row batch at once.
+  --concurrency        Backward-compatible no-op; storage uses full-batch parallelism for maximum throughput.
   --storage-attempts   Storage metadata attempts per file. Default: 3.
   --db-attempts        DB read/update attempts. Default: 5.
   --start-id           Exclusive lower manifest.id bound for resume.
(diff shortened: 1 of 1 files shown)


hadessunxy-code (commented):
I checked the backfill change and the decoded-path fallback looks sensible, but I would be careful with the new fan-out behavior before merging.

`--all` defaults to 8 workers and each worker now `Promise.all`s a full 1000-row batch, so the default path can issue up to ~8,000 storage metadata requests at once. If R2/S3 starts returning transient 429/5xx responses, the retry loop will also sleep and retry those failed calls in large synchronized waves. That is a pretty big behavior change from the previous `--concurrency` cap, and it also makes the still-accepted `--concurrency` flag unable to protect production runs.

I would either keep a real storage concurrency limiter around the per-row `getObjectSizeWithRetry` calls, or make the full-batch fan-out opt-in / bounded by a new explicit high-watermark. That keeps the legacy backfill safe for production operators while still allowing the faster mode when the storage backend can absorb it.

PR #2321, opened 2026-05-21
[codex] Fix update failure log filters

riderx:
## Summary (AI generated)

- Updated the Updates statistics chart log navigation so the `fail` bucket opens logs filtered by every `*_fail` event it aggregates.
- Added repeated `action` query parameter support on the logs page so multiple filters are restored from URL state.
- Kept chart-click log navigation scoped to the clicked day.
- Switched CI Playwright frontend startup to build the app and serve it with `vite preview`, avoiding the Vite dependency optimizer hang seen in the duplicate push run.

## Motivation (AI generated)

Clicking `fail` in Updates statistics previously opened logs for only one failure action, so most users saw no matching events even though the chart counted failures from several event types. The CI frontend change keeps Playwright checks stable while preserving full browser coverage.

This makes update failure investigation reliable from the dashboard, reducing support/debug time and helping users identify real OTA update failure causes faster. The CI hardening also avoids wasting review time on unrelated Playwright startup flakes.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun typecheck`
- [x] `bun run build`
- [x] `bun build scripts/playwright-frontend-preview.ts --outdir /tmp/capgo-playwright-preview-check`
- [x] Parsed `.github/workflows/tests.yml` with the repo YAML parser

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Chart tooltips now filter logs by multiple action types simultaneously.
  * Log table filters correctly initialize from query parameters that include one or more actions.

* **Chores**
  * Updated Playwright test workflow to use a new local frontend preview runner and adjusted preview/startup automation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2321?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -374,7 +374,7 @@ jobs:
         name: Start Playwright frontend
         env:
           BACKGROUND_SERVICE_NAME: Playwright frontend
-          BACKGROUND_RUN_COMMAND: CAPTCHA_KEY='' bun run serve:worktree
+          BACKGROUND_RUN_COMMAND: bun run supabase:with-env -- bun scripts/playwright-frontend-preview.ts
           BACKGROUND_WAIT_ON: |
             http-get://localhost:5173
           BACKGROUND_LOG_PATH: ${{ runner.temp }}/playwright-frontend.log
--- a/scripts/playwright-frontend-preview.ts
+++ b/scripts/playwright-frontend-preview.ts
@@ -1,0 +1,30 @@
+import { spawnSync } from 'node:child_process'
+
+const supabaseUrl = process.env.SUPABASE_URL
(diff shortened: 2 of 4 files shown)


hadessunxy-code (commented):
Reviewed the chart-to-logs filter path. The repeated `action` query params should round-trip correctly through Vue Router, and the log table now restores all matching filters instead of only the first failure action.

No blocker from me. The one follow-up I would consider is moving the update failure action list into a shared stats/logs helper, because it now needs to stay in sync with both the aggregated `fail` statistic and `statsActions.ts`. That would reduce the chance of a future failure action being counted in the chart but omitted from the log drill-down.

PR #2323, opened 2026-05-21
fix(stats): filter legacy download_fail logs

riderx:
## Summary (AI generated)

- Filters legacy `download_fail` events from updater plugins below `7.17.0` and `6.14.25` before writing user logs.
- Reuses the same filter for fail aggregates so raw logs, update charts, and saved stats stay aligned.
- Adds coverage for ignored and accepted `download_fail` plugin versions.

## Motivation (AI generated)

Old updater plugins could report `download_fail` when there was no update to download. The aggregate fail counter already avoided this in one path, but raw user logs still saved the false failure, which could make chart drilldowns disagree with chart totals.

Users see cleaner update failure reporting and fewer false alarms from old clients. This improves trust in Capgo update analytics and avoids wasting debugging time on non-actionable failures.

## Test Plan (AI generated)

- [x] `bunx eslint supabase/functions/_backend/plugins/stats.ts tests/stats.test.ts`
- [x] `bun run lint:backend`
- [x] `bunx vitest run tests/stats.test.ts`

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined stats collection so download-failure events are only recorded for supported plugin versions, with special handling to exclude legacy v6 and other older versions.

* **Tests**
  * Added concurrent, table-driven tests to verify download-failure events are suppressed for legacy plugin versions and recorded for supported versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2323?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/supabase/functions/_backend/plugins/stats.ts
+++ b/supabase/functions/_backend/plugins/stats.ts
@@ -2,7 +2,7 @@ import type { Context } from 'hono'
 import type { MiddlewareKeyVariables } from '../utils/hono.ts'
 import type { Database } from '../utils/supabase.types.ts'
 import type { AppStats, StatsActions } from '../utils/types.ts'
-import { greaterOrEqual, parse } from '@std/semver'
+import { greaterOrEqual, parse, tryParse } from '@std/semver'
 import { Hono } from 'hono/tiny'
 import { getAppStatus, setAppStatus } from '../utils/appStatus.ts'
 import { BRES, simpleError, simpleError200, simpleRateLimit } from '../utils/hono.ts'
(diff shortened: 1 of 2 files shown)


hadessunxy-code (commented):
Reviewed the legacy `download_fail` filter. The shared helper is a good direction: it keeps saved logs and fail aggregates aligned, and the threshold tests cover both v7 and v6 boundaries.

One edge case worth checking before merge: the filter currently runs after the version lookup, so legacy `download_fail` events can still fail with `version_not_found` before they are suppressed if the noisy report references a missing/unknown target version. If old clients only ever send an existing/current `version_name`, this is fine. If not, moving the `download_fail` suppression ahead of `getAppVersionPostgres` would make the ignore behavior complete and avoid batch items turning into errors for the same legacy noise this PR is trying to drop.

PR #2332, opened 2026-05-22
feat(frontend): create RBAC API keys from apikeys page

Dalanir:
## Summary (AI generated)

- Reworked the sidebar /apikeys page to create RBAC API keys instead of legacy mode keys.
- Added RBAC role display for API keys and scope rendering for organizations/apps based on RBAC bindings.
- Updated the creation modal to select multiple manageable organizations by default and configure app-level RBAC bindings.
- Extended DialogV2 width options for the wider RBAC creation modal.

## Motivation (AI generated)

Martin is migrating legacy API keys to RBAC API keys, so the global API keys page needs to support the RBAC creation model instead of the old read/upload/write/all mode flow.

This keeps API key management aligned with the RBAC migration and lets users create organization-scoped and app-scoped keys from the central API keys page without falling back to legacy permissions.

## Test Plan (AI generated)

- [x] Ran `bunx eslint src/pages/ApiKeys.vue src/components/DialogV2.vue src/stores/dialogv2.ts playwright/e2e/apikeys.spec.ts`
- [x] Ran `bun typecheck`
- [x] Ran `bunx playwright test playwright/e2e/apikeys.spec.ts`
- [x] Ran `bun test:front`; observed 18/19 passing locally, with the remaining failure caused by login schema-cache retry before reaching /apikeys.

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * API keys now support role-based access control with per-organization and optional per-application bindings.

* **User Interface Changes**
  * API key list shows assigned role and aggregated org/app scope.
  * Key creation/regeneration dialog updated to collect RBAC role and per-app bindings; copy-key failures now open a dialog.
  * Dialogs support larger size options.

* **Localization**
  * Added messages for permission denial and "Role per organization".

* **Tests**
  * End-to-end API key tests updated for the RBAC flow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2332?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


BataraSurya (commented):
I think the new global create-key modal is missing the same anti-escalation filtering that the org-scoped RBAC API-key editor already applies.\n\nIn src/pages/settings/organization/ApiKeys.[id].vue, orgRoleOptions is filtered by the caller's current org-role priority (
.priority_rank <= callerOrgPriorityRank.value), so the UI only offers roles that the caller can actually grant. Here, src/pages/ApiKeys.vue builds orgRoleOptions from all assignable org roles except the explicit unsupported ones. For a user who can manage API keys in multiple orgs but has a lower-ranked role in at least one selected org, the modal can offer a role choice that createRoleBindingForPrincipal will reject on the backend. Because this create call is atomic across all selected orgs, that turns into a full create failure after the user submits, rather than a disabled/hidden role option up front.\n\nCould this compute the allowed org-role choices from the selected organizations, probably using the minimum caller priority across selectedOrgsForCreation, and filter orgRoleOptions the same way the org-scoped editor does? If a selected org makes the current role invalid, the modal should also clear or downgrade selectedOrgRole before submit.

PR #2333, opened 2026-05-23
fix(storage): add R2 bucket migration workflow

riderx:
## Summary (AI generated)

- Add split upload/download/fallback R2 bucket bindings for the files Worker.
- Add matching Supabase S3 bucket routing via `S3_UPLOAD_BUCKET`, `S3_DOWNLOAD_BUCKET`, and `S3_FALLBACK_BUCKET`.
- Add `scripts/migrate_r2_bucket.mjs` to copy only active DB-referenced bundle and manifest objects to a clean R2 bucket using server-side copy, resumable worker cursors, verification, and failed CSV output.

## Motivation (AI generated)

R2 contains significantly more data than the active bundle storage counted by the database, which indicates stale/orphaned objects from previous cleanup failures. We need a safe migration path that stops new writes to the dirty bucket first, copies only keys still referenced by Capgo, verifies the clean bucket, then switches downloads without local object downloads.

This reduces storage waste and gives Capgo an operationally safe way to move production traffic to a clean R2 bucket while preserving download availability during the migration. It also reduces future debugging risk by making upload/download bucket routing explicit and reversible during staged migrations.

## Test Plan (AI generated)

- [x] `bun run lint:backend`
- [x] `bun run typecheck`
- [x] `bun scripts/migrate_r2_bucket.mjs --help`
- [x] `bun scripts/migrate_r2_bucket.mjs --phase deploy-upload --source-bucket capgo --target-bucket capgo-clean-test --target prod`
- [x] `bun scripts/migrate_r2_bucket.mjs --phase deploy-download --source-bucket capgo --target-bucket capgo-clean-test --target prod`
- [x] `bun scripts/migrate_r2_bucket.mjs --phase create-bucket --source-bucket capgo --target-bucket capgo-clean-test --target prod`
- [x] `bun scripts/migrate_r2_bucket.mjs --phase deploy-final --source-bucket capgo --target-bucket capgo-clean-test --target prod`

Generated with AI

## Release Notes

* **New Features**
  * Added support for multiple attachment buckets (upload, download, and fallback) enabling better resource management and redundancy.
  * Added migration tool (`admin:migrate-r2-bucket`) to facilitate data transfer between buckets.

* **Chores**
  * Updated bucket configuration to support new bucket bindings across all environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2333?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


scripts/migrate_r2_bucket.mjs
```diff
@@ -0,0 +1,1151 @@
...
+    await runPlan()
+  }
+  else if (phase === 'create-bucket') {
+    await createTargetBucket()
+  }
+  else if (phase === 'deploy-upload') {
+    deployStage('upload')
+  }
+  else if (phase === 'deploy-download') {
+    deployStage('download')
```
GabrielSotto, line 1118: Could we put the same verification guard on the standalone download switch?

`--phase all --apply` runs `runVerifyAfterAll()` before `deployStage('download')`, but `--phase deploy-download --apply` reaches this branch and switches the worker/Supabase bucket config immediately. Since `deploy-download` is exposed as a standalone phase, an operator can promote the target bucket before the target has passed the full DB-key verification. The source fallback reduces outage risk, but it also means an incomplete target could be silently promoted and rely on fallback.

Could `deploy-download` either run the same verification gate before applying, require a persisted successful `verify` result for this source/target pair, or require an explicit force flag with a warning?

Assisted by Codesx.

PR #2349, opened 2026-05-28
fix(rbac): restore channel API key permissions

riderx:
## Summary (AI generated)

- Restores channel creation for write/developer API keys by granting `app.create_channel` to `app_developer`.
- Replaces channel insert/update/delete RLS with exact RBAC permission checks.
- Updates CLI channel add/set/delete prechecks to use exact RBAC permissions instead of coarse legacy admin checks.
- Tightens public channel endpoints so service-role writes are gated by exact channel permissions first.

## Motivation (AI generated)

A migrated API key could be reported as `write` by the legacy CLI permission helper even when it had channel/admin RBAC permissions. That made channel create/delete/self-set flows fail before the request reached the real RBAC/RLS checks.

This restores customer channel management for migrated API keys without making write keys destructive: developer/write keys can create and update allowed channel settings, while channel deletion still requires `channel.delete`. It also keeps older CLI compatibility through the existing RPC shape.

## Test Plan (AI generated)

- [x] `bun run supabase:db:reset`
- [x] SQL sanity check: one `channels` RLS policy per operation and `app_developer` has `app.create_channel`
- [x] `bun run cli:lint`
- [x] `bun run lint:backend`
- [x] `bun run cli:typecheck`
- [x] `bun run typecheck:frontend`
- [x] `bun run lint`
- [x] `bun run cli:build`
- [x] `bun run supabase:with-env -- bunx vitest run tests/channel-post.unit.test.ts tests/rbac-permissions.test.ts tests/hashed-apikey-rls.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/cli-channel.test.ts tests/cli-hashed-apikey.test.ts tests/channel.test.ts`

Generated with AI

### DIFF
--- a/RBAC_SYSTEM.md
+++ b/RBAC_SYSTEM.md
@@ -19,32 +19,25 @@ This document explains in detail the Capgo RBAC (Role-Based Access Control) perm
 
 ## Overview
 
-Capgo uses a **hybrid** system that supports two permission management modes:
+Capgo uses RBAC (Role-Based Access Control) as the single authorization system.
+Legacy `org_users` rows still exist as compatibility data for older clients and
+invite flows, but authorization is resolved through `role_bindings` and
+permission checks.
 
-### Legacy System (old)
-- **Main table**: `org_users`
-- **Simple roles**: `super_admin`, `admin`, `write`, `upload`, `read`
-- **Limitation**: one role per user per organization
-- **Granularity**: limited, no control at individual app/channel level
-
-### RBAC System (new)
+### RBAC System
 - **Main tables**: `roles`, `permissions`, `role_bindings`, `role_permissions`
 - **Multiple roles**: a user can have multiple roles at different scopes
 - **Fine granularity**: permissions at org, app, channel, and bundle level
 - **Flexibility**: add/modify permissions without code changes
 
-### Automatic Switching
+### Compatibility
 
-The system automatically switches between legacy and RBAC via:
-- **Org-level flag**: `use_new_rbac` column in the `orgs` table
-- **Global flag**: `rbac_settings` table (singleton) to enable RBAC for all orgs
-- **Auto-detection**: the `rbac_is_enabled_for_org()` function checks both flags
+The old RBAC rollout switches have been removed. `rbac_is_enabled_for_org()`
+is retained only for old code paths and always returns `true`.
 
 ```sql
--- The org uses RBAC if:
--- 1. orgs.use_new_rbac = true OR
--- 2. rbac_settings.use_new_rbac = true
 SELECT rbac_is_enabled_for_org('123e4567-e89b-12d3-a456-426614174000');
+-- true
 ```
 
 ---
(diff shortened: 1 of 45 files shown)


42Clownfish (commented):
Reviewed the channel RBAC/API-key compatibility path. No blocker from my pass.

What I checked:
- `post.ts` now separates create vs update: new channels require `app.create_channel`, existing channels require `channel.update_settings`, and version changes require `channel.promote_bundle` after resolving the target version.
- The migration mirrors that split in RLS for insert/update/delete, so the old CLI compatibility helpers only get callers to the endpoint/RLS gate; they do not become the final authorization decision.
- The regression coverage hits the important security cases: app developer channel creation, promote denial on direct channel version changes, hashed API-key scope, and forged app/channel binding rejection.
- Current checks are green, including backend lint/typecheck, SQL/RLS tests, CLI tests, CodeQL, Sonar, and Socket.

One non-blocking edge to keep an eye on: `post.ts` uses the caller's API-key client to discover `existingChannel`. If a channel exists but is not visible to that key, the code can classify the request as a create attempt and then rely on the upsert/RLS conflict to reject it. That still looks protected, but it may produce a noisier database error instead of the cleaner `cannot_access_app` response. If that matters for CLI UX, a service-role existence lookup followed by the same explicit permission checks would make the error path deterministic.

sanphandinh (commented):
Reviewed the channel RBAC split and I found one edge case worth tightening before merge.

`post.ts` now derives:

```ts
const nextVersion = body.version && !isInternalVersionName(body.version)
  ? await findVersion(...)
  : null
```

Then, for an existing channel, it checks `nextVersion !== existingChannel.version` before allowing the update without `channel.promote_bundle`.

That means an update request which only changes channel settings and omits `version` is treated as a version change to `null` whenever the existing channel currently has a bundle version. In practice, an API key with `channel.update_settings` but not `channel.promote_bundle` can still be blocked from changing non-version settings on any versioned channel.

This may be intentional if the API contract says “omitted version clears the channel version,” but the PR summary says developer/write keys can update allowed channel settings while promote remains separate. To keep that split precise, I would either:

- preserve `existingChannel.version` when `body.version === undefined`, and only set `null` when the caller explicitly asks to clear it, or
- add a regression test showing that a versioned channel settings-only update without promote permission is intentionally rejected.

The current tests cover promote denial on direct version changes, but I did not see the inverse case: a settings-only update on an existing versioned channel with `channel.update_settings` and without `channel.promote_bundle`.

RobinWitch (changes requested):
There is a compatibility gap in the old CLI `channel delete --delete-bundle` path. This PR intentionally makes the legacy `get_org_perm_for_apikey()` return `perm_admin` for `app.create_channel` / `channel.update_settings` so old CLI builds can reach the new RLS checks, and the tests document that an `app_developer` key now gets legacy `perm_admin` while `developer_can_delete_channel` remains false.

That is safe for create/update because the channel write happens at the authoritative RLS gate, but `cli/src/channel/delete.ts` performs the optional bundle deletion before it attempts the channel delete. With `--delete-bundle`, the flow is:

1. `checkAppExistsAndHasPermissionOrgErr(..., OrganizationPerm.admin)` passes because the developer/update key maps to legacy `perm_admin`.
2. `findBundleIdByChannelName()` reads the linked bundle.
3. `deleteAppVersion()` updates `app_versions.deleted = true`; for API keys the app_versions UPDATE policy only requires the app upload permission.
4. Only after that does `delChannel()` hit the new channel DELETE RLS, which correctly requires `channel.delete` and rejects the same key.

So a key with upload/channel-update permissions but no `channel.delete` can soft-delete the channel's linked bundle, then fail to delete the channel, leaving the channel pointing at a deleted bundle. Please either move the bundle deletion until after a successful explicit `channel.delete` authorization, require the appropriate delete permission before `deleteAppVersion()`, or avoid mapping non-delete channel permissions to legacy admin for this delete path. A regression for an app_developer/write API key running `channel delete --delete-bundle` would cover this.

PR #2354, opened 2026-05-28
[codex] Clarify Capacitor expert call copy

riderx:
## Summary (AI generated)

- Updated the plans page expert service copy to describe it as a Capacitor app help call.
- Clarified that the $250/hour service covers app crashes and native issues, separate from Capgo product support.

## Motivation (AI generated)

The previous wording could make users think the paid call was Capgo product support. The new copy makes the service scope explicit before users book.

This reduces support confusion and sets clearer expectations for paid expert services without changing plan pricing or product support entitlement.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bunx vitest run tests/i18n-fallback.unit.test.ts`


42Clownfish (commented):
Reviewed the copy-only diff. No blocker: the new strings make the paid call scope clearer and separate it from Capgo product support, while preserving the existing pricing signal and translation key names. CI is green, and the change is limited to messages/en.json, so I do not see runtime or i18n fallback risk here.

PR #2355, opened 2026-05-28
[codex] fix credit-only billing for former subscribers

riderx:
## Summary (AI generated)

- Treat orgs with active usage credits and no active subscription entitlement as credit-only during plan checks.
- Consume credits against zero included usage for credit-only orgs, so former subscribers do not keep using stale plan allowances before credits are debited.
- Skip the immediate failed-payment Bento event when an org already has active usage credits.
- Add a regression test for former subscribers whose usage is under the old plan limit but should still consume credits.
- Refresh generated read-replica schema and add a scoped typo exclusion for an existing committed migration filename required by CI.

## Motivation (AI generated)

Former subscribers could buy credits after stopping subscription payments, but the cron plan check still used their old Stripe `product_id` limits as the baseline. That meant credits were not consumed until usage exceeded the old plan, and users could still receive payment-related nudges even though they had active credits available.

This makes prepaid credits usable for returning or failed-payment customers, reduces incorrect payment emails, and preserves expected access for customers who have already bought credits.

## Test Plan (AI generated)

- [x] `bun run lint:backend`
- [x] `bun run supabase:with-env -- bunx vitest run tests/credit-only-billing.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/plugin-credits-flag.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/credit-only-billing.test.ts tests/plugin-credits-flag.test.ts`
- [x] `typos .`
- [x] `bash scripts/check-supabase-migration-order.sh`
- [x] `MAIN_SUPABASE_DB_URL=postgresql://postgres:postgres@127.0.0.1:56572/postgres bun run readreplicate:check-schema`
- [x] Commit hook typechecks: `bun run cli:typecheck && bun run typecheck:backend && bun run typecheck:frontend`
- [x] GitHub CI: Run tests, Run Playwright tests, CLI tests, CodeQL, SonarCloud, Socket, DeepSec, dead code, and CLI path checks

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failed payment handling for organizations with active usage credits by conditionally managing payment failure tracking based on credit availability.

* **New Features**
  * Enhanced billing logic to better support credit-only mode organizations with improved credit overage limit management and notification handling.

* **Tests**
  * Added integration test suite for credit-only billing scenarios, validating credit consumption and overage tracking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2355?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


42Clownfish (commented):
Reviewed the credit-only billing change. No blocker from my pass.

What I checked:
- The entitlement split is conservative: trial time must be in the future, succeeded subscriptions remain plan-entitled, and failed/former subscriptions with active usage credits move into the zero-included-usage credit path.
- `userAbovePlan` only forces limits to zero in credit-only mode, so active subscribers keep the normal plan limits while former subscribers consume prepaid credits against actual usage.
- The failed-payment webhook path now skips the Bento failed-payment event only when the org has active usage credits; database status update still runs afterward.
- The new regression test covers the important case: a former subscriber under the old MAU plan limit still gets a credit consumption and overage event with `limit: 0`.
- CodeRabbit's earlier test-harness comments appear addressed in the latest commit, and current checks are green.

Residual risk I would keep in mind after merge: this relies on `orgs.has_usage_credits` being fresh when the Stripe failed-payment event arrives. If the computed flag can lag behind a just-created credit grant, the email suppression path could still fire once. I do not see that as a blocker for this PR, but it is the operational edge I would monitor.

PR #2361, opened 2026-05-29
[codex] Add Supabase Cloudflare network restriction helper

riderx:
## Summary (AI generated)

- Added `scripts/update_supabase_network_restrictions.sh` to apply current Cloudflare IPv4 and IPv6 CIDRs to Supabase database network restrictions.
- Added `bun run supabase:allow-cloudflare` as the package script entrypoint.
- Included dry-run, append, IPv4-only, IPv6-only, extra CIDR, profile, and env-file project ref support.

## Motivation (AI generated)

Capgo needs an operational way to close direct Postgres access to Supabase while still allowing Cloudflare Hyperdrive to reach the database through Cloudflare egress ranges.

This improves database security posture by making it easier to restrict Supabase Postgres connectivity to Cloudflare-controlled paths, reducing direct exposure while preserving Hyperdrive-based operations.

## Test Plan (AI generated)

- [x] `bash -n scripts/update_supabase_network_restrictions.sh`
- [x] `bun -e 'JSON.parse(await Bun.file("package.json").text()); console.log("package.json ok")'`
- [x] `bun run supabase:allow-cloudflare -- --project-ref xvwzpoazmxkqosrdewyv --dry-run --ipv4-only`
- [x] `bun run supabase:allow-cloudflare -- --project-ref xvwzpoazmxkqosrdewyv --dry-run --ipv6-only`
- [x] `bun run typecheck`

Generated with AI

### DIFF
--- a/scripts/update_supabase_network_restrictions.sh
+++ b/scripts/update_supabase_network_restrictions.sh
@@ -1,0 +1,297 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Update Supabase DB network restrictions with current Cloudflare IP ranges.
+# By default this REPLACES existing restrictions with Cloudflare IPv4 + IPv6 ranges.
+# Use --append to keep existing restrictions and add Cloudflare ranges.
+
+CF_IPS_URL="https://api.cloudflare.com/client/v4/ips"
+ENV_FILE="internal/cloudflare/.env.prod"
+PROJECT_REF="${SUPABASE_PROJECT_REF:-}"
+PROFILE="${SUPABASE_PROFILE:-}"
+APPEND="false"
+DRY_RUN="false"
+YES="false"
+INCLUDE_IPV4="true"
+INCLUDE_IPV6="true"
+EXTRA_CIDRS_TEXT=""
+
+usage() {
+  cat <<'EOF'
+Usage:
+  scripts/update_supabase_network_restrictions.sh [options]
+
+Updates Supabase database network restrictions to allow Cloudflare IP ranges.
+This is the firewall control needed when Postgres should only be reachable from
+Cloudflare Hyperdrive and other Cloudflare egress.
+
+Default behavior:
+  - Replaces existing database allowed CIDRs with Cloudflare IPv4 + IPv6 ranges.
+  - Parses the Supabase project ref from SUPABASE_PROJECT_REF or the env file.
+
+Options:
+  --project-ref <ref>    Supabase project ref. Defaults to SUPABASE_PROJECT_REF
+                         or parses MAIN_SUPABASE_DB_URL / SUPABASE_URL from env.
+  --env-file <path>      Env file to parse when --project-ref is omitted
+                         (default: internal/cloudflare/.env.prod).
+  --profile <name>       Supabase CLI profile.
+  --append               Append Cloudflare ranges instead of replacing existing restrictions.
+  --ipv4-only            Only apply Cloudflare IPv4 CIDRs.
+  --ipv6-only            Only apply Cloudflare IPv6 CIDRs.
+  --extra-cidr <cidr>    Add one extra CIDR. Can be repeated.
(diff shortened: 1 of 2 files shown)


sanphandinh (commented):
Reviewed this from the ops/security angle.

What I checked:
- `bash -n scripts/update_supabase_network_restrictions.sh` passes.
- `scripts/update_supabase_network_restrictions.sh --help` renders cleanly.
- Dry run with a real project ref fetched Cloudflare ranges and did not execute Supabase writes:
  `bash scripts/update_supabase_network_restrictions.sh --project-ref xvwzpoazmxkqosrdewyv --dry-run --yes`
  It produced 22 CIDRs and printed the expected `network-restrictions get` and `network-restrictions update --db-allow-cidr ...` commands.
- The approach matches Cloudflare Hyperdrive docs: Hyperdrive connects using Cloudflare IP address ranges, and those ranges are the right allowlist input for database firewalls.

No blocker from my pass. One small hardening suggestion before relying on the env-file fallback in automation: `get_env_value()` returns the raw value after `KEY=`, so a common `.env` line such as `SUPABASE_URL="https://xvwzpoazmxkqosrdewyv.supabase.co"` keeps the surrounding quotes. `extract_project_ref_from_url()` then fails the host regex for URL-style values because the host ends with `"`. Database URLs using the `postgres.<ref>` username path still work, but API URL fallback can fail. It would be safer to strip one layer of matching single/double quotes, and optionally ignore inline comments, before passing the value to `extract_project_ref_from_url()`.

Given the default mode is replace, I also like that the script prints the current restrictions and requires confirmation unless `--yes` is set.

scripts/update_supabase_network_restrictions.sh
```diff
@@ -0,0 +1,297 @@
...
+  update_args+=(--db-allow-cidr "$cidr")
+done
+
+echo "==> Supabase project ref: $PROJECT_REF"
+echo "==> CIDR count: ${#FINAL_CIDRS[@]}"
+echo "==> Mode: $([[ "$APPEND" == "true" ]] && echo 'append' || echo 'replace')"
+echo "==> Dry run: $DRY_RUN"
+
+echo "==> Current Supabase network restrictions:"
+run_supabase network-restrictions get --project-ref "$PROJECT_REF" || true
```
jinhongliang991013, line 270: Fail closed when the current restrictions cannot be read. This is the only preflight before the default replace operation, but `|| true` lets the script proceed after any GET failure. I reproduced this with a stub where `network-restrictions get` exits 42 while `update` succeeds: the script still executed the replacing update and exited 0. In `--yes` automation, a transient GET/API failure can therefore overwrite existing non-Cloudflare CIDRs without ever displaying them. The same suppression at line 295 also prints `Done` when post-update verification fails. Please require a successful preflight in replace mode and propagate the final verification failure.

jinhongliang991013 (commented):
Reviewed the update path with the pinned Supabase CLI behavior and a failing preflight stub. One safety issue can make the default replace mode proceed without a valid view of the existing allowlist.

RobinWitch (changes requested):
Blocking issue: this helper enables Supabase database network restrictions for Cloudflare ranges while the repo still deploys Supabase Edge Functions and those functions still open direct Postgres connections through getPgClient with MAIN_SUPABASE_DB_URL/SUPABASE_DB_URL. Supabase documents Network Restrictions as blocking direct database access from Edge Functions once restrictions are enabled: https://supabase.com/docs/guides/platform/network-restrictions#limitations. Running bun run supabase:allow-cloudflare -- --yes can therefore take the deployed Supabase functions offline for every path that uses the shared pg.ts client. Please gate/refuse this helper until the Supabase Function direct-DB paths are migrated away from direct Postgres, or require an explicit acknowledgement that enabling restrictions will break those functions; this is not only a replace-vs-append concern because enabling the restriction itself is the incompatibility.

riderx:
we cant do that with github action it will be a mess for now

PR #2373, opened 2026-05-30
fix(frontend): show bundles in dependency compare picker (gate on native_packages, not manifest)

WcaleNieWolny:
## Problem

On a bundle's **Dependencies** tab, the "Compare with bundle" dropdown is empty for many apps, so you can't compare two bundles' native dependencies at all.

The compare selector was originally built for the **Manifest** tab and later extracted into a shared `BundleCompareSelect` component and reused on the **Dependencies** tab (#1557). It carried over the Manifest tab's eligibility filter — `manifest_count > 0` — to all three of its candidate queries (latest / deploy-history / search).

That filter is correct for manifest diffs (a bundle with no manifest has nothing to diff), but **wrong for dependency diffs**:

- The Dependencies tab compares `native_packages` (a `jsonb[]` column on `app_versions`).
- `native_packages` is populated on upload **independently** of the manifest. `manifest`/`manifest_count` is only populated for bundles using the partial-update system.
- So any app whose bundles were uploaded without partial updates has `manifest_count = 0` on every bundle — and the dependency compare picker filters them all out, even though every bundle has perfectly comparable native packages.

### Reproduction / impact (real data)

For one affected app with 13 active bundles, every bundle has `manifest_count = 0` and 3 native packages each:

| Picker filter | Bundles offered |
|---|---|
| `manifest_count > 0` (old) | **0** → empty dropdown |
| `native_packages IS NOT NULL` (new) | **12** |

## Fix

Add a `compareMode` prop (`'manifest' | 'dependencies'`, default `'manifest'`) to `BundleCompareSelect`. A single `buildCompareBaseQuery()` helper now gates candidates on the column each tab actually diffs:

- **Manifest tab** → `manifest_count > 0` (unchanged behavior)
- **Dependencies tab** → `native_packages IS NOT NULL`

The Dependencies page passes `compare-mode="dependencies"`; the Manifest page is untouched and keeps its existing (correct) behavior via the default.

## Scope

- `src/components/bundle/BundleCompareSelect.vue` — add `compareMode` prop + `buildCompareBaseQuery()` helper; route all three candidate queries (latest, deploy-history, search) through it.
- `src/pages/app/[app].bundle.[bundle].dependencies.vue` — pass `compare-mode="dependencies"`.

No schema, backend, or API changes. The actual dependency comparison logic on the page is unchanged — this only fixes which bundles are eligible to appear in the picker.

## Test plan

- [ ] On a bundle whose app has multiple bundles with native packages but `manifest_count = 0`, open **Dependencies** → "Compare with bundle" now lists other bundles (latest + searchable by name/ID).
- [ ] Selecting a bundle shows the native-package diff (changed vs unchanged) as before.
- [ ] **Manifest** tab compare dropdown behaves exactly as before (still gated on `manifest_count > 0`).
- [ ] An app with only one bundle still shows an empty compare list (nothing to compare).

## Summary by CodeRabbit

* **New Features**
  * Added mode-aware bundle comparison that filters comparison candidates by either manifest data or native-package dependencies, showing only relevant bundle versions per mode.
  * Dependencies page now uses the dependencies comparison mode so native-package dependency comparisons surface appropriate versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2373?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/bundle/BundleCompareSelect.vue
+++ b/src/components/bundle/BundleCompareSelect.vue
@@ -232,15 +232,17 @@ async function searchCompareVersions(term: string) {
 
   const requestId = ++compareSearchRequestId.value
   compareSearchLoading.value = true
-  const baseQuery = buildCompareBaseQuery()
-    .neq('id', props.currentVersionId)
 
+  // Each chain must start from its own builder: the Supabase query builder is
+  // mutable and returns `this`, so reusing one instance across concurrent
+  // chains would leak filters between the requests.
   const numericId = Number(term)
   let data: VersionRow[] | null = null
   let error: unknown = null
 
   if (Number.isNaN(numericId)) {
-    const response = await baseQuery
+    const response = await buildCompareBaseQuery()
+      .neq('id', props.currentVersionId)
       .ilike('name', `%${term}%`)
       .order('created_at', { ascending: false })
       .limit(5)
@@ -253,11 +255,13 @@ async function searchCompareVersions(term: string) {
   }
   else {
     const [nameResponse, idResponse] = await Promise.all([
-      baseQuery
+      buildCompareBaseQuery()
+        .neq('id', props.currentVersionId)
         .ilike('name', `%${term}%`)
         .order('created_at', { ascending: false })
         .limit(5),
-      baseQuery
+      buildCompareBaseQuery()
+        .neq('id', props.currentVersionId)
         .eq('id', numericId)
         .order('created_at', { ascending: false })
         .limit(5),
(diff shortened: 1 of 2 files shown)


WcaleNieWolny:
Addressed the CodeRabbit "outside diff range" finding on `BundleCompareSelect.vue` (`searchCompareVersions`): the mutable Supabase query builder was reused across the two concurrent chains in the numeric-search `Promise.all`, which could leak filters between requests. Fixed in 850f256a6 by building a fresh `buildCompareBaseQuery()` (each re-applying `.neq('id', currentVersionId)`) per chain. No inline thread existed to resolve (the flagged lines fell outside the PR diff).

PR #2409, opened 2026-06-03
fix: use released updater preview loader

riderx:
## What
- Bump @capgo/capacitor-updater to 8.47.7.
- Refresh iOS SPM and Android Gradle plugin paths to the released updater package.
- Remove the temporary Capgo-side SplashScreen preview loader workaround from the PR diff.

## Why
- The updater plugin now owns the native preview transition loader, so Capgo should use the released plugin instead of keeping app-side loader glue.

## How
- Rebased the PR onto current main.
- Installed @capgo/capacitor-updater@^8.47.7 with Bun.
- Ran Capacitor sync for iOS and Android.

## Testing
- bun run build
- bun typecheck
- bun lint
- Commit hook typecheck

## Not Tested
- Manual native camera/deeplink preview flow on a device in this turn.

### DIFF
--- a/android/capacitor.settings.gradle
+++ b/android/capacitor.settings.gradle
@@ -279,7 +279,7 @@ include ':capgo-capacitor-textinteraction'
 project(':capgo-capacitor-textinteraction').projectDir = new File('../node_modules/.bun/@capgo+capacitor-textinteraction@8.0.25+1b808583819c2ac6/node_modules/@capgo/capacitor-textinteraction/android')
 
 include ':capgo-capacitor-updater'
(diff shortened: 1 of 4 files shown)


riderx:
CI note: both `Run tests` jobs fail during `bun typecheck` before frontend typecheck runs, on unrelated backend organization `onboarding` type errors in `supabase/functions/_backend/public/organization/post.ts` and `tests/organization-put-stripe-sync.unit.test.ts`.

This PR only changes `src/pages/scan.vue` and `src/services/loader.ts`. Local scoped checks passed: `bun run lint`, `bun run typecheck:frontend`, and `bun run build`. CI Playwright, CodeQL, DeepSec, dead-code, benchmarks, and security checks passed. Keeping this draft until the backend typecheck blocker is cleared.

PR #2411, opened 2026-06-03
fix: allow preview QR without manifest

riderx:
## What
- Allow bundle and channel preview pages to render the preview QR/native start even when the current bundle has no manifest.
- Keep the browser iframe preview gated behind manifest availability and non-encrypted bundles, because the iframe still depends on file-by-file preview serving.
- Add a `browserPreview` option to `BundlePreviewFrame` so the same component can render QR-only when the iframe cannot work.

## Why
- The mobile preview flow now uses the preview payload/direct bundle download path and does not require manifest entries. Blocking the whole preview page on `manifest_count` prevented users from scanning a valid QR.

## How
- Changed route-level preview eligibility to only require the app/channel/bundle and `allow_preview`.
- Passed `browserPreview=false` when the selected bundle cannot support web iframe preview.
- In QR-only mode, the component hides the phone iframe and external preview button but keeps QR generation and native start.

## Testing
- `bun run lint`
- `bun run typecheck:frontend`
- `bun run build`

## Not Tested
- `bun run typecheck` is currently blocked on latest `main` by unrelated backend organization `onboarding` type errors in `supabase/functions/_backend/public/organization/post.ts` and `tests/organization-put-stripe-sync.unit.test.ts`.
- Native device scan from a no-manifest bundle was not run locally in this worktree.


riderx:
CI note: both `Run tests` jobs fail during `bun typecheck` on the existing backend organization `onboarding` type mismatch in `supabase/functions/_backend/public/organization/post.ts` and `tests/organization-put-stripe-sync.unit.test.ts`, before frontend typecheck runs.

This PR only changes `src/components/BundlePreviewFrame.vue`, `src/pages/app/[app].bundle.[bundle].preview.vue`, and `src/pages/app/[app].channel.[channel].preview.vue`. Local checks passed: `bun run lint`, `bun run typecheck:frontend`, and `bun run build`. CI Playwright, CodeQL, DeepSec, dead-code, benchmarks, and security checks passed. Keeping draft until the backend blocker is cleared.

PR #2428, opened 2026-06-04
feat(dashboard): default channel compatibility warning banner

WcaleNieWolny:
## What

A small, **frontend-only** banner on the app dashboard that warns when the bundle currently live on the app's **default download channel** is OTA-incompatible with the **immediately previous** bundle on that channel.

## How (deliberately minimal)

- Reads the default channel = first `channels` row with `public = true` (same selection `ReleaseBanner`/`DeploymentBanner` use).
- Reads the **two most recent `deploy_history` rows** for that channel → current vs previous `version_id`.
- Loads `native_packages` for both from `app_versions` and runs the existing `comparePackages` / `summarizeCompatibility` (`src/services/bundleCompatibility.ts`).
- If incompatible, shows one amber banner with a link to the existing dependency-diff page (`/app/:app/bundle/:current/dependencies?compare=:previous`).

No backend, no migration, no new table, no RPC. Only the current-vs-previous comparison on the default channel.

## Files
- `src/components/dashboard/CompatibilityBanner.vue` (new)
- `src/pages/app/[app].vue` (mount beside `ReleaseBanner`)
- `messages/en.json` (3 keys)

## Notes / edge cases (all → no banner)
- App has no public/default channel, or fewer than 2 deploy_history entries.
- The two latest deployments point to the same version.
- Either bundle is missing `native_packages`.

## Test plan
- [ ] Default channel whose latest deploy changed native deps → amber banner shows; CTA opens the dependency diff with `?compare=`.
- [ ] Latest deploy is native-compatible with the previous → no banner.
- [ ] App with <2 deploy_history rows / no public channel → no banner.

## Summary by CodeRabbit

* **New Features**
  * Added a compatibility banner to the app dashboard that automatically detects native dependency mismatches between bundle versions. When incompatibilities are detected, a warning banner displays showing affected bundle information and provides quick access to the dependency comparison details.


WcaleNieWolny:
CleanShot 2026-06-04 at 10 39 02@2x

PR #2454, opened 2026-06-08
fix(db): separate mfa and email otp checks

riderx:
## Summary (AI generated)

- Update `public.verify_mfa()` so MFA-gated RLS/admin checks rely only on Supabase AAL session assurance.
- Add `public.verify_email_otp_auth()` as the separate first-factor OTP auth-method helper.
- Make `private/verify_email_otp` call `verify_email_otp_auth()` with the verified OTP session before recording `email_otp_verified_at`.
- Add pgTAP, endpoint, unit, and execute-grant regression coverage for the split behavior.

## Motivation (AI generated)

Email OTP can be a first-factor `aal1` login method, while completed MFA must be represented by `aal2`. Keeping those checks in one function let OTP first-factor sessions satisfy MFA-gated paths for users with enrolled MFA. Splitting the semantics keeps expected `aal1` access for users without MFA while preventing OTP first-factor sessions from being treated as MFA.

This preserves org 2FA enforcement and platform-admin MFA expectations, reducing account-takeover blast radius for customers who enable 2FA or depend on admin impersonation protections. The email OTP reauthentication flow remains supported, but now records its marker only after the OTP-specific helper validates the returned OTP session.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun lint:backend`
- [x] `bun run supabase:with-env -- bunx vitest run tests/verify-email-otp-auth-session.unit.test.ts tests/verify-email-otp.test.ts`
- [x] `bunx supabase test db --local --workdir .context/supabase-worktrees/b3ffd30a supabase/tests/00-supabase_test_helpers.sql supabase/tests/58_test_mfa_session_otp_split.sql`
- [x] `bunx supabase test db --local --workdir .context/supabase-worktrees/b3ffd30a supabase/tests`
- [x] `bun run supabase:with-env -- bunx vitest run tests/security-definer-execute-hardening.test.ts`
- [x] `bun test:backend`
- [x] `bun typecheck`
- [x] EXPLAIN check for the MFA factor lookup stayed bounded by `user_id`: local plan used an index-backed scan on `auth.mfa_factors`, `Execution Time: 0.082 ms`, `Buffers: shared hit=1`.

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Added email OTP authentication verification with session validation during the verification flow.

* **Tests**
  * Added comprehensive test coverage for OTP authentication verification and MFA session handling.

### DIFF
--- a/src/types/supabase.types.ts
+++ b/src/types/supabase.types.ts
@@ -4787,6 +4787,7 @@ export type Database = {
         Args: { plain_key: string; stored_hash: string }
         Returns: boolean
       }
+      verify_email_otp_auth: { Args: never; Returns: boolean }
       verify_mfa: { Args: never; Returns: boolean }
     }
     Enums: {
--- a/supabase/functions/_backend/private/verify_email_otp.ts
+++ b/supabase/functions/_backend/private/verify_email_otp.ts
@@ -4,7 +4,7 @@ import { createHono, getClaimsFromJWT, middlewareAuth, parseBody, quickError, si
 import { cloudlog } from '../utils/logging.ts'
 import { clearFailedAccountAuth, isAccountRateLimited, recordFailedAccountAuth } from '../utils/rate_limit.ts'
 import { buildRateLimitInfo } from '../utils/rateLimitInfo.ts'
-import { emptySupabase, supabaseAdmin } from '../utils/supabase.ts'
+import { emptySupabase, supabaseAdmin, supabaseClient } from '../utils/supabase.ts'
 import { version } from '../utils/version.ts'
 
 const bodySchema = type({
@@ -16,6 +16,15 @@ const bodySchema = type({
 export const app = createHono('', version)
 
 app.use('/', useCors)
+
+export async function verifyEmailOtpAuthSession(c: Parameters<typeof supabaseClient>[0], accessToken: string) {
+  const { data, error } = await supabaseClient(c, `Bearer ${accessToken}`).rpc('verify_email_otp_auth')
+
+  return {
+    verified: data === true,
+    error,
+  }
+}
 
 app.post('/', middlewareAuth, async (c) => {
   const rawBody = await parseBody<{ token?: string, token_hash?: string, type?: 'email' | 'magiclink' }>(c)
@@ -102,6 +111,25 @@ app.post('/', middlewareAuth, async (c) => {
   if (!verifyData.user?.id) {
     await recordFailedAccountAuth(c, auth.userId)
     return quickError(500, 'no_user', 'No user associated with OTP')
+  }
+
+  const emailOtpAuth = await verifyEmailOtpAuthSession(c, verifyData.session.access_token)
+  if (emailOtpAuth.error) {
+    cloudlog({
+      requestId: c.get('requestId'),
+      context: 'verify_email_otp - OTP session auth method check errored',
+      error: emailOtpAuth.error.message,
+    })
+    return quickError(500, 'otp_auth_check_failed', 'OTP session verification check failed')
+  }
+
+  if (!emailOtpAuth.verified) {
+    await recordFailedAccountAuth(c, auth.userId)
+    cloudlog({
+      requestId: c.get('requestId'),
+      context: 'verify_email_otp - OTP session auth method check failed',
+    })
+    return quickError(401, 'invalid_otp_auth', 'OTP session verification failed')
   }
 
   await clearFailedAccountAuth(c, auth.userId)
--- a/supabase/functions/_backend/utils/supabase.types.ts
+++ b/supabase/functions/_backend/utils/supabase.types.ts
@@ -4787,6 +4787,7 @@ export type Database = {
         Args: { plain_key: string; stored_hash: string }
         Returns: boolean
       }
+      verify_email_otp_auth: { Args: never; Returns: boolean }
       verify_mfa: { Args: never; Returns: boolean }
     }
     Enums: {
--- a/supabase/migrations/20260608114543_split_mfa_session_and_email_otp_checks.sql
+++ b/supabase/migrations/20260608114543_split_mfa_session_and_email_otp_checks.sql
@@ -1,0 +1,64 @@
+-- Split MFA session assurance from email OTP first-factor checks.
+-- `aal2` is the source of truth for completed MFA. Email OTP can be an `aal1`
+-- login method, so it must not satisfy the MFA gate used by RLS/admin checks.
+
+CREATE OR REPLACE FUNCTION public.verify_mfa()
+RETURNS boolean
+LANGUAGE sql
+STABLE
+SECURITY DEFINER
+SET search_path = ''
+AS $$
+  SELECT
+    array[(SELECT COALESCE(auth.jwt()->>'aal', 'aal1'))] <@ (
+      SELECT
+        CASE
+          WHEN count(id) > 0 THEN array['aal2']
+          ELSE array['aal1', 'aal2']
+        END AS aal
+      FROM auth.mfa_factors
+      WHERE (SELECT auth.uid()) = user_id
+        AND status = 'verified'
+    );
+$$;
+
+COMMENT ON FUNCTION public.verify_mfa() IS
+'Returns true when the current session satisfies Supabase MFA assurance. Users with verified MFA factors require aal2; users without verified factors may use aal1 or aal2.';
+
+ALTER FUNCTION public.verify_mfa() OWNER TO postgres;
+REVOKE ALL ON FUNCTION public.verify_mfa() FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION public.verify_mfa() TO anon;
+GRANT EXECUTE ON FUNCTION public.verify_mfa() TO authenticated;
+GRANT EXECUTE ON FUNCTION public.verify_mfa() TO service_role;
+
+CREATE OR REPLACE FUNCTION public.verify_email_otp_auth()
+RETURNS boolean
+LANGUAGE sql
+STABLE
+SET search_path = ''
+AS $$
+  WITH jwt_claims AS (
+    SELECT auth.jwt() AS claims
+  ),
+  amr AS (
+    SELECT
+      CASE
+        WHEN pg_catalog.jsonb_typeof(claims->'amr') = 'array' THEN claims->'amr'
+        ELSE '[]'::jsonb
+      END AS entries
+    FROM jwt_claims
+  )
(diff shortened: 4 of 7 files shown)


WcaleNieWolny (changes requested):
## ⚠️ This breaks `log_as` (admin impersonation) for users who have 2FA enabled

Removing the `amr.method = 'otp'` branch from `verify_mfa()` closes a real bypass, but that branch is **load-bearing for the admin spoof** (`private/log_as`). As written, this PR locks platform admins out of exactly the 2FA accounts they most need to support, and `log_as.ts` isn't touched to compensate.

### How `log_as` actually works

1. Verify the **calling admin** is a platform admin → `is_platform_admin()` → checks the *admin's own* `verify_mfa()` (admin has real TOTP → `aal2`, fine).
2. Mint a session **for the target user**: `supabaseAdmin.auth.admin.generateLink({ type: 'magiclink' })` → `verifyOtp({ type: 'email' })`.
3. Return that JWT to the frontend, which `setSession`s it. The admin is now browsing as the target user.

The session minted in step 2 is **single-factor**, so its claims are:

- `aal = aal1` (the target user's TOTP was never completed — the admin can't complete it)
- `amr = [{ method: 'otp' }]`

### Why removing the branch breaks it

The target user's `apps`, `orgs`, `channels`, etc. carry the **RESTRICTIVE** policy `USING (verify_mfa())` ("Prevent non 2FA access"). For a 2FA-enrolled user, the impersonation session must pass `verify_mfa()` or the admin sees an empty account.

| | aal1 + `amr.otp`, user has TOTP |
|---|---|
| **Old `verify_mfa()`** | first branch fails, **`OR (amr has 'otp')` → TRUE** ✅ spoof works |
| **New `verify_mfa()` (this PR)** | first branch fails, no otp branch → **FALSE** ❌ RLS denies every row |

This isn't incidental — the otp branch was added in commit `292ad2741` **"fix: 2fa for admins"** specifically to make impersonation of 2FA users work.

### Why the obvious workarounds don't help

- **"Just disable magic-link sign-in in Supabase"** — Supabase exposes magic link and email OTP as one `/otp` provider; there's no toggle to disable first-factor magic-link login while keeping email OTP. And the app *needs* that endpoint on: `emailOtp.ts` (`sendEmailOtpVerification` → `signInWithOtp`) and `ManageTwoFactor.vue` use it for MFA-enrollment / sensitive-action step-up — the very `email_otp_verified_at` flow this PR hardens. So the same endpoint is both the attack vector and a required feature; you can't disable it.
- **Can't tell impersonation from abuse by `amr`** — both a real first-factor email-OTP login and a `log_as` impersonation token are `aal1` + `method: 'otp'`. They differ only by **provenance**: `log_as` mints server-side via service-role `generateLink`, a path the public can't reach.

### Suggested fix (ship alongside the branch removal)

Mark the impersonation session explicitly so `verify_mfa()` can trust *impersonation* without trusting *any* otp session. A session-bound marker is the tightest, lowest-footprint option (no auth hook, no Supabase config change):

**1. `log_as.ts` — record the minted session (service_role):**
```ts
const sessionId = JSON.parse(atob(jwt.split('.')[1])).session_id
await supabaseAdmin.from('impersonation_sessions').insert({
  session_id: sessionId,
  target_user_id: user_id,
  admin_user_id: callerId,
  expires_at: new Date(Date.now() + 60 * 60_000).toISOString(),
})
```

**2. `verify_mfa()` — replace the removed otp branch with a narrowly-scoped, trusted one:**
```sql
OR EXISTS (
  SELECT 1 FROM public.impersonation_sessions s
  WHERE s.session_id = (auth.jwt() ->> 'session_id')
    AND s.expires_at > now()
)
```

**3.** Lock `impersonation_sessions` behind deny-all RLS, `GRANT` to `service_role` only — so the public `signInWithOtp` flow can never forge it.

Why this beats a custom access-token hook here:
- Bypass is bound to the **one** `session_id` `log_as` minted — not to `target_user_id`, so the real user's own `aal1` sessions never inherit it.
- `session_id` is stable across refresh and auto-expires via `expires_at` — no global hook re-firing on every login/refresh to reason about.
- Provenance enforced in the code path (service-role insert), not a remote console toggle that can be silently flipped back on.

### TL;DR

The `amr.otp` branch is simultaneously the vulnerability **and** the mechanism behind admin impersonation of 2FA users. Removing it without giving `log_as` a trusted, distinguishable marker ships an admin-impersonation regressin for 2FA accounts. `log_as.ts` + `verify_mfa()` need to change together.

Heyzerohey (commented):
### Code Review: hardens split MFA and OTP session verification

Thanks for this PR! The split of MFA session assurance from email OTP first-factor checks is a great security improvement. Here are two optimizations that simplify the implementation and improve performance.

---

### 1. Optimize `verifyEmailOtpAuthSession` by verifying JWT locally (TypeScript-level)

Since the `accessToken` is returned directly from a successful `supabase.auth.verifyOtp` call, we already know the token is authentic and signed. Instead of making an extra network HTTP RPC roundtrip to PostgreSQL via PostgREST to check the `amr` claim, we can decode the claims locally using the existing `getClaimsFromJWT` utility.

This **saves a database network roundtrip** (saving 15–50ms latency) and **completely eliminates the database function `verify_email_otp_auth`** from migrations, security-definer audit lists, and TypeScript type updates.

#### Proposed TypeScript Refactoring
In `supabase/functions/_backend/private/verify_email_otp.ts`:

```typescript
import { getClaimsFromJWT } from '../utils/hono.ts'

export async function verifyEmailOtpAuthSession(c: Parameters<typeof getClaimsFromJWT>[0], accessToken: string) {
  const claims = await getClaimsFromJWT(c, accessToken)
  const amr = claims?.amr
  
  // Verify that the JWT authentication methods references (amr) contains 'otp'
  const verified = Array.isArray(amr) && amr.some((elem: any) => elem?.method === 'otp')
  
  return {
    verified,
    error: claims ? null : new Error('Invalid token claims'),
  }
}
```

If we adopt this local check:
- We can completely remove `verify_email_otp_auth` from `supabase/migrations/20260608114543_split_mfa_session_and_email_otp_checks.sql`.
- We can remove the changes in `tests/security-definer-execute-hardening.test.ts`.
- We don't need to update `supabase.types.ts`.

---

### 2. Optimize SQL `verify_mfa()` using `EXISTS`

If we keep `verify_mfa()` in SQL, we can optimize the `count(id) > 0` and array comparison pattern. Using `EXISTS` allows the database planner to short-circuit index scanning as soon as the first verified factor is found, which is more performant than scanning/counting all rows.

#### Proposed SQL Optimization
In `supabase/migrations/20260608114543_split_mfa_session_and_email_otp_checks.sql`:

```sql
CREATE OR REPLACE FUNCTION public.verify_mfa()
RETURNS boolean
LANGUAGE sql
STABLE
SECURITY DEFINER
SET search_path = ''
AS $$
  SELECT 
    CASE 
      WHEN EXISTS (
        SELECT 1 
        FROM auth.mfa_factors 
        WHERE user_id = auth.uid() 
          AND status = 'verified'
      ) THEN COALESCE(auth.jwt()->>'aal', 'aal1') = 'aal2'
      ELSE true
    END;
$$;
```

#### Why this is cleaner:
- **Short-circuiting:** `EXISTS` is the standard SQL pattern for checks like this.
- **Simpler Logic:** Avoids array construction (`array['aal2']`) and array containment operators (`<@`), making the function much more readable.
- **Redundancy:** Removes the redundant `(SELECT auth.uid())` subquery wrapper in the `WHERE` clause.

Heyzerohey (commented):
### Follow-up: Resolving the `log_as` (Admin Impersonation) bypass cleanly

To resolve @WcaleNieWolny's concern about breaking admin impersonation for 2FA-enabled accounts, we need a session-bound trust marker. Here is a complete implementation that defines the `impersonation_sessions` table, integrates it into `log_as.ts`, and updates the optimized `verify_mfa()` function.

---

### 1. Database Migration: Create `impersonation_sessions` table
We lock down the table entirely so only `service_role` (and `verify_mfa()` running under `SECURITY DEFINER`) can read/write it:

```sql
CREATE TABLE public.impersonation_sessions (
  session_id uuid PRIMARY KEY,
  target_user_id uuid NOT NULL,
  admin_user_id uuid NOT NULL,
  expires_at timestamp with time zone NOT NULL,
  created_at timestamp with time zone DEFAULT now() NOT NULL
);

-- Enable RLS and deny all public access
ALTER TABLE public.impersonation_sessions ENABLE ROW LEVEL SECURITY;

REVOKE ALL ON public.impersonation_sessions FROM PUBLIC;
GRANT SELECT, INSERT, UPDATE, DELETE ON public.impersonation_sessions TO service_role;
```

---

### 2. TypeScript: Record Impersonation Session in `/log_as`
In `supabase/functions/_backend/private/log_as.ts`, after generating and verifying the target user's OTP, we record the session:

```typescript
  const jwt = authData.session?.access_token
  const refreshToken = authData.session?.refresh_token

  if (!jwt) {
    throw simpleError('no_jwt', 'No jwt', { authData })
  }

  // Record impersonation session to bypass MFA check in verify_mfa()
  try {
    const payload = JSON.parse(atob(jwt.split('.')[1]))
    const sessionId = payload.session_id
    const callerId = c.get('auth')?.userId
    
    if (sessionId && callerId) {
      await supabaseAdmin.from('impersonation_sessions').insert({
        session_id: sessionId,
        target_user_id: user_id,
        admin_user_id: callerId,
        expires_at: new Date(Date.now() + 60 * 60_000).toISOString(), // 1 hour expiration
      })
    }
  } catch (err) {
    throw simpleError('record_impersonation_session_error', 'Failed to record impersonation session', { err })
  }

  return c.json({ jwt, refreshToken })
```

---

### 3. SQL: Trust impersonation sessions in optimized `verify_mfa()`
We update the optimized `verify_mfa()` to allow either `aal2` or a valid impersonation session:

```sql
CREATE OR REPLACE FUNCTION public.verify_mfa()
RETURNS boolean
LANGUAGE sql
STABLE
SECURITY DEFINER
SET search_path = ''
AS $$
  SELECT 
    CASE 
      WHEN EXISTS (
        SELECT 1 
        FROM auth.mfa_factors 
        WHERE user_id = auth.uid() 
          AND status = 'verified'
      ) THEN (
        COALESCE(auth.jwt()->>'aal', 'aal1') = 'aal2'
        OR EXISTS (
          SELECT 1 
          FROM public.impersonation_sessions s
          WHERE s.session_id = (auth.jwt()->>'session_id')::uuid
            AND s.expires_at > now()
        )
      )
      ELSE true
    END;
$$;
```

PR #2455, opened 2026-06-08
feat(cli): self-update prompt + auto-restart in build onboarding

WcaleNieWolny:
## What

`capgo build init` (the build onboarding wizard) now detects when a newer `@capgo/cli` is available and offers to update as the **first screen of the wizard**, styled exactly like the "Which platform do you want to set up?" picker. On accept it updates and **re-execs** the CLI with the same arguments; the new process resumes from the persisted onboarding step, so there's no manual "re-run onboarding" step.

Previously the build onboarding did no version check at all (only the OTA `init` showed a passive "you're outdated" banner).

## How

The prompt is a real **Ink** screen (`ui/update-prompt.tsx`, `UpdatePrompt`) — same `Header`, same cards↔list responsiveness, same ←/→/Enter driving as `PlatformPicker`. The shell renders it first when `updateInfo` is set, **before** platform selection and even before the `--platform` auto-load (gated `useEffect`). "Update" exits Ink with a new `update-requested` outcome; "Skip" continues on the current version. Either choice transitions away, unmounting the `@inkjs/ui` Select — which avoids its onChange-refires-while-mounted loop.

The actual install + re-exec runs **after Ink tears down** (it needs the primary buffer + stdio inheritance, which it can't get while Ink owns the alt-screen). Strategy depends on how the running process resolved:

- **Project dependency (Path A):** walk from cwd up to the workspace root (`findRoot`, which delegates to Capacitor's own monorepo tooling) to find the `package.json` that *declares* `@capgo/cli` — app sub-package **or** root — bump it (preserving `^`/`~`, leaving `latest`/`workspace:` tags alone), run `<pm> install` **at the workspace root**, then re-exec the freshly-installed bin via `process.execPath <bin> <args>` with the **original cwd**. PM-agnostic, Windows-safe.
- **Ephemeral / npx-style (Path B):** nothing on disk to mutate → re-exec `npx -y @capgo/cli@latest …` (or `bunx` / `pnpm dlx` / `yarn dlx`). The `-y` is npx-only and precedes the package spec.

**No global install.** Re-exec carries `CAPGO_SKIP_UPDATE_PROMPT=1` to prevent an update→relaunch loop; the registry check is timeout-bounded (3s) so a slow/offline registry never stalls startup; and any failure falls back to a manual-update hint instead of silently continuing.

Monorepos work for free: `getPMAndCommand()` detects the PM from the workspace-root lockfile, `findRoot()` is workspace-aware (npm/yarn/pnpm workspaces + Nx + Lerna), install runs at the root, and re-exec keeps the original app cwd.

## Files

- `cli/src/build/onboarding/ui/update-prompt.tsx` — new Ink prompt (+ pure `updatePromptKeyAction`)
- `cli/src/build/onboarding/ui/shell.tsx` — renders the prompt first; `update-requested` exit + gated auto-load
- `cli/src/build/onboarding/command.ts` — resolve `updateInfo` before render; run install + re-exec post-exit
- `cli/src/build/onboarding/self-update.ts` — `checkForCliUpdate()` (data) + `runUpdateAndReexec()` (install/re-exec)
- `cli/src/build/onboarding/types.ts` — `update-requested` outcome
- `cli/test/test-self-update.mjs`, `cli/test/test-update-prompt.mjs` — tests (registered in `package.json`)

## Test plan

- [x] `bun run test:self-update` (26) + `bun run test:update-prompt` (8) pass — monorepo declaration walk, hoisted bin resolution, range-bump preservation, A-vs-B classification, per-PM command construction, key mapping, and a real shell render asserting the prompt is the first screen / precedes `--platform` auto-load / is absent when up to date.
- [x] `oxlint` clean
- [x] `tsc --noEmit` clean (full `bun run build` green)
- [ ] Manual: run an outdated local build (`8.0.6` vs npm `8.0.7`) of `capgo build init` in a Capacitor project — confirm the Ink update screen appears, Update → installs + relaunches into onboarding; Skip → continues.
- [ ] Manual: Path A in a workspace monorepo; Path B via `npx`.

> The live spawn/re-exec path itself isn't covered by an automated test (needs a real TTY + project + network), consistent with the repo's other `spawn` flows; everything around it is.

## Summary by CodeRabbit

* **New Features**
  * Onboarding now offers an “update available” prompt as the first screen and can install and re-run an updated CLI; falls back to a manual update hint if auto-update fails.
  * Onboarding outcome includes an explicit “update requested” result when the user accepts the update.

* **Tests**
  * Added end-to-end and unit tests covering self-update prompts, version detection, dependency resolution, and re-exec/update strategies.


WcaleNieWolny:
Addressed the review feedback in 3bb612f1:
- **Gated the prompt to the real `build init` entrypoint** (new `enableSelfUpdate`) — `onboardingBuilderCommand` is also reached from `bundle upload` (launch-onboarding) and `build credentials manage`, where a re-exec would have replayed the wrapper command. Those paths no longer prompt.
- **Signal-aware exit**: a signal-killed re-exec/install now exits `128+signal` instead of masking it as `0`.
- **Install failure handling**: a signal-interrupted `<pm> install` is now treated as failure, and the rewritten `package.json` is restored on any install failure.
- **Runner-aware manual hint** (no longer hardcodes npx).
- **De-flaked** the render test (poll for first painted frame).

@coderabbitai review

PR #2462, opened 2026-06-08
[codex] Add actor attribution to audit logs

riderx:
## Summary (AI generated)

- Added audit log actor snapshot fields for user, API key, and automated/system actions.
- Removed audit log foreign keys to users and orgs so retained audit rows survive deleted resources.
- Updated the audit log trigger, webhook payloads, API schema, generated types, UI display, and tests.

## Motivation (AI generated)

Audit logs previously only stored a user id, which made it hard to distinguish direct user actions from API key actions or automated backend actions. The old foreign keys could also erase or mutate attribution when dependent resources were deleted.

This improves incident investigation and customer support by making audit trails clearer and more durable. Teams can identify whether an action came from a user, an API key id/name, or automation without exposing API key secrets.

## Test Plan (AI generated)

- [x] `bun lint`
- [x] `bun typecheck`
- [x] `bun run supabase:with-env -- bunx vitest run tests/audit-logs.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/audit-logs.test.ts tests/webhook-delivery-security.unit.test.ts tests/webhook-delivery-redirect.unit.test.ts tests/webhook-queue-processing.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/organization-api.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/chart-refresh-rpc.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/webhook-signature.test.ts`
- [x] `bun run supabase:with-env -- bunx vitest run tests/bundle.test.ts`
- [ ] `bun test:all` was run twice locally; each run had one unrelated parallel-suite failure, and each failed file passed when rerun alone.

Generated with AI

## Summary by CodeRabbit

- **New Features**
  - Audit logs now identify the source of each action, including users, API keys, and automated system activity.
  - Audit log details display relevant user email, user ID, API key ID, and API key name information.
  - Webhook audit events now include actor attribution metadata.

- **Bug Fixes**
  - Audit records retain actor identity snapshots even when related users or organizations are deleted.
  - Deleted organization IDs are prevented from being reused.

---

> [!NOTE]
> **High Risk**
> Touches audit triggers, org lifecycle, webhook dispatch, and account deletion—security-sensitive paths with broad behavioral changes; migration alters constraints and production audit semantics.
> 
> **Overview**
> **Audit logs** now record **who** performed an action via snapshot fields (`actor_type`, user email/id, API key id/name) instead of relying only on `user_id`. The **audit log UI** replaces the email column with an **Actor** column and expands the details modal with source, API key, and user email.
> 
> A large **migration** drops `audit_logs` foreign keys to `users` and `orgs` so retained rows keep attribution after deletes, adds **`org_id_tombstones`** and triggers to block org UUID reuse, and rewrites **`audit_log_trigger`** to classify JWT users vs write-capable API keys vs system, suppress noisy stats/counter updates, and skip orphaned webhooks during org deletion. **Webhook payloads** and the **organization audit API** expose the new actor fields; generated **Supabase types** are updated.
> 
> **Tests and CI**: integration tests assert actor attribution and tombstone behavior; `test:cli` and Vitest **hook timeouts** are raised for CI stability; workflow checkout uses **full history** for path filters on long-lived branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d968f8bbb39f5740d34aa22353408abe69804a19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


riderx:
@coderabbitai review

RobinWitch (changes requested):
`audit_log_trigger()` now resolves the actor by looking at `get_apikey_header()` before checking `auth.uid()`. That changes live JWT requests whenever the client also sends `capgkey`: the existing `31_test_get_apikey_header.sql` case documents that `get_apikey_header()` returns the `capgkey` even when `Authorization: Bearer ...` is present, so lines 215-228 classify the mutation as `actor_type = 'apikey'` and skip the user branch at lines 231-236.

That means a normal Supabase/PostgREST write that succeeds under the JWT/RLS identity on `orgs`, `apps`, `channels`, `app_versions`, or `org_users` can be audited as if it was performed by whichever unexpired API key happened to be in the headers. In particular, a read-only or otherwise unrelated key can appear to have performed the write even though authorization came from the JWT user. The old `get_identity('{read,upload,write,all}')` path returned `auth.uid()` first, so actor attribution should preserve that precedence or otherwise only treat the key as the actor when it is the credential that authorized the operation.

riderx:
Addressed the outstanding review feedback in 6eefb7ba9: legacy actor rows are now marked unknown, authenticated users take precedence over a co-present API key, actor emails are limited to user actors, and the account-deletion function is service-role-only with a privilege regression test.

riderx:
Addressed the org-delete webhook finding in f3ab6ea6e. The audit row is retained, but the trigger now intentionally skips enqueueing the undeliverable org DELETE event because that transaction has already cascaded the organization webhooks away.

riderx:
Addressed the background-write finding in cf1cd8447. The audit trigger now preserves the prior volume contract by returning before audit insertion for unattributed system work; user and valid API-key writes remain attributed. Added a database regression assertion for this path, and raised CI integration hook timeout to 60s for shared-database setup contention.

riderx:
Addressed the remaining audit feedback in b24bd8549: meaningful system events are retained, while only high-frequency app counter updates are excluded; API-key attribution now requires a write-capable RBAC permission; and the migration timestamp is newer than main. Added pgTAP regressions for the counter and read-only-key paths.

riderx:
Final CI follow-up in a53266e21: added the missing accessible label required by the Sonar reliability gate, and changed the push checkout to full history so paths-filter no longer mutates shallow history while finding the merge base. The prior full source test matrix passed; I’m waiting for this head's fresh checks and review state.

riderx:
Rebased onto current main and pushed f2a65d94d. The migration now avoids full audit-history scans/rewrites and blocking unused indexes; org-deletion paths suppress orphaned webhook queue work for both direct cascades and account cleanup; the tombstone-sensitive log-as fixture now uses a fresh UUID per retry; and the current-main device country-code type drift is restored in both generated type files. Local lint and full typecheck pass; the focused audit/log-as integration suite passed before the local Docker daemon restarted. I’m running the fresh GitHub CI and review loop now.

riderx:
Rebased again onto the current main (5ebf11b) and restamped the still-unmerged audit migration to 20260710192458; the migration-order guard passes locally. This supersedes the previous CI cycle, and I’m waiting on the fresh merge-target checks.

riderx:
Fixed the Cloudflare shard failure in babadfe99: the stale-org retry fixture now makes its unique (name, created_by) key dynamic as well as its UUID and email. Local typecheck passes; restarting the full CI/review loop.

riderx:
@codspeedbot explain the regression on /updates manifest response with metadata

PR #2481, opened 2026-06-10
chore(observability): trace version size queue processing

riderx:
## Summary
- Add per-message queue logs for app version update dispatch/results, including msg id, CF id, status, target URL, and sanitized version fields.
- Add structured on_version_update logs for received payloads, branch selection, R2 size resolution, and app_versions_meta upsert results.
- Add R2 existence-check logs when HEAD returns false or throws, plus unit coverage for the queue payload trace summary.

## Motivation
Recent production investigation proved fresh app_versions rows can remain with app_versions_meta.size = 0 even when the R2 object exists with a real size. The exact final failure path is not identifiable from current evidence because queue_consumer deletes successful pgmq messages and only logs aggregate success counts. This PR does not change behavior; it adds the missing proof points so the next occurrence shows whether the final r2 update message was dispatched, skipped, failed storage existence, failed size lookup, or failed DB upsert.

## Business Impact
Accurate bundle size metadata controls storage reporting and billing confidence. These logs reduce time-to-root-cause for production size drift without changing upload/download behavior.

## Test Plan
- [x] bun test tests/queue-consumer-message-shape.unit.test.ts
- [x] bun lint:backend
- [x] bun run typecheck:backend
- [x] commit hook ran bun run cli:typecheck && bun run typecheck:backend && bun run typecheck:frontend

AI-generated-by: Codex

### DIFF
--- a/supabase/functions/_backend/triggers/on_version_update.ts
+++ b/supabase/functions/_backend/triggers/on_version_update.ts
@@ -37,20 +37,65 @@ async function resolveOwnerOrg(c: Context, record: Database['public']['Tables'][
   return data?.owner_org ?? null
 }
 
+function getManifestEntryCount(value: unknown): number {
+  if (Array.isArray(value))
+    return value.length
+  return value ? 1 : 0
+}
+
+function versionUpdateLogFields(
+  record: Database['public']['Tables']['app_versions']['Row'],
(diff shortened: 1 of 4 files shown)


riderx:
@coderabbitai review

PR #2485, opened 2026-06-11
fix(api): avoid false zero app version sizes

riderx:
## Summary
- Stop `on_version_update` from writing `app_versions_meta.size = 0` while an upload is still in `r2-direct`.
- Stop finalized `storage_provider = 'r2'` rows from falling through to zero metadata when `getPath()` cannot read the R2 object path.
- Keep real size writes limited to the branch where `storage_provider = 'r2'` and R2 returns a readable object path, with logs for every branch.
- Add focused unit coverage for `r2-direct`, finalized `r2` with no readable object path, and finalized `r2` real-size writes.

## Motivation
I checked Cloudflare Workers Observability logs in account `9ee3d7479a3c359681e3fab2c8cb22c0` for the last 7 days. The exact production failure is **not identifiable from the retained Cloudflare logs**:

- `getSize returned 0 after retries`: 0 events
- `manifest_size_not_found`: 0 events
- `getSize failed all manifest storage candidates`: 0 events
- `getSize head failed`: 0 events
- `getSize final`: 0 events
- `manifest file_size updated`: 0 events
- `function_name = on_manifest_create`: 0 events across services
- `function_name = on_version_update`: not observable from the indexed fields

So this PR does **not** claim the non-TUS upload-link path is the production root cause. Prod uses TUS, and the previous wording was too strong.

The exact code issue fixed here is still real: before this change, `on_version_update` treated a missing readable R2 path as a reason to fall through into `upsertZeroVersionMetadata()`. That can happen for both `r2-direct` and already-finalized `r2` rows because `getPath()` returns `null` when the object cannot be verified. Missing object metadata is not proof that the bundle size is zero, so we must not persist zero in that path. The handler now logs the branch and skips zero writes until a real R2 size is readable.

## Business Impact
This prevents false zero-byte app version metadata from being persisted when R2 metadata is unavailable. That keeps storage accounting closer to object storage truth and makes the next occurrence diagnosable from explicit branch logs instead of silently writing zero.

## Test Plan
- [x] `bunx vitest run tests/on-version-update-cleanup.unit.test.ts`
- [x] `bun lint:backend`
- [x] Commit hook ran `bun run cli:typecheck && bun run typecheck:backend && bun run typecheck:frontend`
- [x] GitHub CI passed on the amended PR

AI-generated-by: Codex

### DIFF
--- a/supabase/functions/_backend/triggers/on_version_update.ts
+++ b/supabase/functions/_backend/triggers/on_version_update.ts
@@ -199,12 +199,49 @@ async function handleManifest(c: Context, record: Database['public']['Tables']['
     cloudlog({ requestId: c.get('requestId'), message: 'error delete manifest in app_versions', error: deleteError })
 }
 
+function getVersionMetadataBranch(v2Path: string | null, storageProvider: string): 'r2_bundle_size' | 'r2_missing_path' | 'r2_direct_waiting_for_finalize' | 'zero_metadata' {
+  if (storageProvider === 'r2')
+    return v2Path ? 'r2_bundle_size' : 'r2_missing_path'
+  if (storageProvider === 'r2-direct')
+    return 'r2_direct_waiting_for_finalize'
+  return 'zero_metadata'
+}
+
+async function upsertZeroVersionMetadata(c: Context, record: Database['public']['Tables']['app_versions']['Row'], v2Path: string | null): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


riderx:
@coderabbitai review

riderx:
Closing this PR because the production root cause was not proven. Replacing it with an observability-only PR that records the exact storage lookup failure reason before making another fix.

PR #2492, opened 2026-06-12
feat(mcp): full MCP-conducted Builder onboarding — iOS + android + tail, gate flipped (PR 2)

WcaleNieWolny:
## What

PR 2 of the MCP/TUI onboarding integration (stacked on #PR1-branch `wolny/mcp-tui-integration`): the MCP now conducts the COMPLETE Builder onboarding for both platforms — every journey the TUI can do is completable through the three MCP tools — and `__CAPGO_MCP_ONBOARDING__` is flipped ON, shipping the tools in the release bundle (+141 KB, +4.8%).

## How it's built

The MCP is a thin driver over the SAME shared headless engines the TUI uses (ios/flow.ts, android/flow.ts, tail/flow.ts) — no duplicated flow logic:

- **iOS create-new, granular** — decideIos mirrors decideAndroid: total resume + auto effects + structured views; verify-app gate is multi-turn (verifyAction/verifyAppId); cert-limit revoke picker, duplicate-profile recovery, structured error screen (retry/restart/exit/email-support + corrected-key re-collect).
- **iOS import-existing** — capability-gated setup fork (macOS + scan deps), identity/profile pickers, no-match recovery + Developer-Portal walkthrough, manual .mobileprovision path input, Keychain export inside one call with the user pre-warned about the macOS dialog.
- **Post-save tail, both platforms** — CI secrets (gh + glab, overwrite gates, not-ready retry with re-detection), GitHub Actions workflow write behind an explicit preview consent gate ('view' returns the YAML in the result), .env export (0600, overwrite gates), package-manager/build-script pickers with lockfile recommendation.
- **Cross-call state** — mcp/session-state.ts (process-local carried registry; restart degrades to engine self-heal, pinned by tests) + mcp/tail-progress.ts (slim whitelist-only disk persistence: markers + non-secret prefs; the resume routers do the rest).
- **Both data-safety gates** (credentials-exist, backup/cancel) work on iOS too, including dual-platform projects.
- **Secrets discipline**: carried secrets (p8/p12/passwords/CAPGO_TOKEN) never serialize into tool results or progress files — enforced by a redaction helper asserted on every test walk + verify-tree no-leak markers.

## Found by the live AI e2e (fixed here)

1. iOS credentials-exist gate bounced to platform-select on dual-platform projects.
2. A server restart at the unanswered GitHub-Actions consent gate silently dropped the user's with-workflow choice (resume skipped the gate).
3. The terminal build-complete result carried no evidence for the upload it claimed — the Codex judge refused to take the agent's word for it. It now surfaces the upload summary / workflow path / .env path, like the TUI screen.

## Tested

- **Private suites** (cli-mcp-tests, overlay): test-mcp-onboarding 87 · ios-flow 23 · tail 28 · ios-import 13 · explain-coverage 5 (authored explanation for all 63 interactive states, 90 reachable inventoried) · schema-gate 6 · session-state 14 · android-flow 111. Engine↔schema parity 43/43 keys; strict one-answer-per-call gates everywhere.
- **Hermetic e2e tree** (no LLM): 20/20 paths, 429 assertions, on-disk artifact + no-leak checks.
- **LIVE Codex runs** (gpt-5.5 actor + judge, isolated CODEX_HOME): **20/20 paths PASS, zero flakes** — full android (keystore generate/import incl. JKS edges, SA-import + 3-way validation recovery, OAuth fire-and-poll incl. token-reject, both GCP arms, credentials gates, full build + tail), full iOS (create-new happy through the tail, verify-app wrong-build-id + autofix, cert-limit revoke, both gate arms).
- **capgo chain**: full 97-segment `bun run test` green on the flipped bundle; tsc clean; dev-gate test now POSITIVELY asserts the 3 tools ship while dev markers stay forbidden.
- **TUI untouched**: the frozen TUI E2E suite runs 79/79 journeys green against this branch's dist; zero e2e-tui files changed (requirement).

## Notes for review

- The MCP build phase keeps the C2/D2 contract (Terminal-launch/agent hand-off + BuildOutputRecord polling); the tail enters after a confirmed build. requesting-build never runs in-process.
- Deliberate exclusions from 'every journey' (documented): ai-analysis-* (the conducting agent reads logs itself — no AI-calling-AI), self-update re-exec, native file-picker dialogs (manual path inputs instead), the TUI-only resume-prompt fork (MCP resumes silently).
- The private submodule pin stays at the PR-1 commit (CI's TUI preview workflow is unaffected); the MCP suites + 20-path tree live at cli-mcp-tests main `c3e682b` and run locally via run-tests.sh.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Enabled MCP onboarding support for cloud build workflows
  * Added cloud build management with status monitoring, log streaming, and cancellation capabilities
  * Enhanced iOS and Android onboarding flows with improved error recovery and resume handling
  * Strengthened build credential security with stricter file permissions and validation

* **Bug Fixes**
  * Improved OAuth session state handling for more reliable sign-in flows
  * Enhanced platform selection and progress tracking during onboarding

* **Tests**
  * Added comprehensive test coverage for onboarding flows, OAuth operations, and platform selection


wsyyj-050320 (commented):
I found one security/privacy issue worth fixing before merge.

`mapAndroidView()` surfaces an auto-generated Android keystore password in the MCP tool result when `service-account-method-select` is rendered:

- `summary` appends `(password: ${opts.keystorePassword})`
- `context` includes `{ keystorePassword: opts.keystorePassword }`

The rationale in the comment makes sense for an interactive TUI: the user has not otherwise seen the generated password. But this is an MCP result, so it is likely to be captured in agent transcripts, tool logs, CI/debug artifacts, or downstream orchestrator memory. That seems to violate the PR's own invariant that carried secrets must never serialize into MCP tool results. It also creates an inconsistent rule: manual passwords are not echoed, but generated passwords are still full secrets for release signing.

Suggested fix: keep writing the generated password to the local credentials/backup path, but do not include the raw password in `summary` or `context`. Instead, surface a non-secret instruction such as "the generated keystore password was saved in the local credentials file; back up that file now". If the UX truly requires one-time display, consider a dedicated, explicitly sensitive output channel rather than the normal MCP result payload.

cli/src/build/onboarding/mcp/onboarding-tools.ts
```diff
@@ -213,97 +466,40 @@ function buildDeps(sdk: CapgoSDK): EngineDeps {
...
       // Defense-in-depth: the resolved path must stay inside androidAppDir.
       const resolvedDir = resolve(androidAppDir)
       const resolvedFile = resolve(filePath)
       if (resolvedFile !== resolvedDir && !resolvedFile.startsWith(resolvedDir + sep))
-        throw new Error('Refusing to write keystore outside the android/app directory.')
+        throw new Error('Refusing to write keystore outside the android app directory.')
       const bytes = Buffer.from(base64, 'base64')
-      await writeFile(filePath, bytes)
+      // Owner-only: the .p12 holds the signing private key (hostile-review 2026-06-12).
+      await writeFile(filePath, bytes, { mode: 0o600 })
```
jinhongliang991013, line 499: **[P1] Replace or harden an existing keystore path before writing**

The `mode: 0o600` option only controls creation; it does not chmod a file that already exists, and `writeFile` follows an existing symlink at this predictable `<alias>.p12` path. I reproduced the permission case by pre-creating `release.p12` as `0666` and then running the same `writeFile(..., { mode: 0o600 })`: the resulting private-key file was still `0666`. A repository can also contain or pre-plant a symlink at this location, redirecting the signing-key write outside `androidAppDir` despite the lexical `resolve` check. Please `lstat` and reject links, then replace the destination with an owner-only regular file, for example by writing a `0600` temp file and atomically renaming it, followed by `chmod(0600)` on the final file.

jinhongliang991013 (commented):
I found one additional signing-key handling issue in the Android MCP path.

PR #2493, opened 2026-06-12
feat(cli): in-repo macOS ASC key helper + iOS onboarding p8 fork + stdout stats protocol

WcaleNieWolny:
## What & why

Two things:

1. **Moves the native macOS Swift helper into this repo** at `cli/native/asc-key-helper/` (a SwiftPM package, product `P8Extract`) so it builds and ships **with** the CLI instead of living in a separate repo.
2. **Wires it into the `build init` iOS onboarding** as the `.p8` entry fork, and forwards the helper's usage statistics to PostHog over a stdout protocol.

The helper is a guided macOS app: it walks the user through creating an App Store Connect **team** API key in an embedded browser, auto-captures the **Issuer ID + Key ID**, intercepts the **one-time `.p8`**, validates it against Apple, and saves it to `~/.appstoreconnect/private_keys/` — no copy-pasting credentials.

## The onboarding UX (the new path)

At the App Store Connect API-key step of `build init` (iOS), the wizard asks:

```
Do you already have an App Store Connect API key (.p8 file)?
  ✨ No — create one for me (guided, opens a window)   → spawns the signed helper
  ✓  Yes — I have a .p8 file                            → existing import flows (file picker / type path), unchanged
```

Picking **"create one for me"** launches the helper directly — the helper's own in-app intro/consent screen explains the safety model and offers a manual fallback, so there's no extra TUI sub-question. The automated path (`asc-key-generating`) streams the helper's stats to PostHog, captures `keyId`/`issuerId`/`.p8`, shows the full-screen **`asc-key-created`** success gate (press Enter), then **converges on the existing `verifying-key` → save flow** — the rest of onboarding (cert/profile/CI secrets/build) is untouched.

**The fork is only shown when guided creation can actually run** (see *Trust gating* below). When it can't — non-macOS, macOS < 14, the helper isn't installed, or its signature/team doesn't verify — the wizard skips the fork entirely and goes **straight to the manual `.p8` instructions**, exactly as before the helper existed (and as it always does on Linux). A user is never offered "create one for me" only to be rejected when the helper can't launch.

New steps: `p8-source-select`, `asc-key-generating`, `asc-key-created`. Also adds a standalone entry: `build credentials apple-key` (alias `asc-key`).

## Distribution & trust

**One signed npm package ships both macOS helpers.** The keychain-export helper and the ASC key helper are unified into `@capgo/cli-helper-darwin-{arm64,x64}` (now **`1.1.0`**, published to `latest`), each tarball containing two Developer-ID-signed + **notarized + stapled** app bundles:

| App | Role |
| --- | --- |
| `CapgoAscKeyHelper.app` | the guided App Store Connect API-key helper (this PR) |
| `CapgoKeychainHelper.app` | the existing keychain-export helper |

These are listed in `@capgo/cli`'s **`optionalDependencies` (`^1.1.0`)**, so the arch-matching package **auto-installs on macOS** with a plain `npm i @capgo/cli` and is **skipped by `os`/`cpu`** on Linux/Windows and the non-matching Mac arch — no Linux/Windows install bloat. One merged CI workflow (`publish_cli_helper.yml`) builds, signs, notarizes, staples, and publishes both apps in a single package per arch.

### Trust gating (`probeGuidedHelper`)

Before the wizard renders, the CLI probes — once — whether guided creation can actually run: **macOS + the signed helper installed + its Developer-ID signature and Capgo Team ID verified** (`codesign --verify --strict -R '… anchor apple generic … certificate leaf[subject.OU] = "UVTJ336J2D" … identifier "app.capgo.asc-key-helper"'`). The single verified boolean gates every "create one for me" branch. A helper that is **wrongly signed, wrong-team, tampered, or absent** is treated exactly like not-installed — the guided option is withheld. The signed-package bundle is also **re-verified just before spawn** (verify-then-spawn), so an untrusted bundle is never launched. The dev-only env overrides (`CAPGO_ASC_KEY_HELPER_PATH`, etc.) are dead-code-eliminated from release builds (CI asserts their absence).

## The stdout stats protocol (helper ⇄ CLI ⇄ PostHog)

Newline-delimited JSON on **stdout**, every line tagged `capgoAscKey:1` so chatter is ignored. `event` lines are forwarded to PostHog (channel `app-store-connect-key`); the terminal `result` line delivers the credentials. Diagnostics stay on **stderr**. Full spec: `cli/src/build/onboarding/asc-key/PROTOCOL.md`.

```jsonc
{"capgoAscKey":1,"kind":"event","ts":12,"runId":"…","name":"step_changed","props":{"from":"login","to":"verifyAccess","elapsed_ms_on_prev":340}}
{"capgoAscKey":1,"kind":"result","ts":900,"runId":"…","ok":true,"keyId":"…","issuerId":"…","privateKey":"…"}
```

### Security
- The **private key appears only in the `result` line** and is **never** forwarded to analytics.
- Defence-in-depth: the CLI strips any prop key matching `private_key|secret|p8|pem|password|token` before building PostHog tags (tested).
- Honours `CAPGO_DISABLE_TELEMETRY` / `CAPGO_DISABLE_POSTHOG`.
- An `ASC Key: Helper Untrusted` telemetry event + a support-log breadcrumb fire when an installed helper fails verification (a tampered / wrong-build / wrong-team signal).

## End-to-end test coverage

- **CLI-side harness** (`test:asc-key`, macOS): drives the real `runAscKeyHelper` against a fake protocol-speaking binary (success / cancel / crash / spawn-fail / secret-guard / analytics-off / abort), plus `probeGuidedHelper` cases for every reason — **not-installed**, **wrong-signature** (injected runner *and* a real ad-hoc-`codesign` bundle), and **wrong-team** (the designated requirement pins the team OU). The keychain signature test installs the real published package and verifies the genuine Developer-ID bundle vs ad-hoc / tampered.
- **TUI journeys** (`test:e2e-tui`, macOS), both user-visible directions:
  - **Available** → `guided helper available → create-for-me → key created`: a usable helper (spoofed at the resolver cache path) is offered, the wizard runs it and shows the `asc-key-created` success screen (golden-locked, incl. the "✨ create one for me" fork).
  - **Unavailable** → `guided helper unavailable → manual .p8, guided never offered`: the fork is skipped and the manual path is taken.

## Layout

| Path | Role |
| --- | --- |
| `cli/native/asc-key-helper/` | **The Swift app, in-repo** (SwiftPM package + README + license) |
| `cli/src/build/onboarding/asc-key/protocol.ts` | Typed contract + tolerant streaming parser + PostHog mapping + secret guard |
| `cli/src/build/onboarding/asc-key/helper.ts` | Resolve the signed helper, verify signature, spawn, forward events → `trackEvent`, return credentials; `probeGuidedHelper` |
| `cli/src/build/onboarding/macos-signing.ts` | Shared Developer-ID verifier + unified package name + designated requirement |
| `cli/src/build/onboarding/command.ts` | Pre-render probe + `guidedHelperUsable` threading |
| `cli/src/build/onboarding/ui/app.tsx`, `shell.tsx`, `steps/ios-credentials.tsx`, `types.ts` | Wizard fork + new steps + async helper invocation |
| `.github/workflows/publish_cli_helper.yml` | Build + sign + notarize + staple + publish both apps in one package per arch |

## Distribution note

The macOS-only apps are **not** bundled in the npm tarball of `@capgo/cli` itself — they ship as the separate per-arch `@capgo/cli-helper-darwin-*` optional dependency, so Linux/Windows installs stay lean.


WcaleNieWolny:
### Added e2e/journey tests for the helper integration

We can't drive the Swift GUI, but the whole CLI side of the journey is now tested against a **fake helper binary** that speaks the stdout stats protocol — `cli/test/test-asc-key-helper.mjs` (wired into the aggregate `test` script):

- **success** — events streamed → forwarded to PostHog (`app-store-connect-key`, humanized names, `prop_*`/`helper_run_id` tags), credentials returned to the caller, and the **private key never appears in any forwarded request body**.
- **cancel** — `USER_CANCELLED` result line → failure outcome.
- **crash** — no result line + nonzero exit → `NO_RESULT` with stderr captured.
- **spawn-failed** — missing binary → `SPAWN_FAILED` (no throw).
- **secret guard** — a stray `privateKey` in event props is stripped before analytics; non-secret props still forwarded.
- **analytics-off** — `forwardToAnalytics:false` returns credentials + fires `onEvent` but sends nothing.
- **macOS gate** — on a non-darwin runner it asserts `NotMacOSError` (so the suite is meaningful cross-platform; the spawn journeys run on macOS).

Plus `test-asc-key-protocol.mjs` (parser/mapping/secret-strip unit tests). Both run green locally.

The ink wizard routing (the `p8-source-select` → `asc-key-generating` fork) is covered structurally by `tsc` here; the visual frame-fit harness for the 3 new step components runs in CI (needs a full `bun install` for `string-width`, which isn't present in my sandbox).

WcaleNieWolny:
### Tests live in the private suite

Per our split (hermetic **unit** tests in capgo, integration **journeys** in the private suite — same as keychain signing), the e2e/journey tests for this helper are in **Cap-go/cli-mcp-tests#5**:
- `test:asc-key` — spawn the helper (fake binary) → stream the stdout stats protocol → forward to PostHog → return/route credentials; covers success/cancel/crash/spawn-fail/secret-guard/analytics-off.
- `walk.mjs` traverses the new `p8-source-select` fork; `drift.mjs` registers the 3 new step ids.

What stays here: `cli/test/test-asc-key-protocol.mjs` (the hermetic pure-parser unit test, analogous to `test-macos-signing.mjs`). The submodule pointer should bump to that branch when this lands.

cli/native/asc-key-helper/Sources/Models/KeyCredentials.swift
```diff
@@ -0,0 +1,54 @@
...
+    private static func savePrivateKeyCopy(_ credentials: KeyCredentials) {
+        let directory = FileManager.default.homeDirectoryForCurrentUser
+            .appendingPathComponent(".appstoreconnect/private_keys", isDirectory: true)
+        let file = directory.appendingPathComponent("AuthKey_\(credentials.keyId).p8")
+        do {
+            try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+            guard !FileManager.default.fileExists(atPath: file.path) else { return }
+            try Data(credentials.privateKey.utf8).write(to: file)
+            try FileManager.default.setAttributes([.posixPermissions: 0o600], ofItemAtPath: file.path)
+        } catch {
```
jinhongliang991013, line 94: The helper still reports success when the only durable key copy was not saved. `savePrivateKeyCopy` catches directory/write/chmod failures and returns normally, so `emit` sends `ok: true`; the standalone command then prints `Saved .p8: ...` (and its JSON deliberately omits the private key) even when that path was never created. Because Apple does not allow this key to be downloaded again, running without `--appId` can lose the only usable copy as soon as the process exits. Please make the save operation throw/return a verified status and only emit success after the exact key has been persisted, or propagate a failure that the CLI can handle without claiming the path is saved.

jinhongliang991013 (commented):
Found one data-loss path in the new key helper.

wsyyj-050320 (commented):
I found one security/privacy issue worth fixing before merge.

The new ASC helper delivers the captured `.p8` contents on the stdout NDJSON `result` line:

- `cli/native/asc-key-helper/Sources/Models/StatsProtocol.swift` documents and emits `privateKey: credentials.privateKey` in `StatsProtocol.result()`.
- `cli/src/build/onboarding/asc-key/protocol.ts` accepts `privateKey` from any success `result` line.
- `cli/test/test-asc-key-protocol.mjs` makes the raw private key in stdout part of the expected protocol behavior.

Even if the current CLI does not forward that result line to PostHog, stdout is still a broad capture surface: terminal transcripts, parent process logs, CI/e2e harnesses, process supervisors, debug artifacts, and future wrappers commonly record it. The PR also names this stream a stats/telemetry protocol, so future maintainers may reasonably treat protocol lines as safe to persist. A one-time App Store Connect `.p8` key is long-lived signing material; putting it on stdout greatly increases the chance of accidental disclosure compared with the existing local key-file path.

Suggested fix: keep the helper writing `~/.appstoreconnect/private_keys/AuthKey_<keyId>.p8` with restrictive permissions, but have stdout carry only non-secret fields such as `keyId`, `issuerId`, and `p8Path`/an opaque handle. The CLI can then read that file path when it needs to update saved credentials, without making the raw key part of the protocol stream. Please also update `PROTOCOL.md` and `test-asc-key-protocol.mjs` so tests fail if a success result contains raw `privateKey` material.

PR #2498, opened 2026-06-13
refactor: send deploy emails once per bundle

Pitiakova:
## Summary (AI generated)

- Updated deploy history notifications to use the one-time org notification helper.
- Added the deployed version id to the notification payload and dedupe key so each bundle deploy email is sent once per version.

## Motivation (AI generated)

Deploy history inserts can trigger duplicate bundle deployed emails for the same version. This keeps deploy email behavior consistent and avoids repeated notifications to organization members.

Customers receive fewer duplicate deploy emails, reducing notification noise while preserving deployment visibility.

## Test Plan (AI generated)

- [ ] Not run; change is limited to notification helper wiring for deploy history creation.

Generated with AI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the deployment notification system to prevent duplicate “bundle deployed” notifications to organization members.
  * Notifications now consistently include the deployed version information for clearer context, while keeping delivery behavior idempotent.


RobinWitch (changes requested):
I found one more deployment-notification edge case that looks separate from the existing background-task/client-lifetime finding.

The one-time notification key only includes `record.version_id`, but `deploy_history` is channel-scoped. `record_deployment_history()` inserts a row each time a channel changes version, and this handler includes `channel_id` in the event payload. If the same bundle/version is first deployed to one public channel and later promoted to another public channel in the same org, the first run claims `(owner_org, 'bundle:deployed', 'bundle:deployed:<version_id>')` in `notifications`; the second channel's deploy then hits the same claim and `sendNotifToOrgMembersOnce()` returns before sending. That suppresses a legitimate production-channel deploy email just because the bundle was already deployed to another channel.

I think the idempotency key needs the channel as well, for example `bundle:deployed:${record.version_id}:${record.channel_id}`, or a deploy-history row identifier if the intended unit is one deploy history event. `version_id` can still stay in the payload for template context.

Evidence I checked:
- `supabase/functions/_backend/triggers/on_deploy_history_create.ts:73-79` passes `channel_id` in event data but dedupes only on `version_id`.
- `supabase/functions/_backend/utils/notifications.ts:100-104` and `:145-147` use `(owner_org, event, uniq_id)` as the notification lookup/conflict target.
- `supabase/migrations/20260519151250_remove_builtin_unknown_app_versions.sql:52-68` records deployments on each channel version change, so two channels can produce two deploy_history inserts for the same `version_id`.

PR #2507, opened 2026-06-16
feat(dashboard): surface guided P8 key creation in builder promo (macOS-gated)

WcaleNieWolny:
## What & why

The Capgo Builder promo modal on the dashboard assumed users already had an App Store Connect API key (`.p8`). It never told them about the recently-merged **guided key-creation** flow — where, on macOS, `capgo build init` opens a window, walks you through App Store Connect, and captures the key for you.

This adds that beat to **Slide 2 ("Set up once, in minutes")** — but only for macOS visitors, since the guided helper is a macOS-only desktop app. Non-macOS visitors keep the existing generic "guided signing" copy, so we never imply a capability that isn't there for them.

## Changes

- **`BuilderPresentationModal.vue`** — `isMacOS` user-agent check (excludes iPadOS, which reports a "Macintosh" UA but exposes touch points). Slide 2's lead + 4th checklist item route through `computed` wrappers that pick the macOS or generic string.
- **`messages/en.json`** — two new macOS-only keys; originals left untouched:
  - `builder-promo-s2-lead-macos`
  - `builder-promo-s2-item4-macos`

## Test plan

- [x] `oxlint` + `eslint` on the changed component — clean
- [x] App boots against prod (`bun run dev`, Branch main) with 0 console errors
- [ ] On macOS: open an app with no native builds → **Learn more** → Slide 2 shows the guided-key-creation copy
- [ ] On non-macOS (or iPad): Slide 2 shows the original generic copy

## Follow-up (separate PR, `website` repo)

Marketing page (`native-build.astro`) + docs (`ios.mdx`, `build.mdx`) get the same "no key? we create it for you, guided" angle, with a macOS-only footnote on the marketing page.

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -492,6 +492,8 @@
   "build-step-wait-subtitle": "This page will self-refresh when the build is detected",
   "build-time": "Build Time",
   "build-timeout-help": "Native cloud builds are stopped after this many minutes. The default is 15 minutes.",
+  "builder-promo-s2-lead-macos": "A guided setup configures signing for iOS & Android — and if you don't have an App Store Connect key yet, it opens a guided window and creates one for you, step by step. The rest — build machines, Xcode, Fastlane, CI — is already running, nothing to manage.",
+  "builder-promo-s2-item4-macos": "No Apple key yet? A guided window creates it for you",
   "build-timeout-invalid": "Build timeout must be between 5 and 360 minutes",
   "build-timeout-label": "Build timeout (minutes)",
   "build-time-usage": "Build time usage: ",
--- a/src/components/dashboard/BuilderPresentationModal.vue
+++ b/src/components/dashboard/BuilderPresentationModal.vue
@@ -26,6 +26,15 @@ const router = useRouter()
 const config = getLocalConfig()
 const reduce = typeof matchMedia !== 'undefined' && matchMedia('(prefers-reduced-motion: reduce)').matches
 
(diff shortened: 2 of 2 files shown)


Monster5860 (commented):
I reviewed the two-file change and did not find a blocking runtime issue in the new macOS-gated copy path:

- `computed` is already imported from Vue, and the template refs (`s2Lead`, `s2Item4`) will unwrap correctly.
- The `navigator` access is guarded for non-browser/SSR contexts.
- The iPadOS exclusion with `maxTouchPoints` matches the intent described in the PR.

A few items I would address or at least record before merge:

1. The PR is currently blocked by SonarCloud's maintainability gate. I checked the public Sonar issue API for this PR and it reports 5 code smells in `BuilderPresentationModal.vue`: three `window` vs `globalThis` findings, the existing `role="dialog"` accessibility finding, and one CTA contrast finding. These appear to be older issues surfaced because this PR touches the file, but the quality gate is still failing, so the author/maintainer should either clean them up here or explicitly decide to handle/suppress them separately.
2. The test plan leaves both platform-specific rendering checks unchecked. Since the only behavior change is UA-dependent copy selection, I would want at least one recorded manual check/screenshot for macOS and one for non-macOS/iPad before merging.
3. If this pattern grows beyond these two strings, consider extracting the platform detection/copy selection into a small helper or computed object. It is fine as-is for this patch, but duplicating the `isMacOS ? t(...) : t(...)` pattern on future slides would make the modal harder to maintain.

PR #2509, opened 2026-06-16
feat(supabase): hourly auto-complete of finished onboarding apps

WcaleNieWolny:
## Problem

Apps created via the dashboard onboarding flow are flagged `apps.need_onboarding = true`, and the flag is only ever cleared by `capgo init` or an explicit dashboard PUT. Users who set their app up any **other** way — normal `bundle upload` — stay flagged forever and keep seeing the **"This app is still marked as pending onboarding. Demo data is temporary…"** banner on a real, in-use app.

### Production impact (Capgo-EU)
- **~478 apps** flagged `need_onboarding = true`; **187** are real, actively-used apps (avg ~33 bundle versions), ~90 of them in paying orgs.
- This cron auto-completes **137** of them on the first run (real bundle + created >15d ago + no seeded demo data).

## What this does

- **`app_has_real_bundle(text)`** — selection helper; true when an app has a real bundle (excludes `builtin`/`unknown` placeholders and seeded demo bundles under `demo/<app_id>/…`).
- **`cleanup_completed_onboarding_apps()`** — flips `need_onboarding → false` for apps with `>=1 real bundle AND created_at < now()-15d AND NOT has_seeded_demo_data`, **per-app inside an exception block**.
- **Hourly registration** in the consolidated `cron_tasks` system (`hour_interval=1, run_at_minute=0`).

## Why it's safe (this is the important part)

Flipping `need_onboarding → false` fires the existing `cleanup_onboarding_app_data_on_complete` trigger → `clear_onboarding_app_data`. As of migration `20260519123613` that path is **provenance-based**: it calls `reset_onboarding_demo_app_data`, which deletes only rows explicitly fingerprinted as demo-seeded (tracked in `onboarding_demo_data`) and `RAISE EXCEPTION`s rather than cascade into any untracked/real row.

So:
- **This PR does NOT modify `clear_onboarding_app_data`.** It relies on the existing provenance-safe cleanup. (An earlier revision of this PR hardened a guard on that function — that was based on a stale, pre-`20260519123613` definition and would have reverted the provenance rewrite. It has been removed.)
- Of the 137 target apps, **only 1** has any demo provenance; flipping the other 136 deletes nothing. The 1 is handled safely by the provenance reset (cleans tracked demo rows or refuses).
- The **per-app exception block** means that if the provenance reset refuses for any app, that app is left pending and the batch continues — never a partial delete, never a whole-batch abort.

## Test plan

- [x] Postgres-level test `supabase/tests/60_test_cleanup_completed_onboarding_apps.sql`:
  - `app_has_real_bundle` true for a real upload, false for seeded-demo and placeholder-only apps
  - real app >15d auto-completed; real app <15d, placeholder-only, and seeded-demo apps stay pending
  - flipping a real app **preserves its bundle** (provenance cleanup has nothing tracked → deletes nothing)
  - hourly `cron_tasks` registration present
- [x] Static validation vs prod schema: referenced columns/constraints exist, `ON CONFLICT (name)` valid, predicate selects 137 apps, migration ordering valid, no redefinition of `clear_onboarding_app_data`.
- [ ] `supabase test db` + `supabase db lint` — run in CI (Docker unavailable in authoring env).

## Notes / follow-ups (out of scope)

- `cleanup_expired_demo_apps()` still does an unguarded `DELETE` of `need_onboarding` apps >14d old with seeded demo data; if an app ever had both seeded demo data and real uploads it'd be deleted wholesale (overlap is **0** in prod today). Worth a separate provenance-aware fix.

## Summary by CodeRabbit

* **New Features**
  * Onboarding apps are now automatically cleaned up on an hourly schedule after 15+ days when demo data is cleared and a real bundle exists.

* **Tests**
  * Added comprehensive test coverage for the onboarding app cleanup functionality.


supabase/migrations/20260616065214_safe_onboarding_auto_complete.sql
```diff
@@ -0,0 +1,138 @@
...
+SECURITY DEFINER
+SET search_path = ''
+AS $$
+  SELECT EXISTS (
+    SELECT 1
+    FROM public.app_versions v
+    WHERE v.app_id = p_app_id
+      AND v.deleted = false
+      AND v.name NOT IN ('builtin', 'unknown')
+      AND NOT EXISTS (
```
riderx, line 49: This helper is using the old `demo/<app_id>/...` manifest path as the demo discriminator, but current demo ownership is recorded in `onboarding_demo_data` and this query already runs behind `NOT has_seeded_demo_data(app_id)`. More importantly, any non-deleted non-placeholder `app_versions` row with no finished upload also passes here. The CLI creates the version row before it uploads and later fills `r2_path`; `--dry-upload` or a failed cleanup can leave metadata-only rows. That means the cron can clear onboarding for an app that does not actually have a usable bundle. Please reuse the existing upload-ready predicate from `complete_onboarding_after_first_upload` (`external_url` present for external bundles, or `r2_path` present for stored bundles) and, if demo exclusion is still needed, check `onboarding_demo_data` provenance rather than the manifest path.

riderx (commented):
I found one actionable issue after re-checking the demo seeding path.

supabase/migrations/20260616065214_safe_onboarding_auto_complete.sql
```diff
@@ -0,0 +1,138 @@
...
+  run_on_dow,
+  run_on_day
+) VALUES (
+  'cleanup_completed_onboarding_apps',
+  'Hourly: clear apps.need_onboarding for apps that finished real onboarding (>=1 real bundle, created >15 days ago, no seeded demo data)',
+  'function',
+  'public.cleanup_completed_onboarding_apps()',
+  null,
+  null,
+  1,
```
riderx: Running this every hour is unnecessary for a state transition gated by `created_at < now() - interval '15 days'`. An app can only become eligible once, and clearing a banner a few hours earlier does not justify scanning pending onboarding apps 24 times per day. Please make this a daily task instead, for example set `hour_interval = null`, `run_at_hour` to a fixed UTC hour, and keep `run_at_minute = 0`.

riderx (commented):
One scheduling issue after re-checking the cron cadence.

WcaleNieWolny:
Thanks for the reviews — addressed in `db95be84e`:

- **@chatgpt-codex-connector (P1, provenance regression):** Resolved. The migration no longer redefines `clear_onboarding_app_data` at all — it relies on the existing provenance-based `clear_onboarding_app_data → reset_onboarding_demo_app_data` path (the broad-delete redefinition was removed before your comment's commit; confirmed no `CREATE OR REPLACE … clear_onboarding_app_data` remains).
- **@riderx (real-bundle = upload-ready):** Fixed. `app_has_real_bundle` now requires an upload-ready bundle using the same predicate as the reverted `complete_onboarding_after_first_upload` trigger — `storage_provider='external'` with non-blank `external_url`, or `storage_provider <> 'r2-direct'` with non-blank `r2_path`. Metadata-only / `--dry-upload` rows no longer qualify. Demo exclusion now uses the `onboarding_demo_data` provenance table instead of the `demo/` manifest path. (Verified against prod: same 137 apps selected, 0 dropped — so no regression, but future metadata-only rows are now excluded.)
- **@riderx (hourly → daily):** Fixed. Now a daily task: `hour_interval=null, run_at_hour=4, run_at_minute=0` (04:00 UTC, just after `cleanup_expired_demo_apps` at 03:00).
- **@coderabbitai (non-upload-ready fixture):** Added. The test now has a `noupload.old` fixture (non-placeholder version with no `r2_path`/`external_url`) and asserts both `app_has_real_bundle = false` and that it stays pending — plus a demo-provenance fixture.

riderx (changes requested):
Please remove `.zigrep_archive` and `codedb.snapshot` from this PR before merging. They are local/generated code-index artifacts unrelated to the onboarding cleanup; `codedb.snapshot` adds ~17.9 MB of binary repo/history noise, and `.zigrep_archive` is binary tooling state too. The SQL migration and test changes look fine after the last update, but these artifacts should not ship.

PR #2521, opened 2026-06-17
feat(dashboard): show build duration in builds table

WcaleNieWolny:
## What

Replace the confusing **Wait Time** column in the builds table with a **Build Duration** column.

`Wait Time` showed `runner_wait_seconds` — the time a job sits in the queue before a runner picks it up (usually 2–9s). It told users nothing about how long their build actually took.

`Build Duration` shows the real per-job wall-clock build time from `build_logs.build_time_unit`, formatted like `2m 43s`.

## How

- The table already loads a page of `build_requests`. After that load, it fires one extra query against `build_logs` (`select build_id, build_time_unit where build_id in (...)`) and maps results by `builder_job_id`.
  - The join is `build_logs.build_id = build_requests.builder_job_id` (verified against prod: 2667/3228 requests join cleanly; `build_logs.build_id` does **not** match `build_requests.id`).
- Durations load **before** `elements` is set, so the first paint already shows the correct value (no re-render flash).
- Rows with no build log yet (pending / running / abandoned) render an em dash `—`.
- Uses `build_time_unit` (true seconds), **not** `billable_seconds` (which is doubled for iOS billing), so the column reflects actual build time.
- Renamed the local `formatWaitTime` helper to `formatDuration` (it's a generic seconds formatter; "Wait" was now misleading).
- i18n: replaced the now-unused `runner-wait-time` key with `build-duration` in `messages/en.json`.

## Why not `updated_at − created_at`?

A background `ai_analyzed` pass bumps `updated_at` days after the build finishes, so end-to-end elapsed is unreliable (prod shows "2-day builds"). `build_time_unit` is the authoritative per-job duration.

## Scope

This is **Part 1** of a larger builds-dashboard improvement. Follow-ups (separate PRs): per-app **Builds** stacked column chart (iOS/Android × succeeded/failed) and a **Build Time** line chart on the app dashboard.

## Test plan

- [x] `eslint` on the changed component — clean
- [x] `vue-tsc` frontend typecheck — passes
- [x] `messages/en.json` valid JSON
- [ ] Manual: open `/app/:appId/builds`, confirm the **Build Duration** column shows formatted durations for completed builds and `—` for pending/running
- [ ] Manual: paginate and confirm durations load per page

## Summary by CodeRabbit

* **New Features**
  * The Build table now shows per-build **duration** (build log duration) instead of runner **wait time**, with improved formatting and a “—” fallback when unavailable.
* **Bug Fixes**
  * Updated access so build duration data is available for app-scoped views.
  * Build duration fetching now reports failures with a clear error message.
  * Improved app-id resolution on the Devices Stats dashboard to follow the current route or provided context.
* **Refactor**
  * Steps Bundle polling no longer fetches/stores the latest version details.
* **Documentation**
  * Updated English UI labels and added messaging for build-duration fetch errors.


WcaleNieWolny:
## Review feedback addressed

Thanks for the reviews. Changes since the initial commit:

**Codex P2 — app-scoped users saw `—` for durations** → fixed in `supabase/migrations/20260617090000_build_logs_app_scoped_select.sql`.
The Build Duration column reads `build_logs`, whose SELECT policy only granted **org-level** read (`get_identity_org_allowed`), while `build_requests` is **app-aware** (`get_identity_org_appid`). `build_logs` gained an `app_id` column since that policy was written (FK + `idx_build_logs_app_id_created_at`), so the policy is now app-aware too. It's **additive**: the original org-level clause is kept (still covers org members and logs whose `app_id` went `NULL` when an app was deleted via `ON DELETE SET NULL`), `OR`'d with an app-aware clause mirroring `build_requests`. Read access only widens.

**Duplicate/cross-org `build_id`** → the client lookup is now scoped with `.eq('org_id', orgId)`. `build_logs` is unique only per `(build_id, org_id)`, so an unscoped `.in('build_id', …)` could surface a colliding row from another org a multi-org user can read.

**Silent failure on the durations query** → on error we now clear the map and `toast.error(t('error-fetching-build-durations'))`, matching the `build_requests` fetch, instead of console-only.

**Pre-existing oxlint `vue/no-dupe-keys`** (a local `appId` ref shadowing the `appId` prop in `StepsBundle.vue` / `DevicesStats.vue`) was failing lint on every PR; renamed the locals (`latestVersionId` / `routeAppId`) and removed the now-dead `getLatestVersionId` + write-only ref.

Re: CodeRabbit's `routeAppId` "receives values from both props.appId and the route" note — `routeAppId` only ever receives the route-derived `packageId` (or `''`); `props.appId` is read separately in the `activeAppId` computed, never assigned into `routeAppId`. So the name is accurate.

WcaleNieWolny:
@coderabbitai review

Addressed the `orgId` capture nitpick in `f431499`: `getData()` now captures `orgId` once (right after the `currentOrganization` guard) and passes it to `loadBuildDurations(builds, orgId)`, removing the post-await store read and the redundant `!orgId` guard. The earlier Codex P2 (app-scoped durations) is fixed via the app-aware `build_logs` RLS migration. All CI is green.

PR #2582, opened 2026-06-25
fix(dashboard): show support dialog when removing org owner

WcaleNieWolny:
## Problem

When you try to remove the `super_admin` who **created (owns)** an organization, the `delete_org_member_role` RPC raises `CANNOT_CHANGE_OWNER_ROLE` (the org owner is protected server-side). The dashboard surfaced this only as a transient error toast, so it wasn't clear to the user why removal failed or what to do about it.

## Change

In `src/pages/settings/organization/Members.vue`, the `CANNOT_CHANGE_OWNER_ROLE` branch of `_deleteMember` now opens a dialog instead of a toast. The dialog explains that this user created the organization and cannot be removed automatically, and offers an **Email support** button that opens a `mailto:support@capgo.app`.

- New `ownerCannotBeRemoved()` dialog helper, mirroring the existing `cannotDeleteOwner()` pattern (DialogV2 store).
- Reuses the existing DialogV2 `href` button support to render a native `mailto:` action — no new component needed.
- Three new i18n keys in `messages/en.json`: `alert-cannot-remove-owner-title`, `alert-cannot-remove-owner-body`, `email-support`.

Note: the "last super admin" path (`cannotDeleteOwner()`) is unchanged — that already shows its own dialog. This only affects the case where the owner is one of several super admins, which previously fell through to a toast.

## Test plan

- [ ] Open an org where your account is the **owner** (creator) and there is **another** super_admin present.
- [ ] Attempt to remove the owner from the Members table.
- [ ] Confirm a dialog appears ("Cannot remove the organization owner") instead of a toast, with **Cancel** and **Email support** buttons.
- [ ] Confirm **Email support** opens a mail composer to `support@capgo.app`.
- [ ] Confirm removing a non-owner member still works normally.

## Checks

- `oxlint` ✅
- `eslint` ✅
- `vue-tsc` (frontend typecheck) ✅

## Summary by CodeRabbit

* **New Features**
  * Added dedicated owner-protection dialogs for both owner-role changes and owner removal on the organization members page, with messaging tailored to “self” vs “other” cases.
  * Added new English translation strings and support email templates (including mailto subjects/bodies) surfaced in the dialogs.
* **Bug Fixes**
  * Member deletion and permission updates now route “owner cannot be removed” and “owner role cannot be changed” scenarios through the correct dialogs instead of generic toasts, including attempts to remove the organization creator.


WcaleNieWolny:
@coderabbitai The owner-first ordering in `deleteMember` is intentional, not a regression.

Rationale:
- The org owner (`orgs.created_by`) is protected server-side (`delete_org_member_role` raises `CANNOT_CHANGE_OWNER_ROLE`) and can never be removed. The owner's role also can't be changed, so the owner is **always** an `org_super_admin`.
- Therefore when `numberOfSuperAdmins === 1`, that sole super admin **is** the owner. Routing them into `cannotDeleteOwner()` (delete-org / delegate) offers a "delegate then step down" flow whose final `_deleteMember(owner)` the RPC always rejects — a dead end for the true owner.
- The product decision (confirmed with the maintainer) is that the owner must contact support to transfer ownership, which is exactly what `ownerCannotBeRemoved()` communicates (with prefilled mailto to support@capgo.app).
- `cannotDeleteOwner()` is intentionally kept for the edge case where the last super admin is **not** the owner (e.g. the owner was already removed out-of-band).

So the owner check must run first by design. Resolving as intended.

@coderabbitai resolve

PR #2619, opened 2026-07-04
feat(cloudflare): local plugin-replica worker with Postgres read subscriber

riderx:
## Summary (AI generated)

- Added `cloudflare_workers/plugin-replica/` worker that reuses the existing plugin routes
- Added a Postgres 17 Docker subscriber that auto-applies `read_replicate/schema_replicate.sql` and subscribes to local Supabase logical replication
- Added `LOCAL_READ_REPLICA_DB_URL` routing in `pg.ts` so read-only plugin queries hit the local replica
- Added `bun run plugin-replica:dev` orchestration script and local replication helpers in `read_replicate/`

## Motivation (AI generated)

Local Cloudflare plugin testing currently hits the primary Supabase database directly, while production plugin workers read from regional Postgres replicas. This change mirrors production read-replica behavior locally so plugin hot paths can be exercised against a subscriber database.

Improves local and CI fidelity for `/updates`, `/stats`, and `/channel_self` without changing production behavior. Reduces risk of shipping plugin changes that work on primary DB but fail on read replicas.

## Test Plan (AI generated)

- [ ] `bun run supabase:start && bun run supabase:db:reset`
- [ ] `bun run plugin-replica:dev` and verify worker responds on `http://127.0.0.1:8790/ok`
- [ ] Confirm read-only plugin requests use `X-Database-Source: LOCAL_READ_REPLICA`
- [ ] `bunx vitest run tests/pg-local-read-replica.unit.test.ts`

Generated with AI


RobinWitch (changes requested):
`read_replicate/common.sh` currently embeds a literal `0x01` byte in the `sed` replacement at `apply_local_subscription_source_connection`, not the `\1` capture-group reference.

That means the local replica path rewrites a valid source URL such as `postgresql://postgres:postgres@127.0.0.1:54322/postgres?...` into a URL whose password is the control character (`postgresql://postgres:<0x01>@...`). The next `psql "$SOURCE_DB_URL"` calls in `replicate_setup_source.sh` / `replicate_to_replica.sh` then authenticate with the wrong password, so `bun run readreplicate:local` and `bun run plugin-replica:dev` fail before the publication/subscription setup can run.

The extraction above uses `\1`, so this line should do the same instead of storing the literal control byte.

riderx:
Closing: superseded by colo-cache + targeted invalidation approach (new PR incoming).

PR #2622, opened 2026-07-05
fix(api): validate Analytics Engine SQL and repair invalid queries

riderx:
## Summary (AI generated)

- Fix production failures on `POST /private/devices` with `installSourceCounts: true` caused by invalid Analytics Engine SQL (`COUNT(*)`, unsupported `CASE WHEN` in `argMax`)
- Add static linting plus fixture collection for all Analytics Engine read queries, with an optional live validation script against Cloudflare
- Repair additional invalid queries found during live validation (`multiIf`, admin success/failure rate alias usage, org bandwidth JOIN)

## Motivation (AI generated)

Cloudflare Analytics Engine SQL is not Postgres and has no local emulator. Invalid syntax only surfaced in production as generic `runQueryToCFA encountered an error` alerts. We need CI coverage that catches these before deploy.

- Stops recurring API error alerts when console users load install-source device stats
- Prevents similar silent Analytics Engine query regressions across admin metrics and device analytics
- Reduces time-to-diagnose for future SQL changes with explicit lint failures instead of opaque 422s

## Test Plan (AI generated)

- [x] `bun run test:analytics-sql`
- [x] `bunx vitest run tests/cloudflare-device-pagination.unit.test.ts tests/stats-device-count.unit.test.ts`
- [x] `bun run test:analytics-sql:live` (40 queries validated against Cloudflare)
- [ ] CI backend/unit tests pass

Generated with AI

## Summary by CodeRabbit

* **New Features**
  * Added an Analytics Engine SQL linter with optional live validation to detect unsupported SQL constructs early.
  * Added new test commands for analytics SQL, including a live validation mode.
* **Bug Fixes**
  * Improved Analytics Engine query generation for install-source/device filtering and updated admin/org metrics calculations for more accurate results.
* **Tests**
  * Added a comprehensive linter test suite with fixture coverage and updated SQL expectations to match the new query generation behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,7 +179,9 @@
     "deploy:supabase_env:prod": "bunx supabase secrets set --project-ref xvwzpoazmxkqosrdewyv --env-file internal/cloudflare/.env.prod",
     "deploy:supabase_env:preprod": "bunx supabase secrets set --project-ref ibwjdnhknbkcqfbabwei --env-file internal/cloudflare/.env.preprod",
     "deploy:supabase_env:dev": "bunx supabase secrets set --project-ref aucsybvnhavogdmzwtcw --env-file internal/cloudflare/.env.alpha",
-    "size": "bunx vite-bundle-visualizer"
+    "size": "bunx vite-bundle-visualizer",
+    "test:analytics-sql": "vitest run tests/analytics-engine-sql.unit.test.ts",
+    "test:analytics-sql:live": "bun --env-file=internal/cloudflare/.env.prod scripts/validate-analytics-engine-sql.ts"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


RobinWitch (commented):
One regression still looks present in the current head: `getAdminFailureMetrics()` now computes `failure_rate` in JavaScript, but the mapped response drops the `installs` field even though the query still selects it and the previous API response included it.

That means the `failures` admin metric can no longer return the install count alongside daily failures/failure rate. The fix should be small: include `installs: row.installs || 0` in the object returned from the `rows.map(...)` block, and ideally add a fixture/unit assertion for that response shape so this does not silently regress again.

PR #2624, opened 2026-07-06
perf(cloudflare): enable Workers Cache for files worker

riderx:
## Summary
- Enable Cloudflare Workers Cache for the production files Worker, but opt the default router entrypoint out so auth/routing-sensitive paths still run normally.
- Add a `CachedFiles` entrypoint for reusable file/preview reads and forward to it with `cf.cacheKey` so file downloads strip `device_id` while preview cache keys include the preview hostname.
- Make attachment read responses default to `public, max-age=31536000, immutable, no-transform` when object metadata has no cache policy or only legacy `no-transform`.
- Keep non-production files Worker cache disabled to avoid confusing local/staging validation.
- Bump Wrangler to 4.107.0, which is required for the Workers Cache config.

## Scope note
- The existing Cloudflare snippet cache for plugin on-prem/plan-upgrade responses stays unchanged. Cloudflare Workers Cache does not store POST/PUT/PATCH/DELETE requests, while those plugin hot paths rely on synthetic snippet cache keys and non-GET traffic.
- Range requests bypass the new Workers Cache entrypoint so byte-range responses do not share the same key as full-file responses.

## References
- Cloudflare announcement: https://blog.cloudflare.com/workers-cache/
- Cloudflare config docs: https://developers.cloudflare.com/workers/cache/configuration/
- Cloudflare cache key docs: https://developers.cloudflare.com/workers/cache/cache-keys/

## Tests
- `bun lint` (passes with existing unrelated Vue newline warnings in admin dashboard files)
- `bun typecheck`
- `bunx vitest run tests/files-r2-error.test.ts tests/files-local-read-proxy.unit.test.ts tests/preview-response-headers.unit.test.ts tests/cloudflare-snippet.unit.test.ts`
- `bunx wrangler deploy --config cloudflare_workers/files/wrangler.jsonc --env=prod --dry-run --outdir /tmp/capgo-files-worker-cache --minify`
- pre-commit hook: `bun run cli:typecheck && bun run typecheck:backend && bun run typecheck:frontend`

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path file and preview delivery and CDN caching behavior; mis-keyed cache entries could serve wrong assets, though scope is limited to defined GET paths with explicit bypass rules.
> 
> **Overview**
> Turns on **Cloudflare Workers Cache** for the production files worker while keeping the default fetch handler uncached so uploads, auth, and other sensitive routes behave as before.
> 
> Eligible **GET** attachment reads (no `Range`) and most preview subdomain reads are routed through a new **`CachedFiles`** entrypoint with custom **`cf.cacheKey`** values: file paths use `/files-cache…` and drop **`device_id`** from the key; preview paths use `/preview-cache/{host}…`. Channel previews (`c\d+-…`), **`/.capgo/preview.json`**, and range requests skip Workers Cache.
> 
> Attachment read responses now default to **`public, max-age=31536000, immutable, no-transform`** when R2 metadata is empty or only legacy `no-transform`, and the in-worker **`caches.default`** lookup normalizes query params the same way (strip tracking, sort).
> 
> **Wrangler** is bumped to **4.107.0** (required for Workers Cache config); non-prod files worker keeps cache disabled. Tests cover cache headers, cache key building, and a **`cloudflare:workers`** mock for vitest.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05581bdd618771be92024a6f7a10d44c8479e84d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -204,12 +204,12 @@
         "vite-plugin-webfont-dl": "^3.12.0",
         "vitest": "^4.1.9",
         "vue-tsc": "3.3.5",
-        "wrangler": "^4.101.0",
+        "wrangler": "^4.107.0",
       },
     },
     "cli": {
       "name": "@capgo/cli",
-      "version": "8.24.5",
+      "version": "8.25.1",
       "bin": {
         "capgo": "dist/index.js",
       },
@@ -277,7 +277,7 @@
     },
     "packages/capacitor-notifications": {
       "name": "@capgo/capacitor-notifications",
-      "version": "0.0.1-private.0",
+      "version": "0.1.1",
       "devDependencies": {
         "@capacitor/android": "^8.0.0",
         "@capacitor/ios": "^8.0.0",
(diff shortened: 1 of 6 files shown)


riderx:
@coderabbitai review

RobinWitch (changes requested):
I think enabling Workers Cache on the default files entrypoint is not safe yet because the current read path relies on the Worker running for every downloadable response.

Cloudflare's Workers Cache docs say `cache.enabled` checks cache before invoking the Worker, and the cache key is the entrypoint plus path/query, not the host/method. The same docs also call out Range handling: Cloudflare strips `Range` before invoking the Worker, asks the Worker for the full body, then slices the cached response; later range requests can be served as cache hits without invoking the Worker.

That breaks two assumptions in this files Worker:

- `getHandler()` records bandwidth inside the Worker via `saveBandwidthUsage()` for both R2 reads and the existing in-worker `caches.default` hits. `createStatsBandwidth()` then writes the `BANDWIDTH_USAGE` metric. If Workers Cache returns a HIT, none of that code runs, so same-URL retries/re-downloads are not counted.
- Today `getHandler()` sees the original `Range` header and uses `object.range`/`calculateBytesTransferred()` to record the actual bytes sent. With Workers Cache enabled, a cold `Range` request is presented to the Worker as a full-body request, so the first partial download can be recorded as the full object size, and subsequent range hits record nothing.

There is also a negative-cache risk for missing files: many file responses only end up with `Cache-Control: no-transform` rather than `no-store` or a short explicit TTL, and Cloudflare's Workers Cache applies default TTLs when responses do not opt out, including 200s and 404s.

Please keep cache disabled on this default files entrypoint, or split the router/accounting layer from a separately cached inner entrypoint so `saveBandwidthUsage()` still runs before serving cached bytes. At minimum, `/files/read` responses and 404/error responses need explicit cache semantics and a design for range-byte accounting before this can be safe in prod.

Refs:
- https://developers.cloudflare.com/workers/cache/configuration/
- https://developers.cloudflare.com/workers/cache/cache-keys/

PR #2626, opened 2026-07-06
[codex] fix iOS marketing version sync before builds

riderx:
## What changed

- Added CLI-owned iOS marketing-version sync logic in `cli/src/build/ios-marketing-version.ts`.
- Added `build sync-ios-version` and `build request --sync-ios-version` to the CLI, with focused tests and generated CLI docs.
- Updated the iOS mobile build workflow to use the released CLI: `bunx @capgo/cli@latest build request --platform ios --path . --sync-ios-version --ai-analytics`.
- Removed the local `cli/dist/index.js` call from the iOS mobile build workflow.
- Kept `bump_version.yml` from calling the new command during the same release bump job, because the CLI package is only published after the `cli-*` tag is pushed.
- Synced the current Xcode project `MARKETING_VERSION` from `12.160.5` to `12.188.2`.

## Why

The failed TestFlight upload showed `BUILD_NUMBER` was bumped, but `CFBundleShortVersionString` stayed at `12.160.5`, matching the previously approved App Store version. Apple requires the marketing version to be higher than the approved version, so build-number bumping alone is not enough.

The upload surface is `build_mobile_ios.yml`, so that workflow now asks the released CLI to sync iOS `MARKETING_VERSION` before the cloud build request is sent.

## Validation

- `bun run --cwd cli test:ios-marketing-version`
- `bun run cli:typecheck`
- `bun run cli:lint`
- `bun run cli:build`
- `git diff --check`
- `bun run typecheck`
- `bun run lint` (passes with existing admin dashboard warnings only)

---

> [!NOTE]
> **Medium Risk**
> Changes iOS store versioning and CI build paths; incorrect sync could still produce App Store–rejected builds, but scope is localized to marketing version alignment and opt-in flags.
> 
> **Overview**
> Adds CLI support so iOS **CFBundleShortVersionString** stays aligned with `package.json` before a cloud build uploads the Xcode project—addressing TestFlight failures when only the build number was bumped.
> 
> **`build sync-ios-version`** derives a semver marketing version from `package.json` (strips prerelease/build metadata), updates all `MARKETING_VERSION` entries in `project.pbxproj`, and supports **`--check`** for CI drift detection. **`build request --sync-ios-version`** runs the same sync for iOS builds before upload.
> 
> The **iOS mobile GitHub workflow** now passes **`--sync-ios-version`** on `bunx @capgo/cli@latest build request`. **`bump_version.yml`** installs dependencies when Capgo (not only CLI) release scope changes. The repo Xcode project **`MARKETING_VERSION`** is updated to match the current package version, with tests and generated CLI docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35b68552df202db9abd49eb4c64ddf9ae0793856. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


riderx:
@coderabbitai review

riderx:
@coderabbitai review

RobinWitch (changes requested):
The iOS rebuild workflow now depends on CLI code from the checked-out release tag. The workflow still accepts an arbitrary `workflow_dispatch` tag and checks out that tag, but this PR changes the upload step from `bunx @capgo/cli@latest` to `bun run cli:build` plus `bun cli/dist/index.js ... --sync-ios-version`. That means after this workflow is merged, rebuilding any older release tag that predates the new CLI package/flag will fail before the build request is even created. For example, `v12.99.1` has no `cli:build` script / `cli/` tree, so the new `Prepare mobile project` step cannot build `cli/dist/index.js`; tags that have the CLI but predate this option would similarly reject `--sync-ios-version`.

This regresses the manual "build this tag" workflow for historical releases. Please keep the uploader CLI independent from the checked-out app tag (for example continue using the latest published CLI), or only pass the new sync option when the checked-out CLI supports it / move the MARKETING_VERSION sync into a workflow-side step that remains compatible with older tags.

PR #2634, opened 2026-07-07
feat(logs): add log insights dashboard

riderx:
## Summary
- Add Logs secondary tabs with an Insights dashboard and Raw logs view.
- Add `/private/stats/insights` aggregation for Cloudflare Analytics Engine, Supabase fallback, and demo data.
- Surface 1, 3, 7, and 30 day log insight periods with top priority, error categories, daily trend, affected versions, and affected devices.

## Screenshot
Log insights dashboard

## Tests
- `bun run lint` (passes with existing unrelated admin dashboard warnings)
- `bun run lint:backend`
- `bun run typecheck`
- `bunx vitest run tests/stats-insights.unit.test.ts`
- `bun run build`
- Headless Chromium capture from a live Vite dev server with mocked Supabase/API data

Note: local Supabase was not started because Docker was unavailable in this environment.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new authenticated analytics endpoint and multi-query aggregations over log data; UI is read-only but incorrect filters or query cost could affect ops visibility.
> 
> **Overview**
> App **Logs** is split into **Insights** (default) and **Raw logs** via new secondary tabs on `/app/:id/logs`, with the existing log table moved to `/logs/raw`.
> 
> The Insights view loads aggregated failure data from a new authenticated **`POST /private/stats/insights`** endpoint (1/3/7/30-day UTC windows, default failure actions). It shows summary cards, top-priority messaging, error categories, daily trend, and top versions/devices, with navigation into raw or device logs using period/action query params.
> 
> The backend adds **`readStatsInsights`** with parallel aggregations on Cloudflare Analytics Engine when `APP_LOG` is available, Postgres fallback via **`readStatsInsightsSB`**, demo-app derivation from generated logs, plus shared normalization in **`statsInsights.ts`** and a unit test for row grouping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2046038f38cad16eed9895b7695755a8116f2db1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


riderx:
@coderabbitai review

RobinWitch (changes requested):
Blocking issue: this PR moves the raw log table from `/app/:app/logs` to `/app/:app/logs/raw`, but the existing filtered-log entry points still navigate to `/app/:app/logs?...`.

`src/components/dashboard/UpdateStatsChart.vue` still pushes `/app/${props.appId}/logs?${params.toString()}` and `src/pages/app/[app].observe.vue` still pushes `{ path: `/app/${appRouteSegment.value}/logs`, query: { action, start, end } }`. After this PR, `/logs` renders the new Insights dashboard, which does not consume those query params; only `LogTable` on `/logs/raw` initializes its date/action filters from `route.query`. So clicking a dashboard chart segment or an Observe "open logs" action lands on Insights instead of the filtered raw logs view, and the selected action/time window is effectively dropped.

Please either update those callers to target `/logs/raw` while preserving the query string, or make `/logs` redirect/delegate to the raw page when `action`/`start`/`end` query params are present.

Also minor but user-visible: `src/pages/app/[app].logs.vue` calls `t('not-authenticated')`, but the existing key is `not-logged-in`, so the no-session toast will render a missing translation key.

RobinWitch:
Additional issue on the same endpoint: the new insights read path also turns backend query failures into a successful empty dashboard.

Both data-source implementations run several aggregate queries in parallel, but `readStatsInsightsCF()` catches any Analytics Engine failure and returns `emptyResult`, and `readStatsInsightsSB()` catches any Postgres/query failure and returns `emptyStatsInsights()`. The route then spreads that into `c.json(...)`, so `/private/stats/insights` still returns HTTP 200. The frontend only checks `response.ok`; a real Analytics token/account problem, rejected SQL, timeout, or Postgres error will therefore render `No failures found in this period` instead of the existing `failed-to-fetch-log-insights` error state.

This is different from a true empty period and makes operational/data-source failures invisible to users. I think these helpers should let errors propagate (or return an explicit error result that the route maps to non-2xx) rather than swallowing them into the empty-state payload.

PR #2637, opened 2026-07-07
perf(ci): parallelize test workflow for faster PR feedback

riderx:
## Summary (AI generated)

Two rounds of CI restructuring toward a ~2-minute test wall-clock:

**Round 1 — parallelize the monolithic job (merged into this PR earlier):**
- Split the serial `test_all` job (~9 min) into parallel jobs: lint/typecheck, backend tests, Cloudflare Workers tests
- Cache Playwright browsers keyed on `bun.lock`

**Round 2 — unit split + sharding:**
- New `Run unit tests` job: the 135 `tests/*.unit.test.ts` files (930 tests) run with **no Docker/Supabase at all**; previously they ran **twice** (once inside each integration suite)
- Backend and Cloudflare integration suites now exclude unit tests and run as 2-way `vitest --shard` matrices on parallel runners
- Dropped the post-start `supabase db reset`: per Supabase docs, seed files are applied on the first `supabase start` of a fresh runner, so the reset only repeated migrations + seed (~37s per Docker job)
- `bun install` / `cli:build` / PostgreSQL 17 client apt install run in the background while `supabase start` boots (backend, Cloudflare, and CLI integration jobs)
- DB-level checks (read-replica snapshot, `supabase test db`, SQL lint) run on shard 1 only
- New scripts `test:unit`, `test:backend:integration`, `test:cloudflare:integration`, documented in AGENTS.md

## Motivation (AI generated)

The `Run tests` job was the critical path of every PR (~8m45s of serial steps). Round 1 brought the wall to ~5 min by parallelizing suites. Round 2 attacks the remaining fat: duplicated unit-test execution, a redundant DB reset, serial dependency installs, and single-runner test execution. Expected wall after this round: **~2.5–3 min**, bounded below by the Docker Supabase boot (~70s + setup) that every integration shard still pays.

**Why not Tinbase (yet):** tinbase.dev was evaluated as a fast Supabase stand-in (~2s startup vs ~70s). Capgo's base migration requires `pg_cron`, `pg_net`, `pgmq`, `supabase_vault`, `hypopg`, and `plpgsql_check`; Tinbase explicitly does not implement full pg_cron/pg_net, so the schema cannot apply today. The unit-vs-integration split in this PR is the prerequisite either way: the 930 unit tests now run with no backend at all (faster than any emulator), and the remaining integration suite is the candidate to migrate if/when Tinbase gains the required extensions.

PR feedback drops from ~9 min to ~2.5–3 min, tightening the merge loop for every contributor and reducing wasted context-switching. Reliability improves: unit tests are isolated from Docker/network flakiness entirely, each integration shard runs on a freshly seeded database, and a flaky shard can be re-run alone. Trade-off: more parallel runners per PR (2+2 shards + unit job), which is a modest runner-minutes increase for a large wall-clock win.

## Test Plan (AI generated)

- [ ] CI on this PR runs `Run unit tests`, `Run backend tests (shard 1/2, 2/2)`, `Run Cloudflare Workers tests (shard 1/2, 2/2)`, lint/typecheck, and Playwright in parallel, all green (this PR touches `.github/workflows/**`, so every suite is triggered)
- [ ] Unit job passes with no Supabase (validated locally: 135 files / 930 tests / 35s, zero references to the local stack)
- [ ] Shard file sets validated locally with `vitest list --shard` (1536 integration tests split, units excluded)
- [ ] Integration tests pass without the explicit `db reset` (seed applied by first `supabase start`)
- [ ] `bump_version.yml` (workflow_call consumer) still gates on the whole workflow result

Generated with AI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> CI behavior changes (no explicit db reset, sharded integration coverage, Playwright service startup) could miss regressions or introduce new flakes if assumptions about fresh DB state or job gating are wrong.
> 
> **Overview**
> Replaces the single **`test_all`** job with parallel CI work: **`lint_typecheck`**, **`test_unit`** (no Supabase), **`test_backend`** and **`test_cloudflare`** as 2-way Vitest shards (unit + CLI tests excluded from shards), plus existing Playwright and other jobs unchanged in scope.
> 
> **CI speed and layout:** Drops the post-**`supabase start`** **`db reset`** where a fresh runner already gets migrations + seed; overlaps **`bun install`** (and PG17 client / **`cli:build`** where needed) with **`supabase start`**. DB checks (read-replica schema, **`supabase test db`**, SQL lint) run only on backend shard **1/2**. **`cli_capgo_integration`** now runs when Capgo backend changes too, not only CLI paths.
> 
> **Playwright job:** Caches browsers on **`bun.lock`**, starts Stripe/backend/frontend via **`nohup`** + **`wait-on`**, sets **`SKIP_SUPABASE_DB_RESET=true`** for the Playwright backend, and probes **`supabase:with-env`** before building the frontend preview.
> 
> **Scripts and docs:** Adds **`test:unit`**, **`test:backend:integration`**, **`test:cloudflare:integration`** in **`package.json`** and **`AGENTS.md`**.
> 
> **Test stability:** **`playwright.config.ts`** retains trace/video/screenshot only on failure in CI; **`fetchWithRetry`** / Supabase client fetch retry **504** as well as **503**; **`channel-rate-limit.test.ts`** uses sequential **`hitRateLimit`** bursts for non-atomic limiter windows under load.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3037419f487c0e2e8759c00261678f62763f94e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added separate test commands for unit, backend integration, and Cloudflare integration runs.
  * Reorganized CI to use faster, more targeted lint/typecheck and sharded integration test jobs.
* **Bug Fixes**
  * Improved Playwright CI recording to only retain traces, videos, and screenshots on failures.
* **Documentation**
  * Updated testing instructions with the new test-run commands.


RobinWitch (changes requested):
The Playwright job is currently deadlocking before the frontend ever starts. In the new background launcher, the frontend process waits on a bare `supabase status` before it runs `bun run supabase:with-env -- bun scripts/playwright-frontend-preview.ts`.

That does not check the same Supabase project that `backend:playwright` starts. `backend:playwright` calls `bun run supabase:start`, which goes through `scripts/supabase-worktree.ts` and appends `--workdir .context/supabase-worktrees/<hash>`; the plain `supabase status` in the workflow probes the default Supabase workdir instead. On a fresh GitHub runner only the worktree-isolated stack becomes healthy, so the loop never reaches `supabase:with-env` and the frontend log stays empty.

This matches the failing checks: both recent Playwright runs get the stripe port and backend ready file, then wait until `http-get://localhost:5173` times out. Please use the wrapper in the readiness loop (for example `bun run supabase:status`) or remove the extra status loop and let `supabase:with-env`/`wait-on` handle the readiness condition, so the frontend preview actually starts against the worktree stack.

riderx:
**Review feedback addressed (AI generated):**

@RobinWitch — spot on about the `supabase status` / worktree-workdir mismatch. Fixed in two steps: 67fa271db switched the readiness loop to the wrapper, and 1a637e62b went further because even the wrapper's plain `status` flips to exit-0 a moment before `API_URL` appears in its JSON — the loop now probes through `supabase:with-env` itself (requiring `SUPABASE_URL` + `SUPABASE_ANON_KEY` in a child process) before starting the vite build. Validated green in run 28883096092 (Playwright 3m36s).

@coderabbitai — CLI duplication fixed in 5506a5462: backend shards now pass `--exclude=tests/cli*` (and skip the CLI build), and `cli_capgo_integration` triggers on the capgo scope too, since `tests/cli*` exercise the backend edge functions and backend-only PRs must keep that contract coverage. Also moved the vitest `--shard` value into an env var to clear the zizmor template-injection warning.

PR #2639, opened 2026-07-07
fix(frontend): show app icons in org selector

riderx:
## Summary

- include app `icon_url` in the organization store app list
- resolve public app icons immediately and sign private storage paths asynchronously
- render app icons in the org selector with loading and initials fallback states

## Root Cause

The org selector app list only fetched `app_id`, `name`, and `owner_org`, so the dropdown had no icon data available and always rendered initials.

## Screenshot

Org selector app icons

## Validation

- `bun lint` passes with existing warnings in admin dashboard pages
- `bun typecheck`
- local headless demo screenshot captured as WebP

---

> [!NOTE]
> **Low Risk**
> UI and read-only icon loading in the org store, mirroring existing logo signing with no auth or data-model changes.
> 
> **Overview**
> The org selector app list now shows real **app icons** instead of always rendering initials.
> 
> The organization store **loads `icon_url`** with each app and applies the same pattern as org logos: public URLs are used immediately, private storage paths are **signed asynchronously**, with `icon_url_loading` driving a spinner until the signed URL is ready. The dropdown renders the image when available, a loading state while signing, and **initials as fallback** when there is no icon.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0d7b7346fcd9d5fdc273f237dbbe1f585738e06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


RobinWitch (changes requested):
I found one race in the new app-icon load path that should be fixed before this leaves draft. organizationAppIconLoadRun is only checked inside updateOrganizationAppIconState(), after the signed icon request resolves. The apps query itself can still apply stale results: if two initial fetchOrganizations() calls overlap while _organizationsByAppId is still empty, both watch(_organizations) callbacks pass the size guard and increment the run. If the newer query returns first, it assigns _appsByOrgId/_appsByAppId and starts signing icons. When the older query returns later, lines 467-470 overwrite those maps with fresh OrganizationApp objects whose private icons are back in icon_url_loading: true; then its own loadOrganizationAppIcons(..., oldRun) updates are ignored by the run check at lines 318-320. For signed storage icons this can leave the org selector permanently showing the loading spinner. Please add a post-query run check before assigning the maps/loading icons, or otherwise dedupe/cancel the initial app-map load so an older apps query cannot overwrite newer icon state.

riderx:
Addressed the stale async app-icon race in f0d7b7346 by checking the app icon load run immediately after the awaited apps query, before assigning app maps or starting icon loads.

PR #2640, opened 2026-07-07
feat(replica): embed read replicas in Cloudflare as per-app Durable Objects

riderx:
## What

Moves the read-replica layer for update checks **inside Cloudflare**, replacing the external Google Cloud SQL fleet (1 subscriber + 9 regional replicas + Hyperdrive hops) with per-app SQLite replicas running as Durable Objects, exposed on a **parallel endpoint `/updates_v2`** so old and new can be load-balanced in production until the old system is decommissioned.

D1 was rejected on purpose: a single shared database has a single throughput ceiling (we already hit it with `DB_STOREAPPS`). One DO per `${region}:${app_id}` spreads reads across as many isolates as there are active apps — each app only competes with itself, and hot apps are further split 9 ways by region.

## Architecture

```
Supabase Postgres (main)
  │  row triggers on the 7 hot-path tables (apps, orgs, stripe_info,
  │  channels, app_versions, channel_devices, manifest)
  ▼
public.replicate_outbox          ordered change log + routing keys
  │  DELETE .. FOR UPDATE SKIP LOCKED, poll every 5s
  ▼
ReplicaRouter DO                 journal + registry (the "1 replica linked
  │                              to Supabase")
  │  push: app rows → that app's replicas, org rows → all replicas of org
  ▼
AppReplica DO per region:app     SQLite slice of one app (+ its org rows)
  ▲  lazy self-seed from the read replica on first read in a region
  │
capgo_plugin /updates_v2         2 region-local RPCs (~1ms) per check
```

## No Postgres fallback — by design

The old system gets destroyed after the transition, so `/updates_v2` **never reads Postgres**. If a replica is not ready (warming up, lease expired, router down), the device gets a plain **"no update"** (`edge_replica_not_ready`, kind `up_to_date`) and retries on its next check while the replica seeds itself in the background. Worst case is a delayed update — never broken devices, never read traffic landing on the main DB.

`/updates` is completely untouched (pure Postgres) and acts as the control group during the load-balanced transition. Both endpoints return identical response shapes.

## Reliability

- **Exactly-once feed**: outbox rows are copied into the router journal before the Postgres delete commits; per-target cursors advance only on a successful push; replicas apply idempotent upserts in outbox order.
- **Why not Supabase Realtime**: `postgres_changes` is fire-and-forget over WebSocket — dropped connections lose rows, no replay cursor, cannot seed. The outbox is transactional and resumable after any downtime.
- **Freshness = lease** (15 min, refreshed by every push/heartbeat; typical data lag ≈ 5s poll vs the 180s tolerated on Cloud SQL today). Reads past the lease answer "no update", not stale data.
- Flaky replicas: 3 failed pushes → invalidate + reseed on next read. `POST /invalidate-all` forces a fleet-wide reseed after schema changes. Replicas idle 7 days self-evict and wipe.

## Cost / load (sized for 55M monthly / 5M daily devices)

- Main DB load: tiny trigger appends on writes + one indexed poll query every 5s. Device reads never touch Postgres.
- The dominant Cloudflare cost is **DO duration** (128 MB x wall-clock while in memory; a request-driven DO stays ~10s in memory after its last request). Requests, rows read (25B/mo included), and storage are noise at this volume.

| checks/device/day | checks/mo | DO duration (est.) | everything else | total |
|---|---|---|---|---|
| 1.0 | 150M | $1,300-2,300 | ~$100-230 | **~$1.4k-2.5k/mo** |
| 1.5 | 225M | $2,000-3,500 | ~$100-250 | **~$2.1k-3.7k/mo** |
| 3.0 | 450M | $4,000-7,000 | ~$180-380 | **~$4.2k-7.4k/mo** |

Roughly cost-neutral to cheaper than the Cloud SQL fleet (10 instances 24/7 + cross-cloud egress, typically $3k-5k/mo at list), while moving reads in-colo (~1ms) and deleting the fleet ops. Replicas are created lazily by traffic and self-evict after 7 idle days, so cost tracks active apps.

**Known follow-up lever**: caching channel resolution in the plugin worker (Cache API, 30-60s TTL, apps without device overrides only) keeps most checks off the DOs entirely and should cut the duration term 2x or more — worth adding before ramping to 100%.

## Rollout

1. Apply the migration (outbox + triggers).
2. `wrangler secret put REPLICATOR_SECRET ... && bun run deploy:cloudflare:replicator:prod`
3. Per plugin region: uncomment the `APP_REPLICA` binding + `"EDGE_REPLICA_MODE": "on"`, deploy. Without the flag, `/updates_v2` just answers "no update" — deploying is always safe.
4. Load-balance a share of device traffic from `/updates` to `/updates_v2`, watch `X-Database-Source: edge_replica` vs `edge_replica_not_ready` and response parity, ramp to 100%.
5. Decommission Cloud SQL; keep one small replica (or the pooler) as the per-app seed source.

Merging this PR changes nothing by itself: `/updates` is untouched and `/updates_v2` is inert until the binding + flag are enabled.

## Verified

- Query builders + reader mapping validated against **real SQLite** (plan validation incl. trial/exceeded flags, channel/device-override/rollout/builtin paths, manifest aggregation, not-ready → no-update path).
- Migration applied + rolled back on local Supabase.
- Both workers bundle via `wrangler deploy --dry-run`.
- `bun run typecheck`, `oxlint`, full unit suite green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
> Touches the device OTA update hot path and introduces distributed replication with lease semantics; mis-rollout could delay updates (no-update responses) until replicas warm, though `/updates` remains unchanged until traffic shifts.
> 
> **Overview**
> Introduces a **Cloudflare-embedded read replica** path for device update checks, meant to replace regional Cloud SQL reads while **`/updates` stays on Postgres** for a side-by-side rollout.
> 
> A new **`capgo_replicator`** worker adds **`ReplicaRouter`** and per-region **`AppReplica`** Durable Objects (SQLite per `${region}:${app_id}`). Postgres **`replicate_outbox`** plus triggers on seven hot tables feed the router; it drains the outbox, journals changes, and fans out to replicas with lease-based freshness and reseed on failure.
> 
> The plugin worker exposes **`/updates_v2`**, wired to **`updateV2`**, which reads only via **`EdgeReplicaReader`** RPCs to `AppReplica`—**no Postgres on the read path**. Unready replicas return **`edge_replica_not_ready`** (treated as up-to-date) instead of falling back. **`updateWithPG`** accepts an optional edge reader for owner/channel lookups. Plugin wrangler envs get commented **`APP_REPLICA`** bindings and **`EDGE_REPLICA_MODE`** rollout notes; deploy scripts and unit tests for the shared **`edge_replica_schema`** are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de6c364e5db82389a25bbf106203bd6d6471e59e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a new `/updates_v2` endpoint for edge-replica-backed updates alongside existing `/updates`.
  * Enabled region-aware embedded read-replica support with fail-fast “no update” behavior during warmup/lease expiry.
* **Bug Fixes**
  * Improved replica availability handling: when the edge replica is not ready or stale, requests now return “no update” for safer retries.
* **Documentation**
  * Added detailed setup, rollout, and operational guidance for the edge-replica system, including admin/status flows.
* **Chores**
  * Updated ignore rules for generated replica artifacts.


RobinWitch (changes requested):
Rechecked the current head (d0f34cb). The freshness heartbeat can mark the D1 replica as safe while unapplied outbox rows are still waiting.

`drainOutbox()` writes `last_applied_at = now()` after every D1 batch, and the reader's `assertFresh()` only checks that timestamp. If the outbox backlog is larger than one alarm can drain (`MAX_BATCHES_PER_ALARM * OUTBOX_BATCH_SIZE`), the worker exits with rows still pending but a fresh heartbeat, so `/updates` will read stale D1 data instead of falling back to Postgres. The same happens right after seeding: `seedStep()` sets both `seeded_at` and `last_applied_at` before the accumulated seed-window outbox rows have been replayed, then waits until the next poll to start draining them.

This breaks the “never worse than today” fallback guarantee. A channel/version change made during seed, or any sustained backlog, can be invisible to devices while `X-Database-Source: edge_replica` is returned as fresh. Please keep the replica unavailable until the seed-window backlog is drained, and make freshness depend on outbox catch-up (for example, no pending rows older than the max lag, or a source commit/outbox watermark) rather than only on the worker's wall-clock poll time. A regression should leave more pending outbox rows than one alarm drains, or enqueue a channel change during seed, and assert the reader falls back instead of serving D1.

RobinWitch:
One more seed-safety issue, separate from the heartbeat/backlog problem above: the initial seed needs a main/outbox-source watermark before it can safely read from an async replica.

`HYPERDRIVE_SEED` is documented/configured as the existing EU read replica, but `replicate_outbox` only captures writes after this migration installs its triggers. If that seed replica is behind main when `/seed` starts, rows committed before the outbox trigger existed but not yet replayed to the seed replica are neither copied by the keyset seed nor present in `replicate_outbox`. Once the seed cursor advances past those rows, D1 can permanently miss apps/channels/versions/manifests until a full reseed.

Please either seed from the main/outbox source, or record/wait for a WAL/LSN/catch-up watermark proving the seed replica has replayed through the outbox-install point before keyset seeding begins.

riderx:
All review findings are addressed in bd6ec0f:

**@RobinWitch — seed watermark**: fixed. `AppReplica.openSeedClient()` now takes `pg_current_wal_lsn()` on the outbox source first, then polls the seed replica's `pg_stat_subscription.latest_end_lsn` until it has replayed past that watermark before any keyset read (a non-subscriber seed source passes trivially). If it doesn't catch up within 10s, the snapshot is taken from the outbox source directly. Rows committed before trigger install / before registration can no longer be missed by both the snapshot and the pushes.

**@RobinWitch — heartbeat/backlog freshness** (your review referenced the old D1 head, but the same class of issue existed in the DO rework): fixed at the source of truth. `drainOutbox()` now reports whether the outbox was fully consumed; while a backlog remains the router loops immediately (250ms), pushes are sent with `extendLease: false` and heartbeats pause — so a lease can only ever assert "caught up as of grant time". A sustained backlog lets leases expire and `/updates_v2` answers plain "no update" (there is no Postgres fallback anymore by design — the old system will be destroyed). The seed-window concern is covered by the reworked seed protocol: register → snapshot → replay buffered pushes → only then mark seeded.

**Cursor Bugbot**:
- *Seed lacks outbox watermark* — LSN watermark above; additionally the org-move path no longer goes through `register` (which resets the cursor): a new `updateOwnerOrg` router RPC updates the routing key while leaving the push cursor untouched.
- *Unseeded replica acks journal pushes* — an unseeded replica with no seed in flight now answers `{ unregister: true }` instead of `{ ok: true }`, and a failed seed unregisters itself, so the router can never advance a cursor past rows that were dropped.
- *Pending batches cleared after register* — seed order is now: unregister (no pushes can arrive) → clear buffer → watermark → register → snapshot → replay buffer. The buffer cleared at the start can only contain leftovers from a previous failed attempt.

**CodeQL** — admin endpoints return a generic `internal error` 500; details stay in the DO logs.

**SonarQube** — all 12 issues fixed (readonly members, `.at(-1)`, single `push`, extracted nested ternary, object-safe stringification, `export … from`, complexity reduced by splitting `seed()` into `openSeedClient`/`snapshotTable`).

Verified: typecheck, oxlint, unit tests, reader SQL validation against real SQLite, and `wrangler deploy --dry-run` for both workers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

riderx:
Second round of findings fixed in 72ab3e6:

**Bugbot — heartbeat extends lease while lagging**: the maintenance heartbeat now only targets replicas whose cursor is at the journal head (`cursor >= journal_head`). A lagging target can only get its lease refreshed by a successful data push, so a lease always means "caught up as of grant".

**Bugbot — journal prunes unconsumed rows**: before age-pruning, any registered target whose cursor is still behind the aged window is unregistered and invalidated, so it reseeds on its next read instead of silently diverging forever. (In practice a target can't lag an hour — 3 failed pushes ≈ 15s already unregister it — but the guard makes it airtight.)

**Bugbot — org rows skipped on move**: rethought. Patching the routing key (`updateOwnerOrg`) was both racy against the cursor advance *and* insufficient — after a move the replica also needs the new org's current `orgs`/`stripe_info` rows, which may never appear in the journal. An org move now triggers a full reseed of that replica (unregister → snapshot with the new owner → re-register); reads answer "no update" during the reseed. A move detected while replaying the seed-window buffer re-runs the seed once with the fresh owner. Org moves are rare and per-app seeds are small, so this is the simple, airtight path. `updateOwnerOrg` is removed.

**CodeRabbit — seed failure backoff**: a failed seed sets `seed_backoff_until` (+60s) in the replica meta; not-ready reads inside the window answer "no update" without opening new Postgres connections.

Verified: typecheck, oxlint, unit tests, SQLite validation of the reader path, wrangler dry-run bundles, and the CI console-guard grep.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

riderx:
Closing: DO-per-app read fleet loses on Durable Object duration billing at 83M req/day (~$15.6/M touched). Superseded by colo-cache + targeted invalidation approach (new PR incoming).

PR #2653, opened 2026-07-10
perf(ci): speed up migration order check

riderx:
## Summary
- skip the network fetch in GitHub pull request and merge-group runs when the checkout already has the synthetic merge commit
- update the migration timestamp parser to avoid per-file `basename` subprocesses and compute the latest base timestamp during the base scan
- add a unit test that fails if the PR merge-commit path tries to run `git fetch`

## Root cause
The migration-order check looked text-only, but it always fetched the target branch and spawned extra shell work while scanning every base migration. In GitHub PR checkouts the base tree is already available as `HEAD^1`, so the fetch is avoidable.

## Validation
- `bash -n scripts/check-supabase-migration-order.sh`
- `/usr/bin/time -p bash scripts/check-supabase-migration-order.sh`
- `bunx vitest run tests/check-supabase-migration-order.unit.test.ts`
- `bun lint`
- `bun test:unit`
- `bun typecheck`
- `bunx eslint tests/check-supabase-migration-order.unit.test.ts`

---

> [!NOTE]
> **Low Risk**
> Changes only affect the migration filename/order guard script and its tests; no runtime app or database behavior.
> 
> **Overview**
> **Speeds up the Supabase migration order CI check** by avoiding unnecessary network and shell work while keeping the same validation rules.
> 
> On **GitHub `pull_request` and `merge_group` runs**, the script now prefers **`HEAD^1`** as the comparison ref when the synthetic merge commit is present, so it **skips `git fetch`** when the base tree is already local. Fetch fallbacks and a **`git merge-base`** check still apply when fetch fails or no merge base exists.
> 
> **`extract_timestamp`** no longer shells out to `basename`; it uses Bash path expansion and can write into a variable. The **latest base migration timestamp** is tracked in the same loop that builds the base file list instead of a separate `awk` pass.
> 
> A new **Vitest unit test** builds a mini repo with a merge commit, stubs `git fetch` to fail if called, and asserts the script uses the local merge parent and passes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbfe9cef6de051ecad1b25e00ea63d8288ddfd95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/check-supabase-migration-order.sh
+++ b/scripts/check-supabase-migration-order.sh
@@ -4,14 +4,37 @@ set -euo pipefail
 
 extract_timestamp() {
   local file_name="$1"
+  local out_var="${2:-}"
   local base_name
-  base_name="$(basename "$file_name")"
+  base_name="${file_name##*/}"
 
   if [[ "$base_name" =~ ^([0-9]{14})_.+\.sql$ ]]; then
-    echo "${BASH_REMATCH[1]}"
+    if [[ -n "$out_var" ]]; then
+      printf -v "$out_var" '%s' "${BASH_REMATCH[1]}"
+    else
+      printf '%s\n' "${BASH_REMATCH[1]}"
+    fi
   else
+    if [[ -n "$out_var" ]]; then
+      printf -v "$out_var" ''
+    fi
     return 1
   fi
+}
+
+resolve_github_merge_base_ref() {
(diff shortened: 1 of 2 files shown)


riderx:
@coderabbitai review

PR #2095, opened 2026-05-10
[codex] Fix missed component release recovery

riderx:
## Summary (AI generated)

- Make release scope resolution compare each component against its latest matching tag instead of only the latest push range.
- Prevent reusable release test runs from cancelling other release test runs on `main`.
- Add release-scope regression coverage for missed CLI changes after a later Capgo-only push.

## Motivation (AI generated)

A cancelled CLI release test run could lose the CLI release forever because the next successful main push only inspected its own push range. The Capgo release could still ship and include CLI changes in the generated changelog, but no `cli-*` tag or npm publish would happen.

## Business Impact (AI generated)

This keeps Capgo and CLI releases aligned when both products changed, so CLI fixes do not silently miss npm publication after CI cancellation or later main pushes. It reduces manual release recovery and avoids users waiting for already-merged CLI fixes.

## Test Plan (AI generated)

- [x] `bunx vitest run tests/release-scope.test.ts`
- [x] `bunx eslint --no-ignore scripts/release-scope.ts tests/release-scope.test.ts`
- [x] `git diff --check`
- [x] Commit hook: `bun run cli:build && vue-tsc --noEmit`

## Summary by CodeRabbit

* **Chores**
  * Updated workflow concurrency grouping logic for reusable workflows.
  * Improved internal release scope script with dependency-injected Git execution for enhanced testability.
  * Expanded test coverage for release scope scenarios, including tag selection and fallback handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Cap-go/capgo/pull/2095)

PR #2656, opened 2026-07-10
fix(ci): filter releases by pushed paths

riderx:
## Summary

- restore release detection to the current GitHub push range (`github.event.before..github.sha`)
- keep existing component path matchers as the release authority
- remove tag-based missed-release recovery that published old components on unrelated pushes
- build CLI and notifications changelogs from each component's last successful GitHub release
- prevent release-scope logic/tests from being treated as package content

## Behavior

- unrelated later push: does not publish the failed component
- next same-component change: publishes cumulative code, including the previously failed change
- release text: spans from the last successful release for that component, so it includes both the previously failed and new changes
- rerunning the original failed workflow remains the immediate recovery path when no new component change exists

## Root cause

PR #2095 changed release detection from the current push range to each component's latest tag. The path matcher worked, but it received an oversized commit range. Separately, `mistricky/ccc` defaults to the immediately previous repository tag, mixing Capgo, CLI, and notifications changelog ranges.

## Validation

- release-scope regression failed before the fix with a false component release
- `bunx vitest run tests/release-scope.test.ts` (11 tests)
- `bun test:unit` (139 files, 946 tests)
- `bun lint`
- Prettier workflow validation
- historical Capgo-only commit `a23e043a`: CLI and notifications both return `should_release=false`
- real CLI and notifications feature commits still return `should_release=true`
- live release lookup resolves `cli-8.25.12` and `notifications-0.1.11` as the current last successful component releases

## Existing main blocker

Current `main` has unrelated generated Supabase type drift: `country_code` was removed from both generated type files while device code still uses it. That currently fails repository typecheck before this change.

---

> [!NOTE]
> **Medium Risk**
> Changes when CLI/notifications publish and what commits appear in release notes; incorrect `from_tag` resolution would fail the workflow or produce wrong changelogs, but runtime app code is untouched.
> 
> **Overview**
> **Release detection** no longer walks from each component’s latest tag back to `HEAD`. `resolveReleaseScope` again only inspects commits in the **current push** (`before..after`), so a Capgo-only push won’t trigger CLI or notifications publishes for older component commits that were never released.
> 
> **Changelog generation** for CLI and notifications now sets `from_tag` on `mistricky/ccc` by resolving the **previous non-draft GitHub release** with the matching tag prefix (`cli-` / `notifications-`), instead of defaulting to the repo’s immediately previous tag (which mixed Capgo, CLI, and notifications).
> 
> **Release-scope maintenance** is decoupled from package releases: `scripts/release-scope.ts` is removed from shared path matchers, and tests assert that edits to release-scope logic/tests don’t match any component for publishing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f151ebdd6ec57c9e5148d1979e7d72075efb6677. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish_cli.yml
+++ b/.github/workflows/publish_cli.yml
@@ -45,6 +45,17 @@ jobs:
             echo "::error::CAPGO_KEYCHAIN_HELPER_PATH leaked into the release bundle — dead-code elimination failed"
             exit 1
           fi
+      - name: Resolve previous successful CLI release
+        id: changelog_base
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          previous_tag=$(gh release list --repo "$GITHUB_REPOSITORY" --limit 1000 --json tagName,isDraft | jq -r --arg current "$GITHUB_REF_NAME" --arg prefix "cli-" '[.[] | select(.isDraft == false and .tagName != $current and (.tagName | startswith($prefix)))] | first | .tagName // empty')
+          if [ -z "$previous_tag" ]; then
+            echo "::error::No previous successful CLI release found"
+            exit 1
+          fi
+          echo "from_tag=$previous_tag" >> "$GITHUB_OUTPUT"
       - name: Generate AI changelog
         id: changelog
         uses: mistricky/ccc@v0.2.6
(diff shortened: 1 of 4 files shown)


riderx:
@coderabbitai review

PR #2665, opened 2026-07-10
fix(replica): verify additive schema sync

riderx:
## Summary

- make the read-replica catalog verification query fresh through Hyperdrive
- re-read the catalog after additive DDL before reporting statements as applied
- cover stale and fresh post-DDL catalog responses with unit tests
- restore generated database types for migrations that have not reached the remote schema yet

## Root cause

The release checker returned its planned DDL as `applied` without first proving that a subsequent catalog read could see it. The following Hyperdrive catalog request could therefore report an older schema view and falsely block the deployment even though the columns and index had been created.

The release workflow stops before applying these migrations, while the automatic remote-type sync removed their generated fields. That made the baseline backend typecheck fail even though the migrations and callers were already committed.

## Impact

The checker now fails the sync endpoint when an applied column or index is still absent or invalid after the post-DDL catalog read. It also keeps the generated type mirrors aligned with the pending migrations, so compiler checks stay truthful until the release applies them.

The pre-existing unexpected replica indexes remain a separate, intentional guard failure and require a maintenance operation; this PR does not hide or remove them.

## Validation

- `bun lint`
- `bun typecheck`
- `bun test:unit` (143 files, 960 tests; also passed from a clean worktree)
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Touches release-time read-replica schema DDL and deployment gating; failures are explicit but incorrect behavior could still block or delay releases.
> 
> **Overview**
> Fixes false release failures where additive DDL was reported as **applied** before a follow-up catalog read could see it (often due to **Hyperdrive** caching).
> 
> **Read-replica additive sync** now re-reads the schema catalog after DDL and only returns `applied` when every executed statement shows up in that fresh catalog; otherwise it throws with the missing column/index names. Invalid indexes still count as not persisted.
> 
> The catalog SQL is wrapped with a **`CURRENT_TIMESTAMP`** predicate so post-DDL verification reads are treated as non-cacheable and reflect the same release run.
> 
> Generated **Supabase types** are restored for columns/stats fields tied to migrations not yet on remote, keeping typecheck aligned with committed callers.
> 
> Unit tests cover stale vs fresh post-DDL catalogs and the Hyperdrive-oriented catalog query shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dcc1cce17f9f995acad8c64efb259586d338597f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/read_replicate/schema_additive_sync.ts
+++ b/read_replicate/schema_additive_sync.ts
@@ -201,10 +201,26 @@ export async function applyReadReplicaAdditiveSchemaSync(client: Queryable, expe
     await client.query('RESET statement_timeout')
   }
 
+  assertAppliedStatementsVisible(plan.statements, expected, await readReplicaSchemaCatalog(client))
+
   return {
     applied: plan.statements.map(({ kind, table, name }) => ({ kind, table, name })),
     skipped: plan.skipped,
   }
+}
+
+function assertAppliedStatementsVisible(applied: readonly SyncStatement[], expected: unknown, actual: unknown): void {
+  const remaining = new Set(planReadReplicaAdditiveSchemaSync(expected, actual).statements.map(statementKey))
+  const unavailable = applied.filter(statement => remaining.has(statementKey(statement)))
+  if (!unavailable.length)
+    return
+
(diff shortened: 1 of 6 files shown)


riderx:
@coderabbitai review

PR #2667, opened 2026-07-11
fix(db): align global_stats base schema with production access model

riderx:
## Summary (AI generated)

- Remove the permissive `Allow anon to select` policy from `global_stats` in the squashed base migration.
- Remove `anon` / `authenticated` grants on `global_stats` in base, keeping `service_role` only.
- Keep the existing `20260308121933_restrict_global_stats_access.sql` migration unchanged (drop + revoke), matching production.

## Motivation (AI generated)

Production already restricts `global_stats` to service-side access (RLS enabled, no client policies, `service_role` only). Dev still inherited the old permissive policy/grants from `base.sql`, which triggered the Supabase linter warning until the later restrict migration ran. This aligns the table declaration with the production end state.

- Closes the gap between dev schema linting and production security posture for sensitive global metrics (MRR, plan mix, etc.).
- No behavior change for production deployments that already applied the restrict migration.

## Test Plan (AI generated)

- [ ] `bun run supabase:db:reset` and confirm `global_stats` has RLS enabled, zero policies, and no `anon`/`authenticated` table privileges
- [ ] Run RLS tests (`26_test_rls_policies.sql`, `27_test_rls_scenarios.sql`)

Generated with AI

### DIFF
--- a/supabase/migrations/20250530233128_base.sql
+++ b/supabase/migrations/20250530233128_base.sql
@@ -4362,6 +4362,10 @@ SELECT
     )
   );
 
+CREATE POLICY "Allow anon to select" ON "public"."global_stats" FOR
+SELECT
+  TO "anon" USING (true);
+
 CREATE POLICY "Allow apikey to read" ON "public"."stats" FOR
 SELECT
   TO "anon" USING (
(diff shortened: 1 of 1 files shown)


riderx:
Closing: no schema diff vs `main` is required to match production.

Production state for `global_stats` is already defined by `20260308121933_restrict_global_stats_access.sql` (merged in #1749):
- RLS enabled
- zero policies
- `service_role` only (anon/authenticated revoked)

The Supabase "RLS enabled, no policy" advisor warning is therefore expected in dev **and** prod. Adding a deny-all policy would be *more* than production.

Editing `base.sql` to remove the original permissive policy/grants was attempted, but CI blocks modifying committed migrations (`check-supabase-migration-order.sh`).

If dev still shows the old permissive policy, run `bun run supabase:db:reset` so `20260308121933` is applied.

PR #3237, opened 2026-09-01
feat(onboarding): recommend WebNativeApp to hosted builders

AdrienADV:
## Summary

- Ask pre-organization onboarding users how they currently build and publish their app before the existing Capgo intent question.
- Recommend WebNativeApp to users coming from hosted AI or web builders, with a referred website CTA and an option to continue with Capgo.
- Persist the selected development environment and track the new onboarding interactions in analytics.
- Add unit coverage for UI copy, persistence, resume behavior, and analytics events.

## Test plan

- [x] bun run lint:backend
- [x] bun run lint
- [x] bun typecheck
- [x] bun test:unit

## Screenshots

Not included. The local browser preview was unavailable in the current environment; the UI states and copy are covered by integration-style unit tests.

## Checklist

- [x] My code follows the code style of this project and passes bun run lint:backend and bun run lint.

---
<a href="https://app.blacksmith.sh/Cap-go/codesmith/capgo.app/pr/3237"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790837152&installation_model_id=430928&pr_number=3237&repository=Cap-go%2Fcapgo.app&return_to=https%3A%2F%2Fgithub.com%2FCap-go%2Fcapgo.app%2Fpull%2F3237&signature=f6ccc865b450f69a82fc41dea6ddcee2a2322b93faefdd5c5b5cb17687778221"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

<a href="https://cubic.dev/pr/Cap-go/capgo.app/pull/3237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added onboarding choices for development environment and publishing an app.
  - Added WebNativeApp recommendations and a guided option to continue with Capgo.
  - Onboarding selections now persist and can be restored when resuming.
  - Added localized English copy and visual guidance for the new onboarding steps.
- **Bug Fixes**
  - Improved onboarding progress and resume handling for newly supported steps.
  - Admin onboarding funnel reporting now includes the publish-app question.
- **Analytics**
  - Added tracking for development-environment selections, publish intent, and WebNativeApp interactions.


supabase/functions/_backend/public/organization/post.ts
```diff
@@ -21,7 +21,7 @@ const bodySchema = z.object({
   email: z.email().optional(),
   estimatedMau: estimatedMauSchema.optional(),
   website: z.string().optional(),
-  intent: z.enum(['ota', 'builder', 'both', 'exploring', 'unknown']).optional(),
+  intent: z.enum(['ota', 'builder', 'both', 'exploring', 'publish', 'unknown']).optional(),
```
dimin4241-svg, line 25: This accepts publish, but public.orgs still has orgs_onboarding_valid, whose allowed intent array is only unknown/ota/builder/both/exploring. The organization insert stores the request intent, so a treatment user choosing Publish reaches a PostgreSQL check-constraint failure instead of completing onboarding. The same PR also writes publish into users.onboarding, while users_onboarding_valid allows only the four old user intents; NOT VALID still enforces the constraint on new/updated rows. Please add a migration that updates both constraints (and the schema snapshot), plus DB-backed regression coverage for saving progress and creating an org with publish.

dimin4241-svg (changes requested):
The new publish intent needs matching database constraints before this experiment can ship. Both affected write paths currently fail against production schema checks; details below.

supabase/functions/_backend/public/organization/post.ts
WcaleNieWolny, line 25: I checked as well, SQL check will disallow this. migration is needed

supabase/functions/_backend/utils/ab_tests.ts
```diff
@@ -166,6 +185,42 @@ async function persistABTestAssignments(
...
 }
 
+export async function getOrCreateUserABTests(
+  c: Context<MiddlewareKeyVariables>,
+  userId: string,
+) {
+  const pgPool = getPgClient(c)
+  let user: AssignmentUser | undefined
+  try {
+    const pgClient = await pgPool.connect()
```
WcaleNieWolny: Connection to db can be read only. It will use replication. If for whatever reason, replication doesn't have this data, use main DB.

No need to connect to write-allowed main DB. This will use valuable resources on main DB

supabase/functions/_backend/utils/org_onboarding_intent.ts
```diff
@@ -42,7 +42,7 @@ export function buildOnboardingIntentBentoEventData(
   const onboardingUrlBuilder = baseUrl ? `${baseUrl}/apps` : null
 
   let onboardingUrl: string | null = onboardingUrlOta
-  if (intent === 'builder')
+  if (intent === 'builder' || intent === 'publish')
```
WcaleNieWolny, line 60: what is the purpose of this code? what are you trying to do? explain please, i do not understand

src/utils/onboardingABTests.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export function resolveOnboardingAnalyticsVersion(onboarding: unknown): OnboardingAnalyticsVersion {
+  if (hasWebNativeDevelopmentEnvironmentTreatment(onboarding))
+    return WEBNATIVE_DEVELOPMENT_ENVIRONMENT_ANALYTICS_VERSION
+  if (hasWebNativePublishIntentTreatment(onboarding))
+    return WEBNATIVE_PUBLISH_INTENT_ANALYTICS_VERSION
+  return ONBOARDING_ANALYTICS_VERSION
+}
+
+export function shouldShowWebNativeRecommendation(options: {
```
WcaleNieWolny, line 78: I do not understand? I though showing recommendation was only suppose to happen when the user has the 

WEBNATIVE_DEVELOPMENT_ENVIRONMENT_AB_TEST (C) and also has selected hosted_builder. I though that the purpose of WEBNATIVE_PUBLISH_INTENT_AB_TEST was gathering data WITHOUT affecting the user flow/experience?

src/components/dashboard/AppOnboardingFlow.vue
```diff
@@ -639,7 +751,8 @@ function showWelcomeOnDesktop() {
   welcomePending.value = Boolean(props.preOrg && welcomeCanvasEligible.value)
 }
 
-function continueFromWelcome() {
+async function continueFromWelcome() {
+  await refreshOnboardingABTests()
```
WcaleNieWolny: Why? During welcome screen it's the perfect time to do A/B assignments. This is too late for me and this adds useless delay. If the backend is slow to respond, onboarding will not continue. Ideally, we would do refreshOnboardingABTests() when user sees welcome screen. Here we would await the result of refreshOnboardingABTests() (max 3 seconds, otherwise continue without waiting for `refreshOnboardingABTests()`)

If user doesn't see welcome screen, here we do `refreshOnboardingABTests()` and also wait (max 3 seconds)

supabase/functions/_backend/private/onboarding_ab_tests.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Hono } from 'hono/tiny'
+import { getOrCreateUserABTests } from '../utils/ab_tests.ts'
+import { quickError, useCors } from '../utils/hono.ts'
+import { middlewareAuth } from '../utils/hono_middleware.ts'
+
+export const app = new Hono<MiddlewareKeyVariables>()
+
+app.use('/', useCors)
+
+app.post('/', middlewareAuth(), async (c) => {
```
WcaleNieWolny, line 11: I think this should be atomic? You:

1. Check if user has all assignments already using replication DB (`getPgClient(c, true)`). If user has all assignments for ALL A/B tests, close connection and return.
2. Close connection to read-only `getPgClient`
3. Open new `getPgClient(c, false)` - using main DB (not read only replication)
4. lock the user row in a SQL transaction (use the drizzle SQL transaction wrapper)
5. Check if user already has ALL assignments, return (and close connection) if user does have everything (you do this to avoid TOCTOU)
5. You generate all assignments DURING the SQL transaction
6. You set the fields for onboarding JSON (user row)
7. You commit (just return from the wrapper of drizzle SQL transaction) and close the connection 

This is kind of racy - race conditions are possible here, which is bad.

supabase/functions/_backend/utils/ab_tests.ts
```diff
@@ -166,6 +185,42 @@ async function persistABTestAssignments(
...
+  }
+
+  if (!user)
+    quickError(404, 'user_not_found', 'User not found')
+
+  const candidates = createABTestAssignments(user)
+  if (Object.keys(candidates).length === 0)
+    return {}
+
+  return await persistABTestAssignments(c, userId, candidates)
```
WcaleNieWolny: You have zero checks if the user is not getting "re-assigned" - calling this endpoint multiple times might alter existing assignments which will destroy Bento's flows and cause undefined behavior.

WcaleNieWolny (changes requested):

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,62 @@
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT "orgs_onboarding_valid";
+
+ALTER TABLE "public"."orgs"
+ADD CONSTRAINT "orgs_onboarding_valid" CHECK (
```
WcaleNieWolny, line 5: Unrelated to this PR, but there are some other fields that are now being written to `orgs.onboarding`, and if you feel like it (I'll merge even if you don't do this), you could add the validation for these extra fields. Mainly the `starting_out` field.

```json
{"intent": "ota", "starting_out": false}
```

If you feel like it, you could add extra verification that:

1. if `starting_out` is present, it's a boolean

supabase/functions/_backend/private/onboarding_ab_tests.ts
WcaleNieWolny: resolved in latest commit, thank you 😊

supabase/migrations/20260902132624_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,62 @@
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT "orgs_onboarding_valid";
```
WcaleNieWolny: FYI (not 100% needed, I'll merge as is): sometimes in migrations, we check if a constraint exists before dropping with `DROP CONSTRAINT IF EXISTS`

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,62 @@
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT "orgs_onboarding_valid";
+
+ALTER TABLE "public"."orgs"
```
WcaleNieWolny, line 4: Also, I partialy agree with AI. Please do `ADD CONSTRAINT "orgs_onboarding_valid" CHECK (...) NOT VALID;` but I do not think you need to run

```sql
ADD CONSTRAINT "orgs_onboarding_valid" CHECK (...) NOT VALID;

ALTER TABLE "public"."orgs"
VALIDATE CONSTRAINT "orgs_onboarding_valid";
```

supabase/functions/_backend/utils/org_onboarding_intent.ts
WcaleNieWolny: I still do not understand. Please explain this to me :)

supabase/functions/_backend/public/organization/post.ts
WcaleNieWolny: Fixed.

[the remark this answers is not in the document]
WcaleNieWolny: fixed :)

WcaleNieWolny: Fixed

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,62 @@
...
+    OR (
+      ("jsonb_typeof"(("onboarding" -> 'development_environment'::"text")) = 'string'::"text")
+      AND (("onboarding" ->> 'development_environment'::"text") = ANY (ARRAY['hosted_builder'::"text", 'local_project'::"text", 'exploring'::"text"]))
+    )
+  )
+  AND (
+    (NOT ("onboarding" ? 'intent'::"text"))
+    OR (
+      ("jsonb_typeof"(("onboarding" -> 'intent'::"text")) = 'string'::"text")
+      AND (("onboarding" ->> 'intent'::"text") = ANY (ARRAY['ota'::"text", 'builder'::"text", 'both'::"text", 'exploring'::"text", 'publish'::"text"]))
```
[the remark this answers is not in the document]
WcaleNieWolny: Ignore this, I cannot guarantee that there are no broken rows in public.users. AI is stupid here I think

supabase/functions/_backend/utils/ab_tests.json
```diff
@@ -1,10 +1,32 @@
 {
```
WcaleNieWolny, line 1: I feel like you could add comments explaining what are your A/B tests doing. it could just be a field called "comment" - it's going to be ignored either way I believe during parsing, so it's fine (I think, double check)

supabase/functions/_backend/utils/ab_tests.ts
WcaleNieWolny: Done. Replica is the fast path (`getPgClient(c, true)`). If that lookup fails or the replica is missing assignments, we fall through to the primary with `FOR UPDATE`.

supabase/functions/_backend/utils/org_onboarding_intent.ts
WcaleNieWolny: Bento emails need a next-step URL. `publish` is a make-a-native-app path, so it uses the same dashboard landing as `builder` (`/apps`) instead of the OTA create-app flow (`/app/new`). Comment is next to the mapping.

src/utils/onboardingABTests.ts
WcaleNieWolny: Agreed. WebNativeApp is recommended only when treatment C is on, the user picked `hosted_builder`, intent is `publish`, and they are starting out. `shouldShowWebNativeRecommendation` now requires all of those.

src/components/dashboard/AppOnboardingFlow.vue
WcaleNieWolny: Assignment now starts on welcome (`void refreshOnboardingABTests()`), and we wait up to 3s on Continue from welcome. If there is no welcome screen, we wait the same 3s before showing the intent step.

supabase/functions/_backend/utils/ab_tests.ts
WcaleNieWolny: Existing assignments win. The SQL merges `existing || candidates` (and the persist helper prefers stored keys), so calling the endpoint again cannot re-roll a branch.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
WcaleNieWolny: Left the extra org fields out of this CHECK on purpose so this PR stays on the WebNative keys. Happy to add the others in a follow-up.

supabase/migrations/20260902132624_expand_webnative_onboarding.sql
WcaleNieWolny: Done. Both drops are now `DROP CONSTRAINT IF EXISTS`.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
WcaleNieWolny: Done. `orgs_onboarding_valid` is added `NOT VALID`. No `VALIDATE CONSTRAINT`.

supabase/functions/_backend/utils/ab_tests.json
WcaleNieWolny: Done. Both tests now have a `comment` field. The parser still ignores unknown keys.

supabase/functions/_backend/private/onboarding_ab_tests.ts
```diff
@@ -0,0 +1,18 @@
+import type { MiddlewareKeyVariables } from '../utils/hono.ts'
+import { Hono } from 'hono/tiny'
+import { getOrCreateUserABTests } from '../utils/ab_tests.ts'
+import { quickError, useCors } from '../utils/hono.ts'
+import { middlewareAuth } from '../utils/hono_middleware.ts'
+
+export const app = new Hono<MiddlewareKeyVariables>()
```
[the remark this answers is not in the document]
WcaleNieWolny: This route is mounted under `private/index.ts` (`createHono('private')`). Wrapping the sub-app in `createHono` would double middleware and add a Deno basePath, same as `/observe`. Left as `new Hono()` on purpose.

tests/webnative-onboarding-schema.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      intent: 'publish',
+    })
+
+    const orgs = await executeSQL<{ onboarding: Record<string, unknown> }>(
+      `INSERT INTO public.orgs (id, name, management_email, created_by, onboarding)
+       VALUES ($1, $2, $3, $4, jsonb_build_object('intent', 'publish'))
+       RETURNING onboarding`,
+      [orgId, `WebNative schema ${orgId}`, email, userId],
+    )
+    expect(orgs[0]?.onboarding).toEqual({ intent: 'publish' })
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. Org fixture is inserted in beforeAll; the first test now SELECTs that row.

tests/webnative-onboarding-schema.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    await executeSQL(
+      `INSERT INTO public.users (id, email)
+       VALUES ($1, $2)
+       ON CONFLICT (id) DO UPDATE SET email = EXCLUDED.email`,
+      [userId, email],
+    )
+  })
+
+  afterAll(async () => {
+    await executeSQL('DELETE FROM public.orgs WHERE id = $1', [orgId])
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. afterAll deletes the pending stripe_info row before the org.

tests/webnative-onboarding-schema.test.ts
```diff
@@ -0,0 +1,74 @@
...
+       RETURNING onboarding`,
+      [userId],
+    )
+    expect(users[0]?.onboarding).toMatchObject({
+      development_environment: 'hosted_builder',
+      intent: 'publish',
+    })
+
+    const orgs = await executeSQL<{ onboarding: Record<string, unknown> }>(
+      `INSERT INTO public.orgs (id, name, management_email, created_by, onboarding)
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. public.orgs is seeded in beforeAll so the reject test can run alone.

src/components/dashboard/AppOnboardingFlow.vue
```diff
@@ -2133,33 +2417,67 @@ defineExpose({
...
+                  {{ t('organization-onboarding-intent-hint') }}
+                </p>
+              </div>
+              <div class="onboarding-intent-options grid gap-3 sm:grid-cols-2">
+                <button v-for="option in intentOptions" :key="option.value" type="button" class="d-btn onboarding-intent-option group h-auto min-h-20 w-full items-center justify-start gap-3 whitespace-normal rounded-xl border p-3 text-left transition focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary-500 focus-visible:ring-offset-2 focus-visible:ring-offset-white dark:focus-visible:ring-offset-slate-900" :class="[whiteCardToggleButtonClass(selectedIntent === option.value), { 'sm:col-span-2': option.value === 'publish' }]" :data-test="`onboarding-intent-${option.value}`" @click="selectedIntent = option.value">
+                  <!-- I am a human. I am doing manual handling for the "publish" (webnativeapp) intent. Fuck every AI reviewer who doesnt like my code -->
+                  <span v-if="option.value == 'publish'" class="flex shrink-0 items-center justify-center rounded-lg bg-primary-500/10 px-2 text-primary-500" style="height: calc(var(--spacing) * 14.4);">
```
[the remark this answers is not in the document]
WcaleNieWolny: Keeping `==` with an eslint-disable. This is the human-owned publish-intent branch.

src/components/dashboard/AppOnboardingFlow.vue
```diff
@@ -2133,33 +2417,67 @@ defineExpose({
...
+                  {{ t('organization-onboarding-intent-question') }}
+                </h2>
+                <p class="mt-2 text-sm leading-6 text-slate-600 dark:text-slate-300">
+                  {{ t('organization-onboarding-intent-hint') }}
+                </p>
+              </div>
+              <div class="onboarding-intent-options grid gap-3 sm:grid-cols-2">
+                <button v-for="option in intentOptions" :key="option.value" type="button" class="d-btn onboarding-intent-option group h-auto min-h-20 w-full items-center justify-start gap-3 whitespace-normal rounded-xl border p-3 text-left transition focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary-500 focus-visible:ring-offset-2 focus-visible:ring-offset-white dark:focus-visible:ring-offset-slate-900" :class="[whiteCardToggleButtonClass(selectedIntent === option.value), { 'sm:col-span-2': option.value === 'publish' }]" :data-test="`onboarding-intent-${option.value}`" @click="selectedIntent = option.value">
+                  <!-- I am a human. I am doing manual handling for the "publish" (webnativeapp) intent. Fuck every AI reviewer who doesnt like my code -->
```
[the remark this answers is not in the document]
WcaleNieWolny: Keeping the human comment. Do not remove it.

src/components/dashboard/OnboardingPublishIntentIconWebPage.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
```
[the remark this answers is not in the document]
WcaleNieWolny: OnboardingPublishIntentIcon already ships the webpage mark inline. The isolated WebPage file is the same mark extracted for layout. Leaving it.

messages/en.json
```diff
@@ -2076,6 +2077,16 @@
   "organization-onboarding-invite-subtitle": "Invite teammates now or finish onboarding and do it later from the members page.",
   "organization-onboarding-invite-success-state": "Invitations sent. You can keep inviting teammates or create your first app.",
   "organization-onboarding-invite-title": "Invite teammates",
+  "organization-onboarding-development-environment-question": "What do you use to build your app?",
+  "organization-onboarding-development-environment-option-hosted_builder-label": "Hosted AI builder",
+  "organization-onboarding-development-environment-option-hosted_builder-desc": "Lovable, Bolt, Replit, v0…",
+  "organization-onboarding-development-environment-option-ai_assistant-label": "AI coding assistant",
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. development-environment keys are now ai_assistant → hand_coded → hosted_builder → other, desc before label. webnative keys sit after website.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,105 @@
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT IF EXISTS "orgs_onboarding_valid";
+
+-- NOT VALID skips existing-row verification, but later UPDATEs still evaluate
+-- the CHECK against the whole row. Strip keys the new org CHECK would reject.
+UPDATE "public"."orgs"
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. The JSONB cleanup UPDATEs now run before ALTER TABLE, so they are not under the DDL lock.

tests/webnative-onboarding-schema.unit.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const orgsDrop = webnativeOnboardingMigration.indexOf('DROP CONSTRAINT IF EXISTS "orgs_onboarding_valid"')
+    const orgsIntentBackfill = webnativeOnboardingMigration.indexOf(`SET "onboarding" = "onboarding" - 'intent'`, orgsDrop)
+    const orgsAdd = webnativeOnboardingMigration.indexOf('ADD CONSTRAINT "orgs_onboarding_valid"')
+
+    expect(orgsDrop).toBeGreaterThan(-1)
+    expect(orgsIntentBackfill).toBeGreaterThan(orgsDrop)
+    expect(orgsAdd).toBeGreaterThan(orgsIntentBackfill)
+    expect(usersDrop).toBeGreaterThan(orgsAdd)
+    expect(usersIntentBackfill).toBeGreaterThan(usersDrop)
+    expect(usersAdd).toBeGreaterThan(usersIntentBackfill)
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 9e405254c. The unit test now asserts both intent and development_environment backfills sit before each ADD CONSTRAINT.

supabase/functions/_backend/utils/ab_tests.json
WcaleNieWolny: Added. Both tests now have a `comment` field: publish intent only adds the option; development environment C plus hosted_builder plus starting out can recommend WebNativeApp. Parsing ignores unknown fields.

supabase/migrations/20260902132624_expand_webnative_onboarding.sql
WcaleNieWolny: The current migration uses DROP CONSTRAINT IF EXISTS for both orgs and users.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,105 @@
...
+UPDATE "public"."users"
+SET "onboarding" = "onboarding" - 'development_environment'
+WHERE ("onboarding" ? 'development_environment'::"text")
+  AND (
+    ("jsonb_typeof"(("onboarding" -> 'development_environment'::"text")) IS DISTINCT FROM 'string'::"text")
+    OR (("onboarding" ->> 'development_environment'::"text") <> ALL (ARRAY['hosted_builder'::"text", 'ai_assistant'::"text", 'hand_coded'::"text", 'other'::"text", 'local_project'::"text", 'exploring'::"text", 'skipped'::"text"]))
+  );
+
+ALTER TABLE "public"."users"
+DROP CONSTRAINT IF EXISTS "users_onboarding_valid";
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 1452b2e53. Drop users_onboarding_valid (and orgs_onboarding_valid) first, backfill invalid keys, then add the new CHECK NOT VALID. Backfill is no longer rechecked against the old step list, and it still sits between short DROP/ADD so it is not held under the DDL lock.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,107 @@
+-- DROP/ADD are short ACCESS EXCLUSIVE. Backfill sits between them so it is
+-- not held under the DDL lock, and so user rows that already use a new
+-- step are not rechecked against the old users_onboarding_valid list.
+
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT IF EXISTS "orgs_onboarding_valid";
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in 6517459f0. This file stays one transactional migration (no second file, no COMMIT/BEGIN). DROP CONSTRAINT would keep ACCESS EXCLUSIVE until commit, so backfill now runs first and the exclusive swap is only DROP+ADD NOT VALID. The old users CHECK already rejects step publish_app_question, so that UPDATE cannot 23514 on the new step.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,108 @@
...
+UPDATE "public"."orgs"
+SET "onboarding" = "onboarding" - 'development_environment'
+WHERE ("onboarding" ? 'development_environment'::"text")
+  AND (
+    ("jsonb_typeof"(("onboarding" -> 'development_environment'::"text")) IS DISTINCT FROM 'string'::"text")
+    OR (("onboarding" ->> 'development_environment'::"text") <> ALL (ARRAY['hosted_builder'::"text", 'ai_assistant'::"text", 'hand_coded'::"text", 'other'::"text", 'local_project'::"text", 'exploring'::"text", 'skipped'::"text"]))
+  );
+
+ALTER TABLE "public"."orgs"
+DROP CONSTRAINT IF EXISTS "orgs_onboarding_valid";
```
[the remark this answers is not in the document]
WcaleNieWolny: Fixed in f5bbfd219. Both orgs and users backfills now run before the first ALTER TABLE. The exclusive swaps are only DROP+ADD NOT VALID at the end of the transaction.

supabase/migrations/20260907163000_expand_webnative_onboarding.sql
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+-- Wizard progress never allows org default intent "unknown". Historical rows
+-- that still have it would 23514 on the next row update under the new CHECK.
+UPDATE "public"."users"
+SET "onboarding" = "onboarding" - 'intent'
+WHERE ("onboarding" ? 'intent'::"text")
+  AND (
+    ("jsonb_typeof"(("onboarding" -> 'intent'::"text")) IS DISTINCT FROM 'string'::"text")
+    OR (("onboarding" ->> 'intent'::"text") <> ALL (ARRAY['ota'::"text", 'builder'::"text", 'both'::"text", 'exploring'::"text", 'publish'::"text"]))
```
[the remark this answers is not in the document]
WcaleNieWolny: Not a 23514 path. The backfill is `onboarding - intent` on rows whose intent is already invalid. CHECK is evaluated on the new tuple; after the key is removed the old users_onboarding_valid clause `(NOT (onboarding ? intent))` passes. Production has 0 user rows with unknown/invalid intent. Dropping the CHECK first would hold ACCESS EXCLUSIVE across the users scan, which is the lock issue we just fixed. Keeping all backfills before the first ALTER TABLE.

WcaleNieWolny (approved):

PR #3284, opened 2026-09-08
feat(onboarding): hard-gate sidebar before first app (explicit Continue)

TorichanCapgo:
## Summary

- Hard-gate Dashboard / Apps / API keys / Scan until the user finishes creating their first app (or confirms **Explore anyway**).
- Intent step requires explicit **Continue** after selecting an option — **no one-tap auto-advance** (removed to avoid miss-clicks and keep room for #3287 / #3237 A/B + intent emails).
- Reuses confirm dialog: **Finish creating your first app?** / **Continue setup** primary / **Explore anyway** secondary.
- Unit tests for hard-gate + redirect behavior.

## Motivation

PostHog: many users hit dashboard exploration before first app create. Pre-create sidebar hard-gate keeps them on setup. One-tap auto-advance was dropped (product B) so intent selection stays intentional and compatible with Wolny’s A/B work.

## Out of scope

- Post-create sticky gate until first upload → #3285 (draft, stacked)
- Intent A/B / Bento emails → #3287

## Test plan

- [ ] Select intent option → stays on step until **Continue**
- [ ] Sidebar Dashboard/Apps/API keys/Scan → confirm dialog; Continue setup returns to onboarding; Explore anyway proceeds
- [ ] Org-switcher path still excluded
- [ ] Unit tests pass

## Screenshots / recording

Intent select (highlight only, no auto-advance) + explicit **Continue**:

intent

Selected intent before Continue:

selected

Sidebar hard-gate confirm dialog:

dialog

Details / Explore anyway path:

details

Recording: [pr3284-one-tap-hardgate.mp4](https://github.com/Cap-go/capgo.app/raw/9afac794e657d1bd84a65ae3f66b4bd730e0066a/.github/pr-assets/3284/pr3284-one-tap-hardgate.mp4)

## Summary by CodeRabbit

- **New Features**
  - Improved onboarding navigation when exploring the dashboard, including clearer options to continue setup or explore the dashboard.
  - Continue-setup actions now preserve the appropriate onboarding step and route.
  - Dashboard exploration confirmations now account for the current onboarding context.

- **Bug Fixes**
  - Fixed onboarding intent selection so choosing an option does not advance until Continue is selected.
  - Improved onboarding option layout and cleanup when leaving the flow.

- **Style**
  - Updated English onboarding text with clearer, more concise descriptions and button labels.


src/utils/onboardingRedirect.ts
```diff
@@ -117,14 +117,49 @@ export function canExploreOnboardingDashboard(userId: string | null | undefined)
...
+  if (
+    (path === '/onboarding/organization' || path.startsWith('/onboarding/organization/'))
+    && options?.source === 'org-switcher'
+  ) {
+    return false
+  }
+  return path === '/app/new'
+    || path === '/onboarding/app'
+    || path.startsWith('/onboarding/')
+}
```
blehstbl, line 155: `path.startsWith('/onboarding/')` also matches `/onboarding/invitation` and `/onboarding/set_password`. Those are not the first-app create flow, but they still trip `shouldConfirmOnboardingDashboardExploration` for `/dashboard`, `/apps`, `/apikeys`, and `/scan`.

The org-switcher exception only covers `/onboarding/organization`. Someone accepting an invite (or setting a password) who then hits Apps/Dashboard in the sidebar will get "Finish creating your first app?".

Safer to allowlist the create routes (`/app/new`, `/onboarding/app`, and `/onboarding/organization` when it is not `org-switcher`) instead of every `/onboarding/*` path.

blehstbl (commented):
One real hole in the hard-gate path: `/onboarding/invitation` and `/onboarding/set_password` are treated as first-app create because of the `/onboarding/` prefix match. Left a comment on the helper.

src/utils/onboardingRedirect.ts
TorichanCapgo: Good catch — agree invitation/set_password should not hard-gate. Narrowing `isPreCreateOnboardingPath` to an allowlist of create routes (`/app/new`, `/onboarding/app`, `/onboarding/organization` unless `org-switcher`) and adding unit coverage. Fix pushing to this branch shortly.

TorichanCapgo: Fixed in `1fe519c`: `isPreCreateOnboardingPath` is now an allowlist (`/app/new`, `/onboarding/app`, `/onboarding/organization` unless `org-switcher`). `/onboarding/invitation` and `/onboarding/set_password` no longer confirm. Unit coverage added in `tests/onboarding-redirect.unit.test.ts`.

src/components/Sidebar.vue
```diff
@@ -154,31 +154,38 @@ async function openTab(tab: Tab) {
...
+    // Primary = stay in setup (safe default). Secondary = explore anyway.
+    if (dialogStore.lastButtonRole === 'primary') {
+      const stepQuery = typeof route.query.step === 'string' ? { step: route.query.step } : {}
+      if (onboardingResumeAppId) {
+        return router.push({
+          path: '/app/new',
+          query: { resume: onboardingResumeAppId, ...stepQuery },
+        })
+      }
+      return router.push({ path: '/app/new', query: stepQuery })
```
Nei1eveN: **[P2] Keep “Continue setup” on the current organization step**

For the first-app flow at `/onboarding/organization?org=fixture-org&step=logo` (or `step=invite`, without `source=org-switcher`), the new guard correctly shows the confirmation. But selecting the primary **Continue setup** action calls `router.push({ path: '/app/new', query: { step: 'logo' } })`: it leaves the organization page and drops `org`.

That destination renders `AppOnboardingFlow` without `pre-org` (`src/pages/app/new.vue`). Its `resumeStep` only accepts `choice`, `install`, or `setup`, and without a resume app its mount sets the flow to app details/name. The organization logo/invite step therefore cannot be resumed by copying its `step` value to this different page. The same branch also leaves an in-progress first-org details form. This is separate from the already-fixed org-switcher exclusion: these are the first-org routes that should remain gated.

Please dismiss/return without navigating when the user chooses to continue an already-active onboarding route, retaining that page and its query/state. Keep any needed redirect-to-resume behavior for entry points outside that active flow.

Validation at `1fe519c`: executed the exact `shouldConfirmOnboardingDashboardExploration` helper and `Sidebar.openTab` with mocked dialog/router/storage boundaries. Primary on organization details/logo/invite produced the incorrect `/app/new` targets; dismissal made no navigation, secondary explored the requested dashboard, org-switcher skipped the dialog, and resumed app-choice preserved its resume/step. Destination behavior above is source analysis; no mounted Vue/browser journey, router middleware, backend, or full project suite was run.

AI-generated review and local test execution by Codex.

Nei1eveN (commented):
AI-generated review by Codex. One first-organization navigation finding, tested with exact source and mocked boundaries; limitations are included inline.

src/components/Sidebar.vue
TorichanCapgo: Agreed — primary **Continue setup** on an already-active first-org route (`/onboarding/organization?org=…&step=logo|invite|details`, no `org-switcher`) should dismiss without navigating, so we keep `org`/step state instead of bouncing to `/app/new`.

Fix is queued on this branch; Cursor cloud-agent capacity is currently exhausted (on-demand usage needed), so the push may lag until that clears. Will reply here with the SHA once it lands.

TorichanCapgo: Fixed in `75dbda9`.

Primary **Continue setup** now returns without navigating when the hard-gate dialog fires, so an already-active first-org route (`/onboarding/organization?org=…&step=logo|invite|details`) keeps `org`/step state instead of bouncing to `/app/new`. Secondary Explore anyway unchanged.

Nei1eveN: Retested `75dbda98f058f3fc0e80747fce73fe678fdaf83b` against the originally reported `1fe519c`. The reported primary-action navigation is fixed in the local handler/helper test.

- On organization details/logo/invite, primary Continue setup now makes **no router push** and leaves the route/query intact; it also does not grant dashboard exploration.
- Dismissal still stays put.
- Secondary Explore still navigates to the requested dashboard, emits the exploration event and records the grant.
- The org-switcher control still skips the confirmation; resumed app-choice now stays on its existing resume/step route without pushing it again.

Executed the exact public Sidebar handler and redirect helper for both revisions in Node 22.22.2, with mocked dialog/router/storage boundaries. This is a targeted retest of my finding, not a mounted Vue/browser journey, router-middleware/backend test, full-suite result or overall PR approval.

AI-generated follow-up and test execution by Codex.

src/utils/onboardingRedirect.ts
TorichanCapgo: Already addressed earlier on this PR in `1fe519c` (allowlist: `/app/new`, `/onboarding/app`, `/onboarding/organization` unless `org-switcher`). `/onboarding/invitation` and `/onboarding/set_password` no longer hard-gate. Fresh CR on `75dbda9` raised two new Sidebar/type nits — fix agent pushing next on this branch.

WcaleNieWolny (changes requested):
I do not like this PR. Auto-advancing after selecting the intent is quite bad from my point of view. 

I am working on a separate PR that adds A/B tests that are dependent on the intent selected by the user. If user selects `OTA` intent and clicks continue I will use the new A/B tests assignment endpoint implemented in this PR: https://github.com/Cap-go/capgo.app/pull/3237 to assign the user A/B tests that require the `OTA` intent. 

For example, I want to add "channel creation" into initial onboarding, but I want to do it via A/B tests. If we merge this PR and the user selects the wrong intent, they may never get assigned the correct A/B test.

Furthermore, intent is CRITICAL for emails. More and more emails will become customized based on the intent. Therefore, allowing the user to select the wrong intent is really, really bad, and I strongly want to avoid having users who miss click the intent.

I would personally love to close this PR without merging

tests/app-onboarding-progress-integration.unit.test.ts
```diff
@@ -602,6 +608,6 @@ describe('app onboarding progress analytics integration', () => {
     expect(demoExit.indexOf('window.dispatchEvent')).toBeLessThan(demoExit.indexOf('allowOnboardingDashboardExploration'))
 
     const confirmedSidebarExit = sidebarSource.slice(sidebarSource.indexOf('if (requiresOnboardingExplorationConfirmation)'), sidebarSource.indexOf('if (tab.onClick)'))
-    expect(confirmedSidebarExit.indexOf(`lastButtonRole !== 'primary'`)).toBeLessThan(confirmedSidebarExit.indexOf('window.dispatchEvent(new Event(ONBOARDING_DASHBOARD_EXPLORED_EVENT))'))
+    expect(confirmedSidebarExit.indexOf(`lastButtonRole !== 'secondary'`)).toBeLessThan(confirmedSidebarExit.indexOf('window.dispatchEvent(new Event(ONBOARDING_DASHBOARD_EXPLORED_EVENT))'))
```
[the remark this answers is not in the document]
TorichanCapgo: Fixed on `c18065d` — assertion now fail-closes: both needles must exist (`>= 0`) before the order check.

WcaleNieWolny:
Don’t merge yet, let me review

TorichanCapgo:
@WcaleNieWolny re-review please — your CHANGES_REQUESTED was about one-tap / auto-advance after intent.

That path is gone on current HEAD (`e3f0ecef7`): intent cards only set selection; advance requires explicit **Continue**. Hard-gate (sidebar + direct `/apps` and `/apikeys`) remains. Compatible with your #3237 intent-scoped A/B assignment on Continue.

TorichanCapgo:
@coderabbitai review

WcaleNieWolny:
@TorichanCapgo I do not understand the issue this PR is trying to solve. What is the underlying business problem / customer issue that made you create this pr ?

PR #3341, opened 2026-09-15
fix(frontend): show log action keys on hover instead of a name/key toggle

riderx:
## Summary (AI generated)

- Removed the logs toolbar **Name / Key name** toggle.
- Action column shows the translated name by default.
- Hover (and keyboard focus) swaps in the canonical `stats_action` key; the native `title` is also the key.
- Original plugin error text still shows under the action label.

## Motivation (AI generated)

The Name/Key toggle was not the requested UX. Operators want the friendly translated label in the table, and the underscore key only when they hover.

Faster log triage without an extra control: readable names in the grid, keys available on demand for docs and support.

## Test Plan (AI generated)

- [x] Unit tests for original error extraction
- [x] Playwright `observe-tabs.spec.ts`: translated name by default, key on hover, original error stays
- [ ] CI full suite

## Visual changes (AI generated)

**Default vs hover (same row):**

Default translated name vs hover action key

**Full table (no Name/Key toggle):**

Logs table

Generated with AI


src/services/logTableDisplay.ts
```diff
@@ -48,3 +48,45 @@ export function logRowDisplayMetadata(versionName: string | null | undefined, me
...
+  'description',
+] as const
+
+export function extractLogOriginalMessage(metadata: LogMetadata): string | null {
+  const normalized = normalizeLogMetadata(metadata)
+  if (!normalized)
+    return null
+
+  for (const key of LOG_ORIGINAL_MESSAGE_KEYS) {
+    const value = normalized[key]?.trim()
```
rihoarvutikonto: `normalized[key]?.trim()` will throw if the metadata value is not a string. `normalizeLogMetadata` types the object as `Record<string, string>` but does not coerce values: jsonb from `/private/stats` and `JSON.parse` of a string payload routinely produce numbers, booleans, or nested objects (`statusCode: 500`, `error: { code: 1 }`, `message: ["a","b"]`). Optional chaining only covers null/undefined, so `.trim()` is a TypeError — and this runs inside `LogTable.vue`'s action `renderFunction`, so one bad row can blank the whole logs table.

Guard before calling string methods, and skip non-strings instead of crashing:

```ts
for (const key of LOG_ORIGINAL_MESSAGE_KEYS) {
  const value = normalized[key]
  if (typeof value === 'string') {
    const trimmed = value.trim()
    if (trimmed)
      return trimmed
  }
}
```

Worth a unit case with `{ message: { text: 'boom' } }` and `{ error: 404 }` so the table stays up. `description` as a last-resort key is also broad — several stats events stash a non-error summary there, which would show as the "original error" once the crash is fixed.

rihoarvutikonto (commented):
File-level: `extractLogOriginalMessage` calls `.trim()` on jsonb values that are not always strings, which can throw inside the logs table renderer.

PR #3346, opened 2026-09-16
fix(auth): preserve password reset verification parameters

WcaleNieWolny:
Password-reset emails can place `type=recovery` and `redirect_to` outside the nested `confirmation_url`. The confirmation page then forwards only the token, and Supabase rejects the request with “Verify requires a verification type.”

Restore missing `type` and `redirect_to` parameters on allowed `/auth/v1/verify` links with a token. Preserve parameters already inside the verification URL, avoid decoding their contents a second time, and continue supporting legacy links with an extra layer of URL encoding. This repairs already-sent emails when the console is deployed; no email-template upload is required.

Validation:
- Both browser regression tests reproduced the failures before the fix and pass afterward, including against a production preview matching CI.
- 17 new unit cases cover malformed recovery links, nested query encoding, parameter precedence, other auth email types, legacy encoding, and redirect safety.
- Full unit suite: 328 files and 2,871 tests pass with `TZ=UTC` (matching CI; existing statistics date fixtures are timezone-sensitive).
- `bun lint`, `bun lint:backend`, `bun typecheck`, `bun run lint:deadcode`, and `bun run build` pass.


WcaleNieWolny:
@coderabbitai approve the PR

PR #3353, opened 2026-09-16
fix(cli): clarify channel access errors and align channel tables

WcaleNieWolny:
`channel list` can fail during its app-access preflight with a generic "Edge Function returned a non-2xx status code" message. App access failures now identify the API key and `app.read` permission to check, and channel access failures identify `app.read_channels`. Other API errors retain their response details instead of claiming the app does not exist. Successful array and single-channel responses both preserve the returned channels.

Replace the channel table's emoji cells with Yes/No values using the shared terminal formatter. Show Unlinked for missing bundles, keep columns aligned, and switch to per-channel Setting/Value tables with wrapped values on narrow terminals. Include iOS/Android in channel API reads so the CLI displays the actual platform settings.

Validation: full `bun run cli:test` suite passed, including MCP smoke and bundle integrity tests. CLI/backend lint and typechecks passed. Channel-list regression tests cover permission failures, aligned wide/narrow output, platform values, pagination, and the built Node CLI's printed errors and exit codes. Channel API integration coverage runs in CI; local Docker startup was stopped after it resumed many existing stacks and overloaded the host.


WcaleNieWolny:
@coderabbitai approve

cli/src/api/channels.ts
```diff
@@ -307,9 +338,18 @@ export async function getActiveChannels(
...
+        if (!options.silent)
+          log.error(message)
+        throw new CliUserError(
+          message,
+          { appId: appid, requiredPermissionKey: 'app.read_channels' },
+        )
+      }
+      throw new Error(`Cannot list channels: ${await formatCapgoCliInvokeError(vError)}`, { cause: vError })
     }
     const batch = Array.isArray(data) ? data : []
```
rihoarvutikonto: `fetchChannelsPage` is typed as `HttpChannel | HttpChannel[]`, but a successful non-array payload is discarded as "no channels":

```ts
const batch = Array.isArray(data) ? data : []
if (!batch.length)
  break
```

If GET `/channel?app_id=&page=` ever returns a single object (one channel on the last page, or the same shape used when `channel=` is set), `channel list` prints `No channels found.` and exits 0. That is the opposite of the permission-error clarity this PR is adding — a 200 with one channel looks like an empty app.

Normalize both shapes:

```ts
const batch = Array.isArray(data) ? data : data ? [data] : []
```

Also, dropping the `GET channel does not currently return ios/android` TODO while still doing `row.ios ?? false` / `row.android ?? false` will print `iOS: No` / `Android: No` whenever those fields are omitted. If the HTTP handler is not in this PR, that is a behavior change from "unknown, defaulted for display" to a lying Yes/No table. Keep an explicit unknown (or only print Yes/No when the key is present) until the API actually returns the flags.

rihoarvutikonto (commented):
File-level: `getActiveChannels` treats a non-array 200 payload as an empty list, so a single-object channel response would print "No channels found."

PR #3364, opened 2026-09-17
feat(cli): detect installed updater in the background

WcaleNieWolny:
Commands run in a Capacitor app now check for an installed `@capgo/capacitor-updater` in a separate best-effort worker and mark the Install Updater Plugin onboarding step (`add_updater`) done when the selected app declares the package and it is installed locally. This includes hoisted and symlinked workspace installations; absent dependencies, ambiguous app selection, errors, and interrupted checks leave progress untouched.

The updater worker reuses project discovery and the existing install verifier. Shared launch/reporting helpers preserve the source worker's behavior, keep both workers silent and unreferenced, and send only the detected step's completion patch. The new check emits direct `scan_started` / `scan_ended` events in the `updater-installed` channel with its own `attempt_id`, respecting analytics opt-out.

Background requests validate both transport and destination trust before sending credentials. The default Capgo API is trusted; custom API origins require an explicit self-host CLI selection or the `CAPGO_TRUSTED_API_ORIGINS` allowlist. Remote HTTP and untrusted destinations are skipped, and telemetry/onboarding requests reject redirects. Other CLI requests retain their existing behavior.

Validation passed: frozen dependency install; CLI lint and typecheck; full CLI suite (including build, MCP smoke and bundle integrity); all 23 background worker tests; root lint and unused-export checks; help/version smoke checks. The background suite covers both workers, declaration-only false positives, hoisted/symlinked installs, app targeting, telemetry/reporting outcomes, foreground shutdown while both workers are busy, destination trust and encryption, and credential isolation across redirects.


WcaleNieWolny:
@coderabbitai approve the pr

PR #3377, opened 2026-09-17
refactor(backend): queue PostgreSQL onboarding refreshes

WcaleNieWolny:
The hourly SQL batch refreshed onboarding features for every app, while the backend queue implementation added a separate lease table and changed the evidence source to Cloudflare Analytics Engine. This PR keeps the old PostgreSQL feature calculations and moves only the bounded execution into the backend.

The existing SQL scheduler now runs a producer every 10 minutes. It selects up to 500 due apps whose organization is paying, in trial, or has an unexpired credit grant with remaining credits. The producer atomically writes `apps.onboarding.queued_refresh_at` and enqueues groups of at most 25 app IDs. The existing minute queue dispatcher handles up to four messages (100 apps) per run. Messages older than 30 minutes can be queued again if no refresh has completed.

The backend consumer reads `devices`, `app_versions`, `daily_version`, and `build_requests` using the old predicates in one Drizzle transaction, then locks the selected apps and merges only the three feature entries plus `refreshed_at` into the current JSON. Replayed or overtaken messages skip apps already refreshed after their enqueue time. The old scheduled batch RPC is removed; the single-app verification RPC remains. No lease table, batch token, or Cloudflare analytics credential is needed.

Validation: frontend/backend lint, typecheck, production build, and the full CLI check passed; 3,205 unit tests passed with `TZ=UTC`, and 294 Tinbase database tests passed. A throwaway PostgreSQL 17 instance applied the migration and passed synthetic assertions for billing eligibility, 25-app grouping, 30-minute recovery, and the 500-app cap. The actual consumer transaction passed milestone, JSON-preservation, and replay assertions there. An `EXPLAIN (ANALYZE, BUFFERS)` selection of 500 from 20,000 synthetic apps used the queued-refresh index and completed in 0.58 ms. The full Supabase/Docker integration run was unavailable locally because the Docker PostgreSQL port did not respond.


WcaleNieWolny:
@coderabbitai approve the pr

PR #3384, opened 2026-09-19
feat(onboarding): check v3 OTA progress throughout AI setup

WcaleNieWolny:
When a user copies the AI setup prompt with explicit `intent=ota` for a Todo list v3 app, the assistant now runs `app todo` after app selection and after milestones that can advance onboarding. Each check reads the current statuses and next pending step instead of assuming work is complete. The prompt uses the v3 app IDs from the login page or dashboard flow, so v2-only apps keep their existing instructions; mixed organizations skip the checklist when a v2 app is selected.

The checkpoints cover channel setup, updater installation, the app-ready call, first upload, native device registration, and first live update. Guided `init` also gets a follow-up check after the user finishes it. The initial CLI result must confirm v3 before later checkpoints run.

Validation: `bun lint` (0 errors; 35 existing warnings), `TZ=UTC bun test:unit` (344 files, 3,209 tests), `bun run typecheck:frontend`, and `bun run build`.

Release dependency satisfied: npm now serves `@capgo/cli@latest` as 8.55.0, which includes `app todo`.


WcaleNieWolny:
@coderabbitai approve

